PR #10, opened 2019-01-24
Prevent react browser from opening for start task only

ashchan:
For start:react task use the default behavior.


ashchan:
Closing in favor of #11.

Issue #6, opened 2019-01-23
Project packaging


ashchan:
Use tools like `electron-packager`/`electron-builder` to make Neuron available as native desktop for Windows, macOS and Linux.

* When we develop and run the app in development mode, the electron app loads `http://localhost:3000`, which is the local react app web server address.
* When packaging:
  * The electron app should load static build of react app.
  * The electron app should load and use our own app icon.
  * We should be able to build with our development provisioning files (but we can do this in a later and separate task).
  * The app should be able to up and running on Windows, macOS and Linux.
  * Anything else?

PR #11, opened 2019-01-24
enable cross env

Keith-CY:
enable cross env to set environment variables as #6


ashchan:
Seems to be an NPM server issue.

```
error An unexpected error occurred: "https://registry.npmjs.org/@lerna/describe-ref/-/describe-ref-3.10.0.tgz: Request failed \"500 Internal Server Error\"".
```

PR #17, opened 2019-01-29
configuration for yarn package-win

wuyuyue:


wuyuyue:
we use electron-builder, I find we could config on seperate file "electron-builder.yml" instead of package.json(success in windows); the detail as the format https://github.com/Azure/BatchExplorer/blob/master/electron-builder.yml

ashchan:
> we use electron-builder, I find we could config on seperate file "electron-builder.yml" instead of package.json(success in windows); the detail as the format https://github.com/Azure/BatchExplorer/blob/master/electron-builder.yml

This looks cool.

packages/electron-app/src/main.ts
```diff
@@ -8,15 +8,15 @@ let mainWindow: Electron.BrowserWindow | null
...
+  PROD: `file://${path.join(__dirname, '../../react-app/build/index.html')}`,
 }
 
 function createWindow() {
   // Create the browser window.
   mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
-    icon: `${__dirname}/../build/icon.icns`,
+    // icon: `${__dirname}/../build/icon.icns`,
```
ashchan: Instead of commenting this out, we can just delete it.

electron-builder.yml
```diff
@@ -0,0 +1,16 @@
+appId: com.nervos.neuron
```
ashchan: There's still an `electron-builder.yml` file in the root folder. Was this supposed to be deleted?

packages/electron-app/src/main.ts
ashchan: Oh I see. That icon file was removed (perhaps by accident) from this commit:

https://github.com/nervosnetwork/neuron/commit/3a36b6c103bd2179d0667451f18adfa30886faea#diff-de61d9af2b61d0fbe27ceb0ee8a5f555

@Keith-CY We had that file in `build` folder with the purpose to bundle that for the Mac app. Could you take a look at that and see if we can bring it back? (maybe we should not ignore the `build` folder?)

PR #20, opened 2019-01-31
feat: add network status ui

cezres:


packages/react-app/src/components/Network/SwitchNode.tsx
```diff
@@ -0,0 +1,118 @@
...
+      })
+    } else {
+      this.setState({
+        inputErrorMessage: 'Please enter a valid node address',
+      })
+    }
+  }
+}
+
+function IsURL(str_url: string) {
```
ashchan: A function should be named in `camelCase`.

ashchan: Function param should be in `camelCase` as well.
```suggestion
function ssURL(strUrl: string) {
```

packages/react-app/src/components/Network/SwitchNode.tsx
```diff
@@ -0,0 +1,118 @@
...
+  ModalDialog,
+  ModalContent,
+  ModalHeader,
+  ModalBody,
+  ModalFooter,
+  Typography,
+  Button,
+  Input,
+  Alert,
+} from '@smooth-ui/core-sc'
```
ashchan: We've dropped smooth in favor of grommet. Please use the new UI lib.

PR #34, opened 2019-02-07
feat: add async get cells example

Keith-CY:


package.json
```diff
@@ -21,7 +21,7 @@
     "lint": "lerna run lint",
     "test": "lerna run --parallel test",
     "start:react": "lerna run --scope=neuron-react-app start",
-    "start:electron": "lerna run --scope=neuron start",
+    "start:electron": "lerna run --scope=neuron watch",
```
ashchan: Is these `watch` script temporary? On my machine it starts the electron app with two instances, with the window title bar overlaid with main content, and unable to de dragged and moved.

PR #35, opened 2019-02-07
feat(react-app): add grid layout

Keith-CY:


packages/react-app/src/styles/layout.scss
```diff
@@ -0,0 +1,23 @@
...
+    'header header header'40px 'sidebar mainContainer mainContainer'1fr 'sidebar mainContainer mainContainer'1fr 'sidebar mainContainer mainContainer'1fr / 240px 1fr 1fr;
+
+  height: 100vh;
+}
+
+header {
+  grid-area: header;
+}
+
+aside {
```
ashchan: Will `header` and `aside` style affect other parts of the HTMl? Say if for some feature we add another `header` on some part of the main content, do they apply these grid style as general CSS rules?

Keith-CY: The grid property only works when` header`'s parent is set to `display: grid` so when we add a `header` in a `section` the `grid-area: header` not works.

ashchan: Wonderful!

PR #38, opened 2019-02-09
chore(neuron-ui): add react hooks rules

Keith-CY:


ashchan:
Menu: Next release of `react-scripts` will include this plugin by default.

ashchan:
@Keith-CY Another thing: this plugin was added to `neuron-ui` package; but we have other aslint plugins on the root package. (In fact `react` related plugins should go to `neuron-ui` package but `airbnb` depends on `eslint-plugin-react`, `eslint-plugin-jsx-a11y` and others so we just place them there. 😿

Keith-CY:
> @Keith-CY Another thing: this plugin was added to `neuron-ui` package; but we have other aslint plugins on the root package. (In fact `react` related plugins should go to `neuron-ui` package but `airbnb` depends on `eslint-plugin-react`, `eslint-plugin-jsx-a11y` and others so we just place them there. 😿

[eslint-config-airbnb](https://github.com/airbnb/javascript) is a general eslint preset working for any JavaScript Projects so it also works for `neuron-wallet`, we can install it in `neuron-wallet` and `neuron-ui`, respectively, but react-related eslint-plugn-xx will be added to `neuron-wallet` too.

ashchan:
Yes. My point is we use it as a general preset but it actually requires `eslint-plugin-react` already.

Keith-CY:
> Yes. My point is we use it as a general preset but it actually requires `eslint-plugin-react` already.

I've got that. We can use [eslint-config-airbnb-base](https://www.npmjs.com/package/eslint-config-airbnb-base) at the root and move React-related to `neuron-ui`.

PR #42, opened 2019-02-11
feat(ui): two-column layout (WIP)

ashchan:


packages/neuron-ui/src/styles/layout.scss
```diff
@@ -1,21 +1,22 @@
...
 }
 
-.mainContainer {
-  grid-area: mainContainer;
+.main-content {
+  grid-area: content;
+  overflow: auto;
 }
 
 .notification {
```
Keith-CY: If `body.grid-template` is set to ` 'sidebar header' 60px 'sidebar content' 1fr 'sidebar content' 1fr / 240px 1fr;`, `.notification`'s `grid-area` should be modified to proper position.

ashchan: Makes sense and we can adjust that later.

I'm think perhaps we should place `notification` inside header to make it more easier to be noticed.

Keith-CY: Adjust it later.

I added it as a container for now, not necessary. I prefer using electron api for notification.

PR #46, opened 2019-02-11
Sidebar nav active state

ashchan:


packages/neuron-ui/src/components/Router/index.tsx
```diff
@@ -23,75 +25,79 @@ interface CustomRoute {
...
   icon?: Icon
-  showInSidebar?: boolean
   exact?: boolean
   component: React.ComponentType
 }
 
 export const routes: CustomRoute[] = [
   {
     name: 'Header',
-    showInSidebar: false,
```
ashchan: @Keith-CY I removed this property because we might use part of routes in several places, not just on the sidebar. Considering that, the full routes collection should not have concern about where they'll be used. (Otherwise it'll have `showInSidebar`, `showInTopNav`, `showInFooter`, etc. and quickly become too complicated.)

packages/neuron-ui/src/components/Router/index.tsx
```diff
@@ -23,75 +25,79 @@ interface CustomRoute {
...
   },
   {
     name: 'Sidebar',
-    showInSidebar: false,
     path: '/',
     exact: false,
     component: Sidebar,
   },
+  {
+    name: '[CurrentWallet]',
```
ashchan: We need to find a way to replace `[CurrentWallet]` (or is there a standard way to present a placeholder?) with the actual wallet name.

We might also want to make this link a dropdown menu, from which we can toggle to show other wallets to choose to switch to, and to import/create wallet quickly.

Keith-CY: And wallet switch might not be a group of links, it should send a message to neuron-wallet to switch wallet, then neuron-wallet send related information to neuron-ui.

Keith-CY: The dropdown can be added as a component to render 

```
{
  name: '[currentWallet]',
   ..., 
  children: [{
    name: wallet1,
    path: 'wallet/1',
  }]
}
```
and only rendered when `children` exists.

Keith-CY: BTW, if the dropdown is for switching wallet, instead of navigating to other pages, it's better to send a message to neuron-wallet but not to be a link.

ashchan: @Keith-CY I agree. Switching wallet should be a message. And the current wallet should always has the fixed route path.

ashchan:
https://github.com/nervosnetwork/neuron/pull/47

PR #48, opened 2019-02-12
add modal UI component sample

wuyuyue:


Keith-CY:
It seems too complicated, we can just create a widget like

```
const Dialog = ({ children, open }) => <dialog open={open}>{children}</dialog>
export default () => createPortal(<Dialog />, document.querySelector('#modal'))
```

and use it in other components like

```
render() {
    <Dialog open={open}> This will be rendered in dialog </Dialog>
}
```

packages/neuron-ui/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+          >
+            x
+          </div>
+        </div>
+      </AppModal>
+    )}
+  </ModalContext.Consumer>
+)
+
+let openGuider: boolean = !getGuidedFlag()
```
ashchan: `Guider` is not a good name for our purpose; `guide` is better.

packages/neuron-ui/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const Container = (props: any) => {
+  const modalContext = useContext(ModalContext)
+  // console.log(props)
```
ashchan: Do not leave debug output (even if they are commented out).

packages/neuron-ui/src/utils/storage.ts
```diff
@@ -0,0 +1,4 @@
+export const getGuidedFlag = () => window.localStorage.getItem('guided')
```
ashchan: The create/import guide is not shown by checking if it's already displayed before. Instead, it will show when user has 0 wallets; or when user explicitly start creating/importing another wallet.

wuyuyue:
> It seems too complicated, we can just create a widget like
> 
> ```
> const Dialog = ({ children, open }) => <dialog open={open}>{children}</dialog>
> export default () => createPortal(<Dialog />, document.querySelector('#modal'))
> ```
> and use it in other components like
> 
> ```
> render() {
>     <Dialog open={true}> This will be rendered in dialog </Dialog>
> }
> ```



> It seems too complicated, we can just create a widget like
> 
> ```
> const Dialog = ({ children, open }) => <dialog open={open}>{children}</dialog>
> export default () => createPortal(<Dialog />, document.querySelector('#modal'))
> ```
> and use it in other components like
> 
> ```
> render() {
>     <Dialog open={true}> This will be rendered in dialog </Dialog>
> }
> ```
yes, keep the open state in seperate sub-dialog seems simple. my solution is keep the open state of Dialog Container  in global context , and expose  the  showModal(), hideModal()  action for usage in event.

ashchan:
Merging this. We could refactor later.

PR #51, opened 2019-02-14
feat[neuron-ui]: add ActionFlow component draft; add import/create wa…

wuyuyue:
…llet page under settings

### DIFF
--- a/packages/neuron-ui/src/components/Settings/actionC.tsx
+++ b/packages/neuron-ui/src/components/Settings/actionC.tsx
@@ -1,0 +1,74 @@
+import React, { useRef, useContext } from 'react'
+import styled from 'styled-components'
+import SettingsContext from '../../contexts/settings'
+
+const InputDiv = styled.div`
+  input {
+    height: 30px;
+    line-height: 30px;
+    font-size: 16px;
+    margin-top: 10px;
+    width: 400px;
+  }
+`
+export default () => {
+  const settingsContext = useContext(SettingsContext)
+  const refP1: { current: HTMLInputElement | null } = useRef(null)
+  const refP2: { current: HTMLInputElement | null } = useRef(null)
+  const refName: { current: HTMLInputElement | null } = useRef(null)
+
+  const checking = () => {
+    if (!refP1.current || !refP2.current || !refName.current) return
+    if (refP1.current.value.length < 8 || refP2.current.value.length < 8) {
+      settingsContext.passwordValid = false
+      return
+    }
+    if (refP1.current.value !== refP2.current.value) {
+      settingsContext.passwordValid = false
+      return
+    }
+    if (!refName.current.value) {
+      settingsContext.passwordValid = false
+      return
+    }
+    settingsContext.passwordValid = true
+  }
+  return (
+    <>
+      <h1>set a strong password for your wallet</h1>
+      <InputDiv>
+        <div>password</div>
+        <input
+          type="password"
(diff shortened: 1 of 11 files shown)


packages/neuron-ui/src/contexts/settings.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext } from 'react'
+
+interface IWalletSettings {
+  seeds: string
+  seedsValid: boolean
+  passwordValid: boolean
+}
+
+export const initSettings: IWalletSettings = {
```
Keith-CY: If possible, I'd like to merge contexts of main components into `wallet context` cause they are paired.

packages/neuron-ui/src/components/Settings/actionC.tsx
```diff
@@ -0,0 +1,74 @@
...
+    height: 30px;
+    line-height: 30px;
+    font-size: 16px;
+    margin-top: 10px;
+    width: 400px;
+  }
+`
+export default () => {
+  const settingsContext = useContext(SettingsContext)
+  const refP1: { current: HTMLInputElement | null } = useRef(null)
```
ashchan: Variable and file names such as `actionC`, `refP1` lack meaning. Please think of names that describe their purpose and function more clearly.

packages/neuron-ui/src/components/Settings/actionD.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+
+export default () => (
```
ashchan: Actions 0 - N should not belong to `Settings`. We'll have the `Settings` component a place for the app Preferences, where we might put many panels for different settings.

packages/neuron-ui/src/components/Settings/index.tsx
```diff
@@ -1,3 +1,63 @@
...
+  .buttonGroup {
+    button {
+      height: 40px;
+      width: 140px;
+      cursor: pointer;
+    }
+  }
+`
+
+export default (props: any) => (
```
ashchan: Like said above, let's make `Settings` the preferences. So the onboard (welcome) component should be an independent part.

packages/neuron-ui/src/contexts/settings.ts
wuyuyue: we should merge some, or we would have too much wrapper on context in withProviders. it is relative with wallet, although it is setting/creating process info.    I would move it first

ashchan:
Note we can accept this PR and refactor later.

packages/neuron-ui/src/contexts/settings.ts
ashchan: Yes if we feel there's too many contexts, just refactor and adjust.

(BTW please pay attention to the commit message. Use `feat(neuron-ui)` instead of `feat[neuron-ui]`.

packages/neuron-ui/src/components/Settings/index.tsx
wuyuyue: ok, I understand. previewly  I create the welcome component. I thought you want to place create/import function in setting from your prototype draft. I would recreate welcome component

packages/neuron-ui/src/components/Settings/actionC.tsx
wuyuyue: yes, I would confirm each action and refactor it in fulture.

PR #60, opened 2019-02-18
feat(neuron-wallet): add ipc channels

Keith-CY:
re [#29](https://github.com/nervosnetwork/neuron-internal/issues/29)


packages/neuron-wallet/src/utils/const.ts
```diff
@@ -1,9 +1,29 @@
...
+  GetTipBlockNumber = 'getTipBlockNumber',
+  GetLocalNodeId = 'getLocalNodeId',
+  GetNetwork = 'getNetwork',
+  GwitchNetwork = 'switchNetwork',
+
+  // Wallet
+  CreateWallet = 'createWallet',
+  ImportWallet = 'importWallet',
+  ExportWallet = 'exportWallet',
+  SwitchAccount = 'switchAccount',
```
ashchan:
```suggestion
  SwitchWallet = 'switchWallet',
```

packages/neuron-wallet/src/IPCChannel.ts
```diff
@@ -1,34 +1,179 @@
...
+        status: 1,
+        result: 'wallets',
+      })
+    }, 1000)
+  })
+  /**
+   * @name SwitchWallet
+   * @description channel to switch wallet
+   */
+  ipcMain.on(IPCChannel.SwitchAccount, (e: Electron.Event) => {
```
ashchan:
```suggestion
  ipcMain.on(IPCChannel.SwitchWallet, (e: Electron.Event) => {
```

packages/neuron-wallet/src/IPCChannel.ts
```diff
@@ -1,34 +1,179 @@
...
+    console.info('switch wallet')
+    setTimeout(() => {
+      e.sender.send(IPCChannel.SwitchAccount, {
```
ashchan:
```suggestion
      e.sender.send(IPCChannel.SwitchWallet, {
```

packages/neuron-wallet/src/IPCChannel.ts
```diff
@@ -1,34 +1,179 @@
...
+          addr: 'wallet addr',
+        },
       })
     }, 1000)
   })
-
-  // channel to get live cells
-  ipcMain.on(IPC_CHANNEL.GET_LIVE_CELL, (e: Electron.Event, ...args: string[]) => {
-    e.sender.send(IPC_CHANNEL.GET_LIVE_CELL, args)
+  /**
```
ashchan: Please leave a blank line between each `on` call.

packages/neuron-ui/src/utils/ipc.ts
```diff
@@ -22,10 +22,10 @@ export const ipcRenderer = (() => {
 
 const getLiveCell = (outpoint: any) => ipcRenderer.send('getLiveCell', outpoint)
 const getCellsByTypeHash = (typeHash: string) => {
-  ipcRenderer.send(IPC_CHANNEL.GET_CELLS_BY_TYPE_HASH, typeHash)
+  ipcRenderer.send(IPCChannel.GetCellsByTypeHash, typeHash)
 }
 const sendCapacity = (addr: string, capacity: string) => {
```
ashchan: Can we use `address` instead of `addr`? Same for some objects that have `{ addr: ... }`.

packages/neuron-ui/src/components/Providers/index.tsx
```diff
@@ -6,7 +6,7 @@ import ModalContext, { initModal, modalReducer, MODAL_ACTION_TYPES } from '../..
 import SettingsContext, { initSettings } from '../../contexts/settings'
 
 import { ipcRenderer } from '../../utils/ipc'
-import { IPC_CHANNEL } from '../../utils/const'
+import { IPCChannel } from '../../utils/const'
```
ashchan: Discussion: is it better to name it `Channel` instead of `IPCChannel`. First: the UI app might not have to know it's an `IPC` channel (which is implementation detail), second: we know `Channel` means communication mechanism and the `IPC` prefix is somewhat unnecessary.

packages/neuron-ui/src/utils/ipc.ts
Keith-CY: Of course, I will update it in another PR

packages/neuron-wallet/src/IPCChannel.ts
Keith-CY: updated

packages/neuron-ui/src/components/Providers/index.tsx
Keith-CY: I will update it in another PR

PR #62, opened 2019-02-19
chore: recover max length to 80, and add eslint-config-prettier to re…

Keith-CY:
…solve conflicts with eslint and


ashchan:
Merging but a little worry: isn't 80 char length too restrict?

PR #63, opened 2019-02-19
add layouts(Default one and Empty one) in container; move home compon…

wuyuyue:
…ent logic to route; add unAuth view for routers


duanyytop:
Commit log should follow the rule of [conventionalcommits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/)

ashchan:
Close this in favor of #66.

PR #65, opened 2019-02-19
feat(neuron-ui): add stateless dialog

Keith-CY:
This dialog could be used in this method:

```javascript
<Dialog open>
    <div>My Dialog Content</div>
</Dialog>
```

And this is a rough one, we should add more interaction like `backdrop` click or sth on it.


ashchan:
Looks good. Can delay the polish work until we really need to use the modal dialog.

ashchan:
@Keith-CY One question: does it make sense to move this from `components` to `widgets` folder?

PR #70, opened 2019-02-20
feat: add pages of token transfer and receive

duanyytop:


packages/neuron-ui/package.json
```diff
@@ -45,7 +45,9 @@
     "react-router-dom": "^4.3.1",
     "react-scripts": "2.1.5",
     "styled-components": "^4.1.3",
-    "typescript": "3.2.2"
+    "typescript": "3.2.2",
+    "qrcode.react": "0.9.3",
+    "@types/qrcode.react": "0.8.2"
```
Keith-CY: `@types/xx` should be listed in `devDependencies`

packages/neuron-ui/src/contexts/transfer.ts
```diff
@@ -0,0 +1,12 @@
+import { createContext } from 'react'
```
Keith-CY: If this really necessary to be a context?

ashchan: No. Context should be at minimum.

packages/neuron-ui/package.json
duanyytop: ok

packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -1,3 +1,32 @@
...
+  width: 300px;
+  margin: 20px 0 30px 150px;
+`
+
+const Address = styled.div`
+  color: black;
+`
+
+const Receive = () => {
+  const address: string = '0xcf078d66b3614C4c32B018ceF9100A39FaE7DC0D'
```
ashchan: For now please use a type hash (`H256`) for dummy address.

packages/neuron-ui/src/contexts/transfer.ts
```diff
@@ -0,0 +1,12 @@
...
+
+interface ITransfer {
```
ashchan: Do not prefix `I` for interfaces.

packages/neuron-ui/src/contexts/transfer.ts
ashchan: We don't need a Transfer context.

We don't need context for every feature to add. :P

duanyytop: If I want transfer data to be remembered, how to do ?

packages/neuron-ui/src/components/Receive/index.tsx
duanyytop: ok

packages/neuron-ui/src/contexts/transfer.ts
duanyytop: ok, I referenced settings.ts and I will rename it.

packages/neuron-ui/src/contexts/transfer.ts
duanyytop: ok

ashchan: The component should use local state/property to handle that. When any writings happen, they will pass through IPC channel to the electron app, and our global wallet gets updated automatically.

We'll implement viewing the wallet context as heart of the UI layer. At the moment keep in mind we just want to mock the UI flow.

duanyytop: ok.I get it

PR #71, opened 2019-02-20
feat(neuron-wallet): add os notification when send capacity and send …

Keith-CY:
…transaction

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "lint": "lerna run lint",
     "test": "lerna run --parallel test",
     "start:ui": "lerna run --scope=@nervosnetwork/neuron-ui start",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "lint": "lerna run lint",
     "test": "lerna run --parallel test",
     "start:ui": "lerna run --scope=@nervosnetwork/neuron-ui start",
-    "start:wallet": "lerna run --scope=@nervosnetwork/neuron-wallet start",
+    "start:wallet": "lerna run --scope=@nervosnetwork/neuron-wallet watch",
```
ashchan: What's the difference between `start` and `watch`?

Keith-CY: `watch` will restart `neuron-wallet` once it changed.

ashchan: I see. 👍

PR #76, opened 2019-02-20
feat(neuron-ui): add Typography Widget

Keith-CY:
Disallow text selection on UI


Keith-CY:
Global CSS is the simplest way to override 3rd-party components

ashchan:
Then let's use CSS for disabling text selection to keep it dead simple. We could add this when it's really necessary, or CSS rule causes any issue.

PR #79, opened 2019-02-21
feat: add about us dialog

xiaoluSF:


packages/neuron-wallet/src/menu.ts
```diff
@@ -1,15 +1,27 @@
...
+}
+
 const menuTemplate = [
   {
     label: app.getName(),
     submenu: [
       {
         role: 'about',
+        click: () => {
+          dialog.showMessageBox(options)
```
ashchan: Use `app.showAboutPanel()` to show the default about dialog on macOS.

xiaoluSF: On macOS, `role: 'about'` priority is higher than `click:()`.
Win and Linux is the opposite.

ashchan: Good to know that! It's great! 👍

PR #80, opened 2019-02-21
feat: add package naming conventions

xiaoluSF:
Generally, electron-builder set default naming conventions on three platforms and different from each other.
And I'm not found common naming conventions, like `* ${productName}-${version}.${ext}` `${productName}-Setup-${version}-${arch}.${ext}`  and so on.
So, I think we can decide what we should custom package name.

### DIFF
--- a/packages/neuron-wallet/electron-builder.yml
+++ b/packages/neuron-wallet/electron-builder.yml
@@ -1,7 +1,7 @@
 appId: com.nervos.neuron
 copyright: Copyright (C) 2019 Nervos Foundation. All rights reserved
 productName: Neuron
-artifactName: "${productName}-${version}.${ext}"
+
 
 asar: false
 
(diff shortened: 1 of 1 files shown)


packages/neuron-wallet/electron-builder.yml
```diff
@@ -1,6 +1,7 @@
 appId: com.nervos.neuron
 copyright: Copyright (C) 2019 Nervos Foundation. All rights reserved
 productName: Neuron
+artifactName: "${productName}-${version}.${ext}"
```
ashchan: I suggest we also include platform in the name so users know which file to download for their OS.

```suggestion
artifactName: "${productName}-${version}-${os}.${ext}"
```

ashchan: BTW for Linux and Windows, it might also be a good idea to append `${arch}` as the last part (after `${os}`).

xiaoluSF:
Nsis installer contains both archs, so Windows's package may can not show the ${arch}.

ashchan:
> Nsis installer contains both archs, so Windows's package may can not show the ${arch}.

Got it. Good for now.

PR #81, opened 2019-02-21
feat(ui): add history view

cezres:


packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -1,3 +1,86 @@
...
+  hash: string
+}
+
+const FooterDiv = styled.div`
+  height: 80px;
+`
+
+const formatterDate = (date: Date) => {
+  const y = date.getFullYear()
+  const m = date.getMonth()
```
Keith-CY: Time in JS is a little tricky, `const m = date.getMonth() + 1`

packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -1,3 +1,86 @@
...
+const History = () => {
+  const walletAddress = '0x22f5Bd068FE2F7bEd72D8c9c8CC22e872AD48Bb9'
+  const pageSize = 14
+  const [page, setPage] = useState(0)
+  const [list, setList] = useState(requestHistory(walletAddress, page, pageSize))
+  const total = 200
+
+  useEffect(() => {
+    setList(requestHistory(walletAddress, page, pageSize))
+  }, [page])
```
Keith-CY: It's ok but not good, once the request failed, the list will not be updated while the page is updated.

But leave it now, the effect will be moved to top container.

packages/neuron-ui/src/components/History/tablePagination.tsx
```diff
@@ -0,0 +1,99 @@
...
+    let idx = page
+    if (idx < 2) {
+      idx = 2
+    } else if (idx >= totalPage - 2) {
+      idx = totalPage - 1 - 2
+    }
+    pages = [-2, -1, 0, 1, 2].map(i => i + idx)
+  }
+
+  const callback = (idx: number) => {
```
Keith-CY: `callback` is not a good name

PR #88, opened 2019-02-22
feat(neuron-ui): pass content props from MainContent container to spe…

Keith-CY:
…cified main content components

For example, sendCapacity in MainContent could be used in Wallet Detail.

### DIFF
--- a/packages/neuron-ui/src/components/Router/RoutesWithProps.tsx
+++ b/packages/neuron-ui/src/components/Router/RoutesWithProps.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+import { Route } from 'react-router-dom'
+import { CustomRoute } from '.'
+
+/**
+ * @name RoutesWithProps
+ * @description with this Component, props from parent will be passed to component in Route
+ */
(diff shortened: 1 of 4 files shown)


Keith-CY:
1. methods in context will be used everywhere for convenience. If the method changes the state, it will be a mess. btw I'd like to keep context simple.

2. this is an example, we may also add error messages on the top, and they don't contexts.

3. for now, ipc is some kind of request module, it shouldn't be recognized by components. I've added `sendCapacity` in containers but left `on` listeners in Providers, the Providers component should be categorized as a container before but I've forgortten. if this design is OK I'll move it.

ashchan:
> 1. methods in context will be used everywhere for convenience. If the method changes the state, it will be a mess. btw I'd like to keep context simple.

Exposing interfaces through several simple contexts need not necessarily to be complex. But having all wallet/chain methods across components would bring UI/App logic coupling.

> 3\. the Providers component should be categorized as a container before but I've forgortten. if this design is OK I'll move it.

Looks good to me.

PR #98, opened 2019-02-25
refactor: wizard on if there is wallet in context

wuyuyue:


packages/neuron-wallet/package.json
```diff
@@ -12,7 +12,7 @@
   "main": "dist/main.js",
   "license": "MIT",
   "scripts": {
-    "start": "tsc && electron .",
+    "start": "tsc && cross-env NODE_ENV=development electron .",
```
ashchan: Please don't alter start/run task scripts with other feature/updates. It has cause too much issue and is hard to revert to trouble shooting.

packages/neuron-ui/src/components/WalletWizard/index.tsx
```diff
@@ -24,7 +28,7 @@ export default (props: any) => {
   const wallet = useContext(WalletContext)
   useEffect(() => {
     const content = document.querySelector('.main-content')
-    if (content) {
+    if (content && !wallet) {
```
Keith-CY: is `content` necessary here? the wizard is mounted in it.

packages/neuron-ui/src/components/WalletWizard/index.tsx
```diff
@@ -7,10 +7,14 @@ import ImportWallet from './importWallet'
 import CreateWallet from './createWallet'
 
 const Wizard = styled.div`
+  .full-screen & {
+    background-color: white;
+    width: 100vw;
```
Keith-CY: better to use `width: 100%`, `width: 100vw` will overflow when scroll bar exists.

packages/neuron-ui/src/containers/Providers/index.tsx
```diff
@@ -20,13 +20,14 @@ const withProviders = (Comp: React.ComponentType) => (props: React.Props<any>) =
...
     ipc.asw()
   }, [])
-
   ipcRenderer.on('ASW', (_e: any, args: Response<any>) => {
-    setWallet({
-      ...wallet,
-      name: 'asw',
-      wallet: args.result,
-    })
+    if (wallet) {
```
Keith-CY: IMO, wallet should not be `null`, we can set it with initial values, by that much validation could be omitted.

packages/neuron-ui/src/components/WalletWizard/index.tsx
wuyuyue: actually this is your code if(content)  when you do the refactor . I think without it, typescript would remind you error

packages/neuron-ui/src/components/WalletWizard/index.tsx
wuyuyue: yes, it would have scroll problem. it is not under <MainContent>. I could use 100% for width and height. I would change it

packages/neuron-ui/src/components/WalletWizard/index.tsx
Keith-CY: My fault, the pr collapsed following code, it is used to add className.

packages/neuron-ui/src/containers/Providers/index.tsx
wuyuyue: but I am not sure, what is the inital values when user first open app if it is not null

ashchan: Good question. Checking null is annoying. Meanwhile ‘wallet.wallet’ is strange. 

Something like ‘wallet.isPersisted’ might be good? Only when this returns true we know it’s a wallet that’s already on disk.

Keith-CY: Maybe `!wallet.address` or `!wallet.name` is enough, `wallet.wallet` will be refactored once the structure of the wallet is determined.

ashchan: @Keith-CY While `!wallet.address` or `!wallet.name` can deduce to the conclusion that this might not be an empty wallet, it's not proper from the usage (what if we allow a wallet to have no name in the future?). So it's simpler to just have a more meaningful method/property, even if that's only a wrap around `!wallet.address`.

👍 on the pending `wallet.wallet` refactoring.

wuyuyue: temporarily agree with @Keith-CY to

1.  move wallet.wallet.address to wallet.address (remove Checking null); 

1. MainContent style as "width: 100%; height: 100%;" for wizard layout; 

demo the process of creation wallet when run the app for first time( set the wallet.address as empty)

packages/neuron-ui/src/components/WalletWizard/actionC.tsx
```diff
@@ -19,14 +19,15 @@ export default () => {
...
-    if (refP1.current.value.length < 8 || refP2.current.value.length < 8) {
+    if (refP1.current.value.length < 2 || refP2.current.value.length < 2) {
       settingsContext.passwordValid = false
       return
     }
     if (refP1.current.value !== refP2.current.value) {
       settingsContext.passwordValid = false
       return
     }
+    settingsContext.name = refName.current.value
```
Keith-CY: updates on contexts should occur in Providers, the contexts are just the state of Providers.

ashchan:
@Keith-CY @wuyuyue Is this ready to merge?

Keith-CY:
> @Keith-CY @wuyuyue Is this ready to merge?

Merge now, some refactoring will be implemented in other PRs

PR #101, opened 2019-02-26
Simplify and clean up tasks

ashchan:
- [x] Merge/clean up start scripts so we can use a single task script to get the whole app (both `neuron-wallet` & `neuron-ui` at the moment) up and running. Do not have `run`, `watch`, `start` etc.
  - [x] The react app (`neuron-ui`) should be able to watch and refresh automatically.
  - [x] It's **OK** if the electron app (`neuron-wallet`) cannot hot load or run in watch mode. 
  - [x] Ideally to only have `start` and `start:ui` to run the whole app and the react UI app only respectively.
- [x] Make sure binary packaging still works.
- [x] Make sure every one understand that changing the start script has risk to break packaging, running the app fully.
- [x] Make browser for `neuron-ui` to open by default.


packages/neuron-ui/package.json
```diff
@@ -12,10 +12,10 @@
   "license": "MIT",
   "scripts": {
     "start": "cross-env BROWSER=none react-app-rewired start",
-    "build": "react-app-rewired build",
-    "test": "react-app-rewired test --env=jsdom --color",
-    "eject": "react-scripts eject",
```
ashchan: Remove `eject` to avoid anyone running and ejecting the CRA app accidentally.

ashchan:
@Keith-CY Didn't bother making the electron run in watch mode. Do that later when we have free time.

packages/neuron-wallet/package.json
```diff
@@ -13,18 +13,17 @@
   "main": "dist/main.js",
   "license": "MIT",
   "scripts": {
-    "start": "tsc && electron .",
+    "start": "cross-env NODE_ENV=production electron .",
+    "start:dev": "tsc && cross-env NODE_ENV=production electron .",
```
Keith-CY: We've used `isPackaged` to distinguish build for dev or prod.

ashchan: Hold on a second, will update this.

packages/neuron-wallet/src/main.ts
```diff
@@ -28,15 +25,19 @@ function createWindow() {
...
+      devTools: isDevelopmentMode,
     },
   })
 
   windowState.manage(mainWindow)
 
   Menu.setApplicationMenu(menu)
 
-  mainWindow.loadURL(app.isPackaged ? ENTRY.PROD : ENTRY.DEV)
+  const ENTRY = {
```
Keith-CY: `global` variables usually defined at the top

ashchan: I'd prefer to move it to where it's used (if there's only one place that uses it).

ashchan: In fact in this case, I believe we should simply delete `ENTRY`, and use the two URLs directly.

Keith-CY: OK, make it simple for now.

PR #106, opened 2019-02-27
feat: add i18n solution react-intl demo 

wuyuyue:


packages/neuron-ui/src/components/App/index.tsx
```diff
@@ -1,7 +1,13 @@
 import React from 'react'
 import withProviders from '../../containers/Providers'
+import I18nProvider from '../../containers/I18nProvider'
+
 import Router from '../Router'
 
-const App = () => <Router />
+const App = () => (
+  <I18nProvider>
```
Keith-CY: Provider could be added in `Providers`

ashchan: Could or should? If it's a rule then make it explicit.

Keith-CY: The provider should be placed in `Providers`

PR #107, opened 2019-02-27
fix: linux package error

xiaoluSF:


packages/neuron-wallet/package.json
```diff
@@ -22,7 +22,7 @@
     "package-mwl": "electron-builder -mwl",
     "package-win": "electron-builder --win --x64",
     "package-mac": "electron-builder --mac",
-    "package-linux": "electron-builder --linux",
+    "package-linux": "./script/installrpm.sh && electron-builder --linux",
```
classicalliu: This script seems only works for macOS, what will happen if I run this on windows and linux ?

ashchan: How about this?

* Add the full `package` script to `/script/package`.
* Only support macOS and Linux (let's forget running this on Windows).
* From that script, check if `rpm` is installed, if not, prompt the user to install.
* Allow passing `PLATFORM=linux` (just example) to that script, so we can leave only one `package` script in `package.json`.

packages/neuron-wallet/script/package.sh
```diff
@@ -0,0 +1,64 @@
...
+        mwl)
+        electron-builder -mwl
+        ;;
+        *)
+        echo "Please add platform parameters like 'mac' 'win' 'linux' 'mwl' ."
+        ;;
+        esac
+}
+
+function checkandinstallrpm() {
```
classicalliu: Using snake_case to split function name

packages/neuron-wallet/script/package.sh
```diff
@@ -0,0 +1,64 @@
...
+                    yumdnf="dnf"
+                fi
+                echo "Please run '$yumdnf install -y rpm' to intall rpm"
+                ;;
+                *)
+                exit 1
+                ;;
+            esac
+        else
+            echo "Other OS: ${OS}"
```
classicalliu: Reminder user to install rpm

packages/neuron-wallet/script/package.sh
```diff
@@ -0,0 +1,64 @@
...
+        fi
+    else
+        package $1
+    fi
+}
+
+cd packages/neuron-wallet/
+getplatform $1
```
classicalliu: Insert new line at the end of file

packages/neuron-wallet/script/package.sh
```diff
@@ -0,0 +1,64 @@
...
+        ;;
+        esac
+}
+
+function checkandinstallrpm() {
+    echo "Check rpm..."
+    if ! [ -x "$(command -v rpm)" ] ; then 
+        OS=`uname -s`
+        if [ ${OS} == "Darwin"  ];then
+            echo "Please run 'brew install rpm' "
```
classicalliu: Also add `to install rpm` at the end as following

packages/neuron-wallet/script/package.sh
```diff
@@ -0,0 +1,64 @@
...
+        electron-builder --win --x64
+        ;;
+        linux)
+        electron-builder --linux
+        ;;
+        mwl)
+        electron-builder -mwl
+        ;;
+        *)
+        echo "Please add platform parameters like 'mac' 'win' 'linux' 'mwl' ."
```
ashchan: I suggest by default we build all. That means `mwl` is not required but set as default, so when we run it without any parameter just build all three platforms.

classicalliu:
And update README to notice developers this command only works in macOS and linux.

ashchan:
@classicalliu Please check the updated script and see if it's good to go.

packages/neuron-wallet/script/package.sh
```diff
@@ -0,0 +1,64 @@
...
+        mac)
+        electron-builder --mac
+        ;;
+        win)
+        electron-builder --win --x64
+        ;;
+        linux)
+        electron-builder --linux
+        ;;
+        mwl)
```
classicalliu: This option is redundant, please remove it.

classicalliu:
> @classicalliu Please check the updated script and see if it's good to go.

Script works well.

PR #111, opened 2019-02-28
refactor: extract env and encapsulate isDevMode

ashchan:


packages/neuron-wallet/src/main.ts
```diff
@@ -37,7 +36,7 @@ function createWindow() {
     DEV: 'http://localhost:3000',
     PROD: `file://${path.join(__dirname, '../ui/index.html')}`,
   }
-  mainWindow.loadURL(isDevelopmentMode ? ENTRY.DEV : ENTRY.PROD)
+  mainWindow.loadURL(env.isDevMode ? ENTRY.DEV : ENTRY.PROD)
```
Keith-CY: I'm not sure if we can place the `url` in `env`

ashchan: > I'm not sure if we can place the `url` in `env`

That looks a good idea to me. Please hold and I'll change that as well.

PR #113, opened 2019-02-28
chore: disable electron app node integration (WIP)

ashchan:
Disable `nodeIntegration` for a more secure setup (note Electron 5 will set this to false as default).


ashchan:
Closing this until we know how to make ipc and other work across electron/renderer.

PR #114, opened 2019-02-28
chore: modify history ui

cezres:


Keith-CY:
Why not to use pagination?

packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -36,19 +47,38 @@ const transactionsToHistory = (transactions: Transaction[]) =>
...
+
+  useEffect(() => {
+    const newItems = transactionsToHistory(chain.transactions.items)
+    setItems(page === 0 ? newItems : items.concat(newItems))
+    setLoading(false)
+  }, [chain])
+
+  const onMore = () => {
+    const element = document.getElementById('root') as HTMLElement
+    const scrollBottom = element.scrollHeight - element.offsetHeight - element.scrollTop
```
ashchan: Doesn't look a good idea to manipulate DOM. We should avoid this as much as possible.

PR #120, opened 2019-03-01
feat(neuron-ui): enable i18n

Keith-CY:


Keith-CY:
> No need to allow user to switch language on the fly.
> 
> Should detect system language from electron app (‘app.getLocale’) and apply that to both neuron-wallet and wallet-ui. If system language is not supported fallback to English.

Updated, removed language switch from Genera Setting, and add setLanguage in Providers to listen to the command from neuron-wallet.

PR #123, opened 2019-03-04
feat(neuron-wallet): push transaction history to neuron-ui

Keith-CY:
- push history on initialization, and monitoring TBD.
- `redirect` listener was added in `Router` in a callback, optimization for re-render will be done in another commit.

### DIFF
--- a/packages/neuron-wallet/src/channel.ts
+++ b/packages/neuron-wallet/src/channel.ts
@@ -248,7 +248,7 @@ const listenToChannel = () => {
 export const setLanguage = (win: Electron.BrowserWindow, lng: string) => {
   win.webContents.send(Channel.SetLanguage, lng)
 }
-export const sendTransactions = (win: Electron.BrowserWindow, pageNo: number, pageSize: number) => {
+export const sendTransactionHistory = (win: Electron.BrowserWindow, pageNo: number, pageSize: number) => {
   win.webContents.send(Channel.GetTransactions, {
     status: 1,
     result: {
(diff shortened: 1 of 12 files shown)


packages/neuron-wallet/src/main.ts
```diff
@@ -20,6 +20,8 @@ const initUILayer = (win: BrowserWindow) => {
   })
   // switch language according to locale
   setLanguage(win, app.getLocale())
+  // send transaction history on window load
```
ashchan: Don't feel the comment is necessary. It just duplicates the function call - the function name already describes what it does well.

So is the switch language comment.

packages/neuron-wallet/src/channel.ts
```diff
@@ -243,5 +248,19 @@ const listenToChannel = () => {
 export const setLanguage = (win: Electron.BrowserWindow, lng: string) => {
   win.webContents.send(Channel.SetLanguage, lng)
 }
+export const sendTransactions = (win: Electron.BrowserWindow, pageNo: number, pageSize: number) => {
```
ashchan: The name `sendTransactions` gives us a impression that this is to send transactions to the node/blockchain.

Keith-CY: renamed to `sendTransactionHistory`

packages/neuron-wallet/src/main.ts
Keith-CY: comments removed

Keith-CY:
updated, removed useless comments and rename `sendTransactions` to `sendTransactionHistory`

PR #124, opened 2019-03-05
fix: fix settings navigation error

duanyytop:


packages/neuron-ui/src/components/Settings/TabBar.tsx
```diff
@@ -27,15 +27,15 @@ const TabBar = (props: any) => {
...
-  useEffect(() => {
-    props.history.push(tabs[0])
-  }, [])
-
   const handleAction = (index: number) => {
     setSelectedIndex(index)
     props.history.push(tabs[index])
   }
 
+  useEffect(() => {
```
Keith-CY: Will this navigate to tab 0 every time the page loads, even we intend to other tabs?

duanyytop: When coming in settings page firstly, neuron would show general settings page (tab 0).

 If you want to navigate to other tabs, you need to click tab titles.

Keith-CY: 1. if handleAction is to `props.history.push(tabs[index])`, why not set the link to setting as `tabs[index]`

2. we may need to other tabs directly, for example, when we clicked on the `network info` in the header, it may show the setting of the chain.

duanyytop: If settings page (include general, wallets and network) is active, clicking tab titles will not call useEffect method. So when you want to navigate to other tabs, tab 0 will not always display.

Keith-CY: Yes, when we are at setting page, the tab bar will not be re-rendered on switching. But if we are at other pages, like wallet detail, and on clicking on the network in the header, it navigates us to wallet-setting, at that time, the tab bar is rendered and useEffect works to navigate us to general-setting.

duanyytop: Ok. I see.

packages/neuron-ui/src/containers/Sidebar/index.tsx
```diff
@@ -44,23 +44,48 @@ const SidebarAside = styled.div`
...
-  ['Wallet', IconWallet],
-  ['Send', IconSend],
-  ['Receive', IconReceive],
-  ['History', IconHistory],
-  ['Addresses', IconAddresses],
-  ['Settings', IconSettings],
+  {
+    name: 'Wallet',
+    path: 'Wallet',
+    Icon: IconWallet,
```
Keith-CY: replace `Icon` with `icon`

duanyytop: OK, I forget.

PR #125, opened 2019-03-05
fix: fix can not run on Linux

xiaoluSF:


packages/neuron-ui/package.json
```diff
@@ -31,7 +31,8 @@
     "not op_mini all"
   ],
   "dependencies": {
-    "@nervosnetwork/ckb-sdk-core": "^0.0.1-alpha.1",
+    "@nervosnetwork/ckb-sdk-core": "^0.0.1-alpha.3",
+    "@nervosnetwork/ckb-sdk-rpc": "0.0.1-alpha.3",
```
Keith-CY: ckb-sdk-core includes ckb-sdk-rpc, so here it's unnecessary.

xiaoluSF:
> update `@nervosnetwork/ckb-types` as well

OK

PR #127, opened 2019-03-05
fix: get locale from electron app

ashchan:
Should call the method to fetch the locale.


ashchan:
@classicalliu This is the new bug that caused your packaged app not loading Chinese - I guess.

PR #132, opened 2019-03-06
feat: add wallet store

BaojunCZ:
I will take store ipc to import wallet,delete wallet ipc when confirm data structure


packages/neuron-ui/src/components/Settings/Wallets.tsx
```diff
@@ -52,55 +54,42 @@ const UpdateWallet = styled(WalletButton)`
...
-    setWalletsState(temps)
+    const walletName = `wallet${Math.ceil(Math.random() * 100)}`
+    UILayer.renameWalletStore(walletName, walletNames.name[walletSelected])
   }
 
   return (
     <ContentPanel>
-      {walletsState.map(wallet => {
-        if (walletSelected === wallet) {
+      {walletNames.name.map(wallet => {
```
ashchan: `Settings/Wallets` should keep a copy of wallet objects, not names. We'll need to access the wallet object for more information.

packages/neuron-ui/src/containers/Providers/index.tsx
```diff
@@ -89,11 +91,39 @@ const withProviders = (Comp: React.ComponentType) => (props: React.Props<any>) =
     },
   )
 
+  UILayer.on(Channel.GetWalletNameListStore, (_e: Event, args: Response<any>) => {
```
ashchan: UI layer should not have knowledge of wallet store or wallet name store. It just ask for a list of wallet objects. The store is an implementation detail in the `neuron-wallet` module.

packages/neuron-ui/src/containers/Providers/index.tsx
```diff
@@ -89,11 +91,39 @@ const withProviders = (Comp: React.ComponentType) => (props: React.Props<any>) =
...
+    if (args.status) {
+      setWalletNames({
+        name: args.result,
+      })
+    }
+  })
+
+  UILayer.on(Channel.SaveWalletStore, (_e: Event, args: Response<any>) => {
```
ashchan: As above, these interfaces should pass a wallet object. So better to have `SaveWallet`, `DeleteWallet` etc.

packages/neuron-ui/src/contexts/WalletNames.ts
```diff
@@ -0,0 +1,11 @@
+import { createContext } from 'react'
+
+export interface WalletNames {
```
ashchan: As pointed out above, we should have access to `wallets` list, not names from UI.

packages/neuron-wallet/src/channel.ts
```diff
@@ -2,6 +2,11 @@ import { ipcMain, Notification } from 'electron'
 import { Channel } from './utils/const'
 import { cell, transactions, transactionCount } from './mock'
 import asw from './wallets/asw'
+import WalletStore, { Wallet } from './store/WalletStore'
+import OtherStore from './store/otherStore'
```
ashchan: What does 'OtherStore' mean? It's not a very good name.

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -47,6 +40,13 @@ export default class WalletStore extends BaseStore {
     }
   }
 
+  getAllWallet(): Wallet[] {
```
ashchan: Should be `getAllWallets`.

packages/neuron-wallet/src/channel.ts
```diff
@@ -291,14 +289,14 @@ ipcMain.on(
  * @name saveWalletStore
  * @description channel to get wallet name list
  */
-ipcMain.on(Channel.GetWalletNameListStore, (e: Electron.Event) => {
+ipcMain.on(Channel.GetAllWalletStore, (e: Electron.Event) => {
```
ashchan: Should be `GetAllWallets` instead of `GetAllWalletStore`.

packages/neuron-wallet/test/fixture.js
```diff
@@ -0,0 +1,20 @@
+const electron = require('electron')
```
ashchan: Please write all code in TypeScript.

packages/neuron-wallet/package.json
```diff
@@ -37,9 +38,11 @@
   },
   "dependencies": {
     "@nervosnetwork/ckb-sdk-core": "^0.0.1-alpha.3",
-    "electron-window-state": "^5.0.3",
     "@types/electron-store": "^1.3.0",
+    "ava": "^1.3.1",
     "electron-store": "^2.0.0",
+    "electron-window-state": "^5.0.3",
+    "execa": "^1.0.0",
```
ashchan: Is this NPM package for writing tests? If so perhaps it should go to `devDepencencies` instead?

BaojunCZ:
remove store from ipc and ui,add wallet store test

packages/neuron-wallet/package.json
```diff
@@ -20,7 +20,8 @@
     "test": "jest --color",
     "lint": "eslint 'src/**/*.{js,ts}'",
     "package": "./script/package.sh",
-    "postinstall": "npm rebuild tiny-secp256k1"
+    "postinstall": "npm rebuild tiny-secp256k1",
+    "testStore": "ava"
```
ashchan: We don't have any script with `camelCase` names.

BaojunCZ: ok

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -0,0 +1,79 @@
...
+      throw new Error(WalletStoreError.ExistKey + walletName)
+    } else {
+      let nameList = this.getWalletNameList()
+      this.save(walletName, wallet)
+      nameList = nameList.concat(walletName)
+      this.save(keyWalletName, nameList)
+    }
+  }
+
+  getAllWallets(): Wallet[] {
```
Keith-CY: `getAllWallets = () => {}` to keep consistent with other methods.

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -0,0 +1,79 @@
...
+    const walletList: Wallet[] = []
+    const nameList = this.getWalletNameList()
+    nameList.forEach(name => walletList.push(this.getWallet(name)))
+    return walletList
+  }
+
+  getWallet(walletName: string): Wallet {
```
Keith-CY: the returned type may be null

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -0,0 +1,79 @@
...
+    return this.get(walletName)
+  }
+
+  renameWallet(newName: string, oldName: string) {
+    const wallet = this.getWallet(oldName)
+    wallet.name = newName
+    this.saveWallet(newName, wallet)
```
Keith-CY: if wallet includes name, why use two params

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -0,0 +1,79 @@
+import BaseStore from './store'
+
+const walletDBName = 'WalletDB'
+
+const keyWalletName = 'WalletName'
```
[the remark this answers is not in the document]
ashchan: It's ok to scope the db path to store implementation. But it indeed should read `env` and use different paths for different environments. This way also allow us to wipe test store anytime without affecting development mode db.

Speaking of these, we might need to use `NODE_ENV` again so `env` knows if it's running under test mode.

packages/neuron-wallet/src/store/WalletStore.ts
BaojunCZ: oldName is a index.It`s a file system nor db,so I should save a index list to find wallet.

packages/neuron-wallet/src/store/WalletStore.ts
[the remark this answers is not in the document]
Keith-CY: I'll do it.

PR #133, opened 2019-03-06
feat: add cell module

classicalliu:


packages/neuron-ui/src/containers/Sidebar/index.tsx
```diff
@@ -60,6 +61,10 @@ const walletMenuItems = [
     name: 'History',
     icon: IconHistory,
   },
+  {
```
ashchan: Think twice if an end user should have to know what cells are.

packages/neuron-ui/src/locales/zh.ts
```diff
@@ -9,6 +9,7 @@ export default {
     Send: '转账',
     Receive: '收款',
     History: '交易历史',
+    UnspentCells: '未使用Cells',
```
ashchan: Like commented above, I don't thin the wallet interface should expose cells data. For a wallet user, she/he only cares about that she/he could spend amount <= balance. No need to expose cells concept to user.

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -127,7 +128,7 @@ const listenToChannel = () => {
     setTimeout(() => {
       e.sender.send(Channel.GetCellsByTypeHash, {
         status: 1,
-        result: [cell],
+        result: ['cells'],
```
Keith-CY: Why not return response of `getUnspentCells`

classicalliu: Not used now, not important

PR #134, opened 2019-03-06
feat: add terminal with basic functionality

Keith-CY:
add basic functionality of terminal.
styling and shortcuts to be done.


packages/neuron-ui/src/containers/Sidebar/index.tsx
```diff
@@ -68,6 +69,10 @@ const walletMenuItems = [
     name: 'Settings',
     icon: IconSettings,
   },
+  {
```
ashchan: Please remove this from Sidebar as it should be a debug only feature.

Perhaps add it to the `Debug` main menu?

Also note from a production design perspective, our current Sidebar already displays all the basic features we want to expose for current stage.

Keith-CY: remove 'terminal' from sidebar, add 'terminal' label in develop menu.

PR #135, opened 2019-03-06
feat(neuron-ui): optimize pagination ui

Keith-CY:
add ellipsis in pagination

### DIFF
--- a/packages/neuron-ui/src/components/History/TablePagination.tsx
+++ b/packages/neuron-ui/src/components/History/TablePagination.tsx
@@ -2,29 +2,67 @@ import React from 'react'
 import { Pagination } from 'react-bootstrap'
 
 interface TablePaginationProps {
-  page: number
+  currentPage: number
   pageSize: number
   total: number
+  displayCount?: number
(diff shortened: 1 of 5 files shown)


packages/neuron-ui/src/components/History/TablePagination.tsx
```diff
@@ -2,29 +2,67 @@ import React from 'react'
...
     })}
-
-    <Pagination.Next />
-    <Pagination.Last />
+    <Pagination.Next disabled={currentPage === pageCount - 1} onClick={() => onChange(currentPage + 1)} />
+    <Pagination.Last onClick={() => onChange(pageCount - 1)} />
   </Pagination>
 )
 
+const TablePagination = ({ total, currentPage, pageSize, displayCount = 5, onChange }: TablePaginationProps) => {
```
ashchan: I feel this component deserves its own file under `widgets`.

Keith-CY: Yes, I'll do it in another PR

PR #138, opened 2019-03-07
feat: add Keystore interface and unit test

duanyytop:
- [x] Add Keystore interface 
- [x] Add mock data
- [x] Add Keystore test cases


packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -1,4 +1,5 @@
 import BaseStore from './store'
+import { KeyStore } from '../keys/keystore'
```
ashchan: `Keystore` is a de fact name of this concept, not `KeyStore`.

duanyytop: OK

packages/neuron-wallet/src/keys/keystore.ts
```diff
@@ -0,0 +1,16 @@
+export interface Master {
```
ashchan: `Master/Child` is too generic names (since we don't have namespaces to scope them).

Will `MasterKey/ChildKey` be better? (just an idea, we should think of a good name.)

packages/neuron-wallet/src/keys/key.ts
```diff
@@ -0,0 +1,73 @@
...
+    const keystore: Keystore = {
+      master,
+    }
+    if (derive) {
+      keystore.children = Key.searchUsedAddress()
+    }
+    return new Key(keystore, mnemonic)
+  }
+
+  private static searchUsedAddress() {
```
ashchan: * If the function returns a collection of addresses, it should use `searchUsedAddresses` instead of `searchUsedAddress`.
* Are `address` and `child` of the same concept? If not, then `address` in the name might not be appropriate.

packages/neuron-wallet/src/keys/key.ts
```diff
@@ -0,0 +1,73 @@
...
+    const children: Child[] = []
+    for (let depth = 0; depth < 4; depth++) {
+      // TODO: refactor search logic to reduce loop
+      for (let index = 0; index < 4; index++) {
```
ashchan: For CKB, we can assume that the first paths of `m / purpose' / coin_type' / account' / change` is fixed:

m/44'/[TBD]'/0'.

`coin_type` will be fixed. We can use a constant that assigned a temporary value for now. (But we'll have to decide if mainnet/testnet uses different coin_types later).

For `change` path, we can assume that we'll use `0` for external (receiving) addresses and `1` for internal (change) addresses for now, just as Bitcoin.

packages/neuron-wallet/src/keys/keystore.ts
duanyytop: I think `Master/Child` are existed in `keys` folder which have showed this module is about `key`, and file name keystore also has the same effect.

packages/neuron-wallet/src/keys/key.ts
duanyytop: Yes, `Addresses` will be better. It's my fault. 

I think the public key(or address) of master and child is same concept. So address should contain master and child.

ashchan: No address is a much simpler concept. It is a `public key` (could be child key's) hash.

duanyytop: Public key and address are  one to one correspondence, and they are all generated from private key which contains master private key and child private key. 

But I think address just is one part of master and child, So searchAddress is not appropriate. 

I will think twice.

ashchan: Let's merge and update later.

packages/neuron-wallet/src/keys/key.ts
```diff
@@ -0,0 +1,73 @@
...
+    return new Key(keystore, '')
+  }
+
+  public static fromKeystoreJson = (json: string) => {
+    return new Key(JSON.parse(json), '')
+  }
+
+  getKeystore = () => this.keystore
+
+  getKeystoreJson = () => JSON.stringify(this.keystore)
```
Keith-CY: why does the `getKeystoreJson` return a string

duanyytop: Is json not a string?

Keith-CY: similar to an object

duanyytop: JSON.parse ?

Keith-CY: this keystore in JSON.stringify is already a so called json, JSON.stringify transfer it to a string, and JSON.parse recover it to an object

duanyytop: OK. I know and I will refactor later

duanyytop: @Keith-CY If I want to provide json string format, what function name will be appropriate？

Keith-CY: `getKeystore` which accepts an `encode` parameter

duanyytop: Like this ?
```
 getKeystore = (encode: boolean) => {
    return encode ? JSON.stringify(this.keystore) : this.keystore
  }
```

PR #139, opened 2019-03-07
feat(neuron-ui): setNetwork in Header

Keith-CY:


Keith-CY:
@BaojunCZ Modification related to tests is just removal of console, please have a look why it alerts `ENOENT: no such file or directory, open`

BaojunCZ:
https://github.com/nervosnetwork/neuron/pull/141

Keith-CY:
I've recovered test cases with `console` now and disabled eslint for `console`, just for not pending too long. And this rule should be enabled once the tests updated.

```
// disabled until test recovered
// "no-console": [2, {
//   "allow": ["warn", "error"]
// }]
```

packages/neuron-ui/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_NETWORKS=[{"name":"Testnet","remote":"http://localhost:8114"},{"name":"Localhost","remote":"http://localhost:8114"}]
```
ashchan: Shouldn't this belong to the `neuron-wallet` module?

And it should not be `ENV`, as we're only to have official testnet/mainnet in the future. This should be in program constants.

packages/neuron-ui/src/components/Network/index.tsx
```diff
@@ -27,28 +33,55 @@ const Span = styled.span`
...
+    <>
+      <FlexDiv>
+        <Span>
+          <Status
+            style={{
+              color: chain.network.status === NetworkStatus.Online ? 'green' : 'red',
+            }}
+          />
+        </Span>
+        <Span>{`${(chain.network.name && chain.network.name.slice(0, 30)) || chain.network.remote} - `}</Span>
```
ashchan: Looks like this should go to `Network`. For example, add a `label` function which use `chain.network.name && chain.network.name.slice(0, 30)) || chain.network.remote`.

packages/neuron-ui/src/components/NetworkEditor/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+          },
+        }),
+      tooltip: TooltipText.Name,
+      placeholder: PlaceHolder.Name,
+    },
+  ]
+
+  return (
+    <Card>
+      <Card.Header>{idx === -1 ? 'Add Network' : params.name}</Card.Header>
```
ashchan: We should use i18n for all UI texts.

packages/neuron-ui/src/components/Settings/RemoveNetworkDialog.tsx
```diff
@@ -0,0 +1,54 @@
...
+      e.stopPropagation()
+    }}
+  >
+    <Card.Header>
+      Remove Address:
+      {network.name || network.remote}
+    </Card.Header>
+    <Card.Body>
+      {`Network of name: ${network.name}, address: ${network.remote} will be removed.`}
+      {isChecked ? <p>{`It's the active network, removing it will make reconnect to ${Testnet}`}</p> : null}
```
ashchan: Just use some prompt as 'Are you sure to remove this network?'.

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -248,6 +249,26 @@ const listenToChannel = () => {
       })
     }, 1000)
   })
+
+  /**
+   * @name setNetwork
+   */
+  ipcMain.on(Channel.SetNetwork, (e: Electron.Event, network: { name: string; remote: string }) => {
+    // TODO:
```
ashchan: I suggest we persist all networks and current network to `electron-store` as JSON file. Note do not save to `keys` (wallets) folders.

PR #140, opened 2019-03-08
chore: check `yarn.lock` in CI

classicalliu:


ashchan:
Nice it fails as expected :-)

PR #141, opened 2019-03-11
refactor: wallet store test remove fs require

BaojunCZ:


packages/neuron-wallet/tests/wallet_store/save-wallet.js
```diff
@@ -15,6 +15,6 @@ walletStore.saveWallet('wallet1', {
   keystore: 'qazwsx',
 })
 
-console.log(walletStore.path())
+console.log(JSON.stringify(walletStore.getStore()))
```
ashchan: Remove all trace/debug purpose console log.

BaojunCZ: It`s for test.Putting test file into execa with electron evm, execa will execute test file and call back log.Take the log to assert

PR #143, opened 2019-03-11
refactor: add uuid as key

BaojunCZ:


packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -1,85 +1,95 @@
-import BaseStore from './store'
+import Store from 'electron-store'
+import { v4 } from 'uuid'
 import { Keystore } from '../keys/keystore'
 import env from '../env'
 
-const keyWalletName = 'WalletName'
+const encryptKey = 'Neuron'
```
Keith-CY: better to be in env

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -1,85 +1,95 @@
...
+
+const WalletIDKey = 'WalletID'
```
Keith-CY: better to be in env

packages/neuron-wallet/src/store/WalletStore.ts
BaojunCZ: ok

packages/neuron-wallet/src/store/WalletStore.ts
BaojunCZ: It is a key just used in wallet store.So I don't think it needs to go to env

packages/neuron-wallet/package.json
```diff
@@ -58,7 +60,7 @@
     "electron-devtools-installer": "^2.2.4",
     "spectron": "^5.0.0",
     "tslint": "^5.12.1",
-    "execa": "^1.0.0",
-    "@types/execa": "^0.9.0"
+    "@types/uuid": "^3.4.4",
```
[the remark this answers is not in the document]
BaojunCZ: yes, I will move it under other @types

packages/neuron-wallet/src/store/WalletStore.ts
ashchan: Let's do it this way: check env, and use different root folder for keys.

For example, for production, use `/keys` (relative to default user data folder), for development, use `/keys-dev`, for test, use `/keys-test`. This is to avoid wiping local data when running tests or affecting production data when running in development on the same machine.

BaojunCZ: ok,test is running in different folder now.I will move dev part to new folder

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -28,7 +26,7 @@ export default class WalletStore {
   constructor() {
     const idOptions: Options = {
       name: env.walletIDName,
-      encryptionKey: encryptKey,
+      encryptionKey: env.storeEncryptKey,
```
ashchan: Do we really need to encrypt files?

package.json
```diff
@@ -28,7 +28,7 @@
 
     "bootstrap": "lerna bootstrap && lerna link",
     "lint": "lerna run lint",
-    "test": "lerna run teststore && lerna run --parallel test",
+    "test": "lerna run --parallel test && lerna run teststore",
```
ashchan: I think it's better to call `teststore` from `neuron-wallet` package instead.

BaojunCZ: ok

packages/neuron-wallet/src/store/WalletStore.ts
BaojunCZ: Encryption or not has little effect.I think better than nothing

PR #145, opened 2019-03-13
refactor: add BIP44 protocol and update researching hd wallet used address

duanyytop:


packages/neuron-wallet/src/keys/tool.ts
```diff
@@ -0,0 +1,83 @@
+import { BIP32 } from 'bip32'
+import { Child } from './keystore'
+
+// BIP44: https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+const PURPOSE = "44'"
```
Keith-CY: these constants could be type of enum

packages/neuron-wallet/src/keys/tool.ts
```diff
@@ -0,0 +1,83 @@
...
+// 360 is tentative value
+const COIN_TYPE_TESTNET: string = "360'"
+const ACCOUNT = "0'"
+// external address
+const CHANGE = 0
+
+export default class Tool {
+  private static root: BIP32
+
+  public static searchUsedChildKeys(root: BIP32) {
```
Keith-CY: `static` method cannot access to `this`, and this should be alerted by `eslint`

packages/neuron-wallet/src/keys/tool.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  private static getAddressFromHDIndex(index: number): string {
+    const path = Tool.getPath(index)
+    const { publicKey } = this.root.derivePath(path)
+    return this.getAddressFromPublicKey(publicKey.toString('hex'))
+  }
+
+  // minUnusedIndex should be big number, 100 is convenient to do test.
+  // depth is loop depth for searching used address
+  private static searchIterationForExternalAddress(
```
Keith-CY: the function name is too long

packages/neuron-wallet/src/keys/tool.ts
duanyytop: Here `Tool` replace `this` ?

packages/neuron-wallet/src/keys/tool.ts
duanyytop: OK

packages/neuron-wallet/src/keys/tool.ts
duanyytop: I think it can show what this function to do, and how long of function name is too long? Does the eslint have a specific number ？

packages/neuron-wallet/src/keys/tool.ts
Keith-CY: if `root` is a property of `Tool` instance, `Tool` class cannot access to it. You can set this method to instance method(remove `static`), or pass the instance to it.

packages/neuron-wallet/src/keys/tool.ts
Keith-CY: IMO, `searchAddress` is enough, `iteration` or `external` can be default params to this function

packages/neuron-wallet/src/keys/tool.ts
duanyytop: root is not a instance of Tool.

packages/neuron-wallet/src/keys/tool.ts
duanyytop: OK. I will think twice.

packages/neuron-wallet/src/keys/tool.ts
Keith-CY: I mean, `this.root` is to get `root` property of `Tool` instance, and `Tool` static method cannot get the instance, namely `Tool` static method cannot get `this.root`

duanyytop: OK. I see. But how can I get `root` from `Tool`? `Tool` is just a tool class, so I think it should be a static class.

Keith-CY: In this function you are going to update the `this.root`, so just pass `this`(namely the instance` in:
```
public static searchUsedChildKeys(tool: Tool, root: BIP32) {
  tool.root = root
}
```

duanyytop:
I have solved questions above.

packages/neuron-wallet/src/keys/tool.ts
```diff
@@ -2,25 +2,27 @@ import { BIP32 } from 'bip32'
...
-const PURPOSE = "44'"
-// 360 is tentative value
-const COIN_TYPE_TESTNET: string = "360'"
-const ACCOUNT = "0'"
-// external address
-const CHANGE = 0
+enum Constant {
+  PURPOSE = "44'",
+  // 360 is tentative value
+  COIN_TYPE_TESTNET = "360'",
```
Keith-CY: enum should be named in `PascalCase`

duanyytop: OK

ashchan: Isn't 'Constant' too generic a name for enum?

duanyytop: Indeed! I will refactor later.

PR #148, opened 2019-03-14
feat(neuron-ui): wallet manage ui

BaojunCZ:


packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
```diff
@@ -0,0 +1,72 @@
...
+              onChange={(e: any) => stePassword(e.currentTarget.value)}
+              isInvalid={errorMsg !== ''}
+            />
+            <Form.Control.Feedback type="invalid">{errorMsg}</Form.Control.Feedback>
+          </Col>
+        </Form.Group>
+      </Card.Body>
+      <Card.Footer className="text-muted">
+        <div
+          style={{
```
duanyytop: Please use styled-component

packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from 'react'
+import { Card, Button, Form, Row, Col } from 'react-bootstrap'
+import { MainActions } from '../../containers/MainContent/reducer'
+import { Wallet } from '../../contexts/Wallet'
+
+const InputWalletPswDialog = ({ wallet, dispatch }: { wallet: Wallet; dispatch: any }) => {
+  const [errorMsg, setErrorMsg] = useState('')
+  const [password, stePassword] = useState('')
```
Keith-CY: setPassword

packages/neuron-ui/src/containers/Providers/index.tsx
```diff
@@ -121,7 +131,9 @@ const withProviders = (Comp: React.ComponentType<{ providerDispatch: ProviderDis
     <SettingsContext.Provider value={providers.settings}>
       <ChainContext.Provider value={providers.chain}>
         <WalletContext.Provider value={providers.wallet}>
-          <Comp {...props} providerDispatch={dispatch} />
+          <WalletsContext.Provider value={providers.wallets}>
```
Keith-CY: don't add a provider, just set wallets in settings.

packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
BaojunCZ: ok

packages/neuron-ui/src/containers/Providers/index.tsx
BaojunCZ: ok

packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
BaojunCZ: ok

packages/neuron-ui/src/widgets/InlineInput/index.tsx
```diff
@@ -7,9 +7,11 @@ export interface InputProps {
   onChange: (event: React.FormEvent<React.PropsWithoutRef<any>>) => void
   tooltip?: string
   placeholder?: string
+  inputtype?: string
```
duanyytop: inputType

packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+          <Button
+            variant="light"
+            onClick={() =>
+              dispatch({
+                type: MainActions.SetDialog,
+                payload: null,
+              })
+            }
+          >
+            Cancel
```
duanyytop: Add i18n to hard core message

packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { Card, Button, Form, Row, Col } from 'react-bootstrap'
+
+import { MainActions } from '../../containers/MainContent/reducer'
+
+const ButtonDiv = styled.div`
+  display: flex;
+  justify-content: space-between;
+`
+
+const InputWalletPswDialog = ({
```
ashchan: Please use full name for `psw`, `msg` etc.

packages/neuron-ui/src/components/Settings/Wallets.tsx
```diff
@@ -1,144 +1,134 @@
...
-    const index = temps.indexOf(walletSelected)
-    const name = prompt(t('Please enter your wallet name'), temp)
-    if (index > -1 && name != null && name !== '') {
-      temps.splice(index, 1, name)
-    }
-    setWalletsState(temps)
-  }
+  const actionItems = (index: number) => [
+    {
+      label: t('select'),
```
ashchan: @BaojunCZ @Keith-CY @duanyytop Please use `feature.section.select` style of `i18n` keys. `feature.section` could be anything that helps us distinguish where we should organize that translation. Using a single word as key would make very hard to translate and organize, and cause errors when we update. (e.g., when we want to call `Edit` on manage wallet view as `Update`, and keep as `Edit` on another view. Just an example.)

BaojunCZ: ok

packages/neuron-ui/src/components/Settings/InputWalletPswDialog.tsx
```diff
@@ -22,16 +23,24 @@ const InputWalletPswDialog = ({
...
-      dispatch({
-        type: MainActions.SetDialog,
-        payload: null,
-      })
-      handle()
-    } else {
+    if (!password) {
       setErrorMsg('Please enter password')
     }
+    checkPassword(password, (passwordValid: boolean) => {
```
Keith-CY: Is it designed that only one password for all wallets? If not, the wallet identifier should be passed too.

BaojunCZ: ok,make sense

ashchan: We didn't mention this yet, but for now let's assume we have separate password for each wallets.

In fact, all operations on a wallet should pass the wallet identifier. Although we have a `CurrentWallet` concept, that should only used to get current active wallet's identifier (or object for simple read) and the identifier should pass to low level.

packages/neuron-ui/src/components/Router/index.tsx
```diff
@@ -123,7 +123,7 @@ export const mainContents: CustomRoute[] = [
   },
   {
     name: `WalletEditor`,
-    path: `${Routes.WalletEditor}/:name`,
+    path: `${Routes.WalletEditor}/:wallet`,
```
ashchan: Are we using wallet id (`identifier`) when constructing this URL? If not (for example if we're using wallet name) please update to make sure we pass wallet identifier instead of name. This also applies to other routes and views where we need to pass or set wallet.

packages/neuron-wallet/src/mock.ts
```diff
@@ -8,25 +8,30 @@ export const transactions = Array.from({
 
 export const transactionCount = 500
 
-interface Wallet {
+export interface MockWallet {
```
ashchan: While we're leaving many wallet implementation detail for later, we should just call this `Wallet` instead of `MockWallet`.

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -258,13 +258,21 @@ export class Listeners {
...
-        status: ResponseStatus.Success,
-        result: password === '1qaz',
-      })
-      // e.returnValue = password === '1qaz'
-    })
+    return ipcMain.on(
+      Channel.CheckWalletPassword,
+      (e: Electron.Event, { walletID, password }: { walletID: string; password: string }) => {
+        let valid = false
+        wallets().forEach((value: MockWallet) => {
```
ashchan: This doesn't look right. It returns valid when any wallet has the correct password?

packages/neuron-ui/src/components/Router/index.tsx
BaojunCZ: Using the whole wallet object now.There is id in object

packages/neuron-wallet/src/channel/index.ts
BaojunCZ: `if (value.id === walletID && value.password === password)`
No,find the wallet with right id,then check password

Keith-CY: IMO, there are two steps

1. find the wallet by the id, otherwise return `wallet not found`.
2. verify its password.

packages/neuron-wallet/src/mock.ts
```diff
@@ -8,14 +8,15 @@ export const transactions = Array.from({
...
-export interface MockWallet {
+export interface Wallet {
   name: string
   id: string
   balance: number
   address: string
   publicKey: Uint8Array
   msg: string
   password: string
+  mnemonic: string
```
Keith-CY: Will mnemonic be cached?

BaojunCZ: I merged from channel/index.ts.It`s temp.Keystore have not been confirmed.

packages/neuron-wallet/src/channel/index.ts
duanyytop: This loop will not stop even if valid is true.

BaojunCZ: I've refactored this part

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -256,12 +256,22 @@ export class Listeners {
     return ipcMain.on(
       Channel.CheckWalletPassword,
       (e: Electron.Event, { walletID, password }: { walletID: string; password: string }) => {
+        let target = -1
         let valid = false
-        wallets().forEach((value: Wallet) => {
-          if (value.id === walletID && value.password === password) {
-            valid = true
+        wallets().forEach((value: Wallet, index: number) => {
```
Keith-CY: use `wallets.find(wallet => wallet.id === walletID)`, avoid using loop.

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -256,12 +256,22 @@ export class Listeners {
...
+          if (value.id === walletID) {
+            target = index
           }
         })
+        if (target === -1) {
```
Keith-CY: The error message should send to UI layer cause the dialog should be dismissed.

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -256,12 +256,22 @@ export class Listeners {
...
+          const notification = new Notification({
+            title: `Check Wallet Password`,
+            body: `Wallet not find`,
+          })
+          notification.show()
+        } else if (wallets()[target].password === password) {
```
Keith-CY: variable `valid` is redundant, just put `send success message` in this logic branch

BaojunCZ:
Refactor over

packages/neuron-ui/src/components/Settings/InputWalletPasswordDialog.tsx
```diff
@@ -27,15 +27,17 @@ const InputWalletPasswordDialog = ({ wallet, dispatch, handle }: InputPasswordPr
     if (!password) {
       setErrorMsg('Please enter password')
     }
-    checkPassword(wallet.name, password, (passwordValid: boolean) => {
-      if (passwordValid) {
+    checkPassword(wallet.name, password, (args: Response<string>) => {
+      if (args.result) {
```
Keith-CY: success or fail are designated by `args.status`, but this works

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -256,16 +256,25 @@ export class Listeners {
...
-        e.sender.send(Channel.CheckWalletPassword, {
-          status: ResponseStatus.Success,
-          result: valid,
-        })
+        const myWallet = wallets().find(wallet => wallet.id === walletID)
+        if (!myWallet) {
+          e.sender.send(Channel.CheckWalletPassword, {
+            status: ResponseStatus.Success,
+            result: false,
+            msg: 'Wallet not find',
```
Keith-CY: `Wallet not found` not `Wallet not find`. I'll fix it.

PR #150, opened 2019-03-19
refactor: i18n key

BaojunCZ:


Keith-CY:
According to the [docs](https://react.i18next.com/latest/usetranslation-hook), namespace can be a param of `useTranslation`

BaojunCZ:
get it

ashchan:
> According to the [docs](https://react.i18next.com/latest/usetranslation-hook), namespace can be a param of `useTranslation`

`t('path.to.key')` usage is fine.

BaojunCZ:
Uniform format or both？I don’t think it has much impact.

ashchan:
> Uniform format or both？I don’t think it has much impact.

Use full path format if possible. Example:

https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/menu.ts#L27

BaojunCZ:
ok,I will use full path format and make keys to be lower cases

PR #151, opened 2019-03-19
chore: update receive ui

duanyytop:


packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -10,33 +10,79 @@ declare global {
...
+
+const QRCodeModal = styled.div`
   text-align: center;
 `
 
-const Receive = () => {
+const Receive = ({ address }: { address: string }) => {
   const [t] = useTranslation()
-  const address: string = '0x0da2fe99fe549e082d4ed483c2e968a89ea8d11aabf5d79e5cbf06522de6e674'
+  const [show, setShow] = useState(false)
```
ashchan: `show` is too generic. Perhaps something like `showLargeQR`?

packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -10,33 +10,79 @@ declare global {
...
         </Card.Header>
         <Card.Body>
-          <QRCodePanel>
-            <QRCode value={address} size={256} />
+          <AddressTip>{t('Address')}</AddressTip>
+          <AddressPanel>
+            <Address onClick={() => window.clipboard.writeText(accountAddress)}>{accountAddress}</Address>
+            <OverlayTrigger placement="bottom" overlay={<Tooltip id="address-tooltip">{t('Copy address')}</Tooltip>}>
+              <CopyImage
+                src="./image/copy.png"
```
ashchan: Have we confirmed that referring resources relatively works perfectly after packaging the app?

duanyytop: I will go to confirm now.

packages/neuron-ui/src/components/Receive/index.tsx
duanyytop: OK

packages/neuron-ui/src/components/Receive/index.tsx
duanyytop: Indeed, it could not work and I import `grommet-icons` to provide copy image

ashchan: Great. But we still need to figure out how to make image and other resource loading work.

duanyytop: OK. I will research this question.

packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -51,21 +41,30 @@ const Receive = ({ address }: { address: string }) => {
...
-          <AddressPanel>
-            <Address onClick={() => window.clipboard.writeText(accountAddress)}>{accountAddress}</Address>
-            <OverlayTrigger placement="bottom" overlay={<Tooltip id="address-tooltip">{t('Copy address')}</Tooltip>}>
-              <CopyImage onClick={() => window.clipboard.writeText(accountAddress)} />
-            </OverlayTrigger>
-          </AddressPanel>
+          <Form.Group>
+            <Form.Label>{t('Address')}</Form.Label>
+            <AddressPanel>
+              <OverlayTrigger placement="bottom" overlay={<Tooltip id="address-tooltip">{t('Copy address')}</Tooltip>}>
```
ashchan: Please make sure we're using namespace i18n keys. (Can be in a future PR to make this one get merged soon.)

duanyytop: OK

duanyytop: Neuron-ui don't use namespace i18n keys now and I will create a new issue and branch to solve this question.

BaojunCZ: yes,I am doing it

packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -10,33 +11,73 @@ declare global {
   }
 }
 
+const AddressPanel = styled.div`
+  dispaly: flex;
+  display: -webkit-flex; /* Safari */
```
Keith-CY: unnecessary, styled-component will add this prefix

packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -10,33 +11,73 @@ declare global {
...
-  margin-top: 50px;
+const QRCodeModal = styled.div`
   text-align: center;
 `
 
-const Receive = () => {
+const Receive = ({ address }: { address: string }) => {
   const [t] = useTranslation()
-  const address: string = '0x0da2fe99fe549e082d4ed483c2e968a89ea8d11aabf5d79e5cbf06522de6e674'
+  const [showLargeQRCode, setShowLargeQRCode] = useState(false)
```
Keith-CY: component shouldn't have state, all state are just management in container, like MainContent

packages/neuron-ui/src/components/Receive/index.tsx
duanyytop: Without `-webkit-flex`, it don't work.

packages/neuron-ui/src/components/Receive/index.tsx
duanyytop: You mean all states of components(belong to main content page) should be put in `containers/MainContent`?

ashchan: From a practice perspective, I think it's fairly fine to have simple 'state' like this within a pure component. This is only a toggle value to determine if the QR should be displayed in a large.

packages/neuron-ui/src/components/Receive/index.tsx
Keith-CY: Interesting, the styled-component should add that automatically.

Leave it, it's running only in chrome, prefix is unnecessary.

packages/neuron-ui/src/components/Receive/index.tsx
Keith-CY: A component with its inner state should be regarded as a widget.

There are three ways to display the larger one.

1. place the larger one in Dialog Widget, and set the dialog state in MainContent to toggle it.
2. add a new widget which has its own state to toggle the size of its picture.
3. use pure CSS to display/hide the image.

It's ok for now. Refactor it later.

ashchan:
@duanyytop Looks good now. Could you please resolve the conflicts?

duanyytop:
OK. I will fix conflict now.

PR #155, opened 2019-03-19
feat: add `transaction` module

classicalliu:


packages/neuron-wallet/src/cell.ts
```diff
@@ -6,6 +6,22 @@ export interface OutPoint {
   index: number
 }
 
+interface TypeScript {
```
Keith-CY: interface Script is enough

packages/neuron-wallet/src/cell.ts
```diff
@@ -6,6 +6,22 @@ export interface OutPoint {
...
+  version: number
+  args: string[]
+  signedArgs: string[]
+  reference: string | null
```
Keith-CY: reference: string | undefined, and better not to use null [guide](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined)

packages/neuron-wallet/src/cell.ts
```diff
@@ -6,6 +6,22 @@ export interface OutPoint {
   index: number
 }
 
+interface TypeScript {
+  version: number
+  args: string[]
```
Keith-CY: in node we prefer to use Buffer or Uint8Array

packages/neuron-wallet/src/cell.ts
```diff
@@ -6,6 +6,22 @@ export interface OutPoint {
...
+  signedArgs: string[]
+  reference: string | null
+  binary?: string
+}
+
+export interface Cell {
+  capacity: number
+  data: string
+  lock: string
+  type?: TypeScript | null
```
Keith-CY: use undefined

packages/neuron-wallet/src/transaction.ts
```diff
@@ -0,0 +1,27 @@
...
+  inputs?: any
+  outputs?: Cell[]
+}
+
+// if addresses set to null, should be all transactions
+export const getTransactions = async (addresses: string[] | null, page: number, perPage: number) => {
+  const transaction: Transaction = mockedTransaction
+
+  return {
+    totalCount: page * perPage,
```
Keith-CY: seems not good

packages/neuron-wallet/src/cell.ts
classicalliu: OK

packages/neuron-wallet/src/cell.ts
classicalliu: OK

packages/neuron-wallet/src/cell.ts
classicalliu: Node use uint8array but RPC use hex string as input and output, I don't know why we should use uint8array

packages/neuron-wallet/src/transaction.ts
classicalliu: Just a mocked data.

packages/neuron-wallet/src/cell.ts
classicalliu: typescript can be a script or null

packages/neuron-wallet/src/cell.ts
classicalliu: should accept null as well, the RPC result will set to null, it should be compatible

packages/neuron-wallet/src/cell.ts
ashchan: I strongly recommend that we use `string` for CKB internal `[uint8]`, and `string[]` for `[[uint8]]` (`vec<vec<uint8>>`), as long as RPC returns hex string. Only use bytes when hashing internally.

PR #156, opened 2019-03-19
Impl send capacity

Keith-CY:


Keith-CY:
- [x] sendCapacity(items: {address: string, capacity: string, unit: Unit}[], password: string)
- [x] read address from qrcode

packages/neuron-ui/src/components/TransferConfirm/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+        e.preventDefault()
+        e.stopPropagation()
+      }}
+    >
+      <Card.Header>{title}</Card.Header>
+      <Card.Body>
+        <Card.Text>{`${t('send.thistransferwillsend')}: `}</Card.Text>
+        {typeof message === 'string' ? <Card.Text>{message}</Card.Text> : message}
+        <InputGroup>
+          <FormControl placeholder={t('send.inputpasswordtoconfirm') as string} value={psw} onChange={onChange} />
```
ashchan: I have to confess that lower cased key without separator like `-` or `_` is not readable. Which would be  better, `send.inputPasswordToConfirm` or `send.input-password-to-confirm`?

packages/neuron-ui/src/containers/MainContent/reducer.ts
```diff
@@ -122,6 +150,7 @@ export const actionCreators = {
       payload: ns,
     }
   },
+
   deleteNetwork: (name: string) => {
     if (name === Testnet) {
```
ashchan: Perhaps better have an internal flag or identifier. (Just in case we remove the unique name requirements, under which name comparison is insufficient.

packages/neuron-ui/src/utils/const.ts
```diff
@@ -62,3 +62,34 @@ export enum Routes {
 export enum LocalStorage {
   Networks = 'networks',
 }
+
+export enum CapacityUnit {
+  CKB = 'ckb',
+  CKKB = 'ckkb',
+  CKGB = 'ckgb',
```
ashchan: Is this temporary unit or RFC or what is already defined by CKB?

packages/neuron-ui/src/utils/const.ts
```diff
@@ -62,3 +62,34 @@ export enum Routes {
...
+  transfer: {
+    Address: 'eg: 0x0da2fe99fe549e082d4ed483c2e968a89ea8d11aabf5d79e5cbf06522de6e674',
+    Capacity: 'eg: 100',
+  },
+}
+
+export const Tooltips = {
+  transfer: {
+    Address: 'Address to send capacity',
+    Capacity: 'Capacity to send',
```
ashchan: Maybe worthy a discussion, but it seems `Amount` is more easier to understand for end user than `Capacity`.

packages/neuron-ui/src/widgets/QRScanner/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+interface QRScannerProps {
+  title: string
+  label: string
+  prompt?: string
+  onConfirm: Function
+  styles?: { [index: string]: any }
+}
+
+const QRScanner = ({ title, label, onConfirm, styles }: QRScannerProps) => {
```
ashchan: Just a question: does this work cross-platform?

packages/neuron-wallet/src/main.ts
```diff
@@ -88,6 +88,9 @@ function createWindow() {
   monitorChain(mainWindow.webContents)
   const terminalChannel = new TerminalChannel(mainWindow.webContents)
   terminalChannel.start()
+  BrowserWindow.addDevToolsExtension(
+    `/Users/ChenYu/Library/Application Support/Google/Chrome/Default/Extensions/fmkadmapgofadopljbjfkapdkoienihi/3.6.0_0`,
```
ashchan: Never use hardcoded local path that's specific to a computer.

Keith-CY: unbelievable mistake, and lucky that no more info leaks. Will never do it again.

packages/neuron-ui/src/components/TransferConfirm/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  title,
+  message,
+  psw,
+  onChange,
+  onSubmit,
+  onCancel,
+}: {
+  title: string
+  message: string | React.ReactNode
+  psw: string
```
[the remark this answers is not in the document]
Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/156/commits/7f84ed03ddf51dbabeb0d732706ef330d9875a01

packages/neuron-ui/src/components/TransferConfirm/index.tsx
Keith-CY: In a previous discussion, we decide to use lowercase, so I prefer `send.input-password-to-confirm`

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/156/commits/e8c2d4d2f1c7d73947c74eb1b1e8f651595339ef

packages/neuron-ui/src/widgets/QRScanner/index.tsx
Keith-CY: According to the code, yes. [Online Demo](https://cozmo.github.io/jsQR/)

packages/neuron-ui/src/utils/const.ts
Keith-CY: Just temporary uint

PR #158, opened 2019-03-20
fix: add clean before tsc

xiaoluSF:


packages/neuron-wallet/package.json
```diff
@@ -14,7 +14,7 @@
   "license": "MIT",
   "scripts": {
     "start": "cross-env electron .",
-    "start:dev": "tsc && cross-env electron .",
+    "start:dev": "yarn clean && tsc && cross-env electron .",
     "build": "tsc",
     "clean": "rm -rf dist/*",
```
Keith-CY: does `rm -rf` works on windows?

xiaoluSF: now we just think about macOS and Linux.

classicalliu: Only **package** on macOS and linux, support development on windows

xiaoluSF: 👌, I try it.

PR #165, opened 2019-03-22
feat: add sendCapacity interface in cell module

classicalliu:


ashchan:
@classicalliu CI fails.

classicalliu:
@ashchan CI passed

packages/neuron-wallet/src/transaction.ts
```diff
@@ -9,19 +10,121 @@ export interface Transaction {
...
   return {
     totalCount: page * perPage,
     transactions: Array.from({
       length: perPage,
     }).map(() => transaction),
-    addresses,
   }
 }
 
+const verifyScriptJsonObject = () => {
```
Keith-CY: This name comes with ruby-sdk, but not proper in JS, we can just name it `verifyScript`. Let's update it in the future.

packages/neuron-wallet/src/transaction.ts
```diff
@@ -9,19 +10,121 @@ export interface Transaction {
...
+  return {
+    version: 0,
+    reference: systemScriptCellHash,
+    signedArgs,
+  }
+}
+
+// it's still a mocked interface
+// FIXME: replace number with BigInt
+const gatherInputs = async (capacity: string, minCapacity: string) => {
```
Keith-CY: This method has overlapped with sdk

packages/neuron-wallet/src/transaction.ts
```diff
@@ -9,19 +10,121 @@ export interface Transaction {
...
+  if (inputCapacities < capacityInt) {
+    throw new Error('Capacity not enough!')
+  }
+  return {
+    inputs,
+    capacities: inputCapacities.toString(),
+  }
+}
+
+export const generateTx = async (address: string, capacity: string) => {
```
Keith-CY: This method has overlapped with sdk

packages/neuron-wallet/src/transaction.ts
```diff
@@ -9,19 +10,121 @@ export interface Transaction {
...
+    version: 0,
+    deps: [systemScriptOutPoint],
+    inputs,
+    outputs,
+  }
+}
+
+// mock interface, use asw as ours wallet as mocked data.
+// TODO: should replace asw with wallet, now SDK data not be compatible
+export const sendCapacity = async (targets: { address: string; capacity: string }[]) => {
```
Keith-CY: This method has overlapped with sdk

classicalliu: Those methods should implement in wallet, not sdk, all data infos are in wallet

classicalliu: It's mocked methods, focus on interface

ashchan: I agree that we should pull out UTXO and tx generation logic out of SDK.

PR #166, opened 2019-03-24
feat(neuron-ui): add history and transaction detail view

Keith-CY:
Loading style needs to be determined.

### DIFF
--- a/packages/neuron-ui/src/components/History/index.tsx
+++ b/packages/neuron-ui/src/components/History/index.tsx
@@ -1,7 +1,7 @@
 import React, { useContext, useEffect, useCallback } from 'react'
 import styled from 'styled-components'
 import { Link, RouteComponentProps } from 'react-router-dom'
-import { Container, Row, Col, Badge, ListGroup, ListGroupItem, Dropdown, Alert } from 'react-bootstrap'
+import { Container, Row, Col, Badge, Table, Dropdown, Alert } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 import { Close as CloseIcon } from 'grommet-icons'
 
@@ -17,8 +17,9 @@ import { TransactionType, EXPLORER } from '../../utils/const'
 import { dateFormatter } from '../../utils/formatters'
 import i18n from '../../utils/i18n'
 
-const TimeHeader = styled(ListGroupItem)`
-  background-color: #eee;
+const MetaData = styled.td`
+  display: flex;
+  flex-direction: column;
 `
 
 const AddressBadge = styled(Badge)`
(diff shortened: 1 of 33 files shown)


packages/neuron-ui/.env
```diff
@@ -1 +1,2 @@
 REACT_APP_NETWORKS=[{"name":"Testnet","remote":"http://localhost:8114"},{"name":"Localhost","remote":"http://localhost:8114"}]
+REACT_EXPLORER=http://localhost:3000
```
ashchan: Please move both `REACT_APP_NETWORKS` and `REACT_EXPLORER` to in-code constants (and update names as well). For the foreseen period there's no need to make it so flexible to be configurable via ENV.

packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -1,63 +1,172 @@
...
+          <div>{i18n.t(`history.${TransactionType[tx.type as TransactionType]}`.toLowerCase())}</div>
+          <div>{time}</div>
+        </>
+      ),
+      key: tx.hash,
+      date,
+    }
+  })
+
+const groupHistory = (items: Transaction[]) => {
```
ashchan: How about grouping transactions first, then constructing the list (`JSX.Element`s).

The former is how we organize the data, while the later is for UI display.

packages/neuron-ui/src/containers/MainContent/actionCreators/networks.ts
```diff
@@ -0,0 +1,92 @@
...
+  saveNetworks: (idx: number, networks: Network[], editorNetwork: Network, navTo: (path: string) => void) => {
+    if (!editorNetwork.name) {
+      return {
+        type: MainActions.ErrorMessage,
+        payload: {
+          networks: Message.NameIsRequired,
+        },
+      }
+    }
+    if (editorNetwork.name.length > 28) {
```
ashchan: Please make `28` a constant.

packages/neuron-ui/.env
Keith-CY: REACT_APP_NETWORKS will be removed by another PR which move network management to neuron-wallet.

REACT_EXPLORER will be removed soon.

Keith-CY:
> There're many changes that are not related to history/transaction feature but are arch and other refactorings/improvements. Next time please separate them into smaller PRs.

My fault, I should do it earlier.

ashchan:
Not a problem, cheers on the refactoring. 👍

packages/neuron-ui/src/components/History/index.tsx
Keith-CY: I've tried before, it requires one more step to format data. But it's indeed more clearly. I'll do it now.

packages/neuron-ui/src/containers/MainContent/actionCreators/networks.ts
Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/166/commits/a3a5908cfa0a67cee9ff7c6f5c79906c16e9e5a7

packages/neuron-ui/.env
Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/166/commits/ba8fae7ebe7c88d03cad0879d52c0b7071b012f3

packages/neuron-ui/src/components/History/index.tsx
Keith-CY: updated in https://github.com/nervosnetwork/neuron/commit/051ad900754fc53de19ba414d575547058bd125e

PR #168, opened 2019-03-25
refactor(neuron-ui): display max-network-name-length with interpolation

Keith-CY:


packages/neuron-ui/src/containers/MainContent/actionCreators/networks.ts
```diff
@@ -26,11 +25,11 @@ export default {
         },
       }
     }
-    if (editorNetwork.name.length > 28) {
+    if (editorNetwork.name.length > MAX_NETWORK_NAME_LENGTH) {
       return {
         type: MainActions.ErrorMessage,
         payload: {
-          networks: `${i18n.t(Message.LengthOfNameShouldBeLessThanOrEqualTo)} ${MAX_NETWORK_NAME_LENGTH}`,
+          networks: Message.LengthOfNameShouldBeLessThanOrEqualTo,
```
ashchan: I guess `t(`messages.${errorMsgs.networks}`, {  length: MAX_NETWORK_NAME_LENGTH, })` is more suitable. The current usage assumes that `<Alert variant="warning">` always knows what the error message payload is.

Keith-CY: Use i18n in payload is ok in our case. But the message of the current language is stored in the state, which will not be updated on language switching. I'll update the code if we are sure it's not a problem.

ashchan: If so it's still a usage problem.

The current code would break if the `networks` payload message (i18n key) is changed to another one (for example if we have another error), and the outside is still treating it as a fixed error message.

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/168/commits/c59faa6df267f3e11315a9098cacb2bfd6f0db57

PR #169, opened 2019-03-26
feat: import wallet to store

BaojunCZ:


packages/neuron-ui/src/containers/MainContent/state.ts
```diff
@@ -2,9 +2,14 @@ import { CapacityUnit } from '../../utils/const'
 
 export const initState = {
   tempWallet: {
```
Keith-CY: It seems that `tempWallet` have the same functionality as `createWallet` in the state.

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -165,12 +168,47 @@ export class Listeners {
...
+          }
+          if (storedKeystore) {
+            walletStore.saveWallet(walletName, storedKeystore)
+            e.sender.send(Channel.ImportWallet, {
+              status: ResponseStatus.Success,
+              result: true,
+            })
+          } else {
+            e.sender.send(Channel.ImportWallet, {
+              status: ResponseStatus.Success,
```
Keith-CY: ResponseState.Fail

packages/neuron-wallet/src/channel/index.ts
```diff
@@ -165,12 +168,47 @@ export class Listeners {
...
+              result: false,
+              msg: 'Error',
+            })
+          }
+        } catch (error) {
+          e.sender.send(Channel.ImportWallet, {
+            status: ResponseStatus.Success,
```
Keith-CY: ResponseState.Fail

BaojunCZ: ok

Keith-CY: I've fix it in another PR.

BaojunCZ: ok

PR #171, opened 2019-03-26
Refactor neuron wallet channel

Keith-CY:
- refactor channels of transactions history and network
- clean unused methods


Keith-CY:
Wallet store updated today, refactor of wallet channel will be submit soon

Keith-CY:
wallet channel refactored

PR #172, opened 2019-03-27
refactor: import wallet store

BaojunCZ:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -87,37 +87,34 @@ class WalletsController {
...
+    password,
+    mnemonic,
+    keystore,
+  }: {
+    name: string
+    password: string
+    mnemonic: string
+    keystore: string
+  }): Response<String> => {
+    const walletId = WalletsController.service.import({
```
Keith-CY: service should not be related to business logic, format data for DB in the controller and use `WalletController.service.create` to save it.

BaojunCZ: So `WalletController.service.create` is for saving wallet.I will move `generateKey` from `WalletController.service.create` to controller.

ashchan: How do we tell that `service` is not business logic?

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -86,38 +89,43 @@ class WalletsController {
...
-  //       status: ResponseCode.Success,
-  //       result: true,
-  //     }
-  //   }
-  //   return {
-  //     status: ResponseCode.Fail,
-  //     msg: 'Failed to import wallet',
-  //   }
-  // }
+  public static import = ({
```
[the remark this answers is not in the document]
BaojunCZ: get it

packages/neuron-ui/src/containers/MainContent/actionCreators/wallets.ts
```diff
@@ -51,8 +51,12 @@ export default {
       payload: id,
     }
   },
-  importWallet: (params: typeof initState.tempWallet) => {
-    wallets(WalletsMethod.Import, params)
+  importWallet: (isKeystore: boolean, params: typeof initState.tempWallet) => {
+    if (isKeystore) {
+      wallets(WalletsMethod.ImportKeystore, params)
+    } else {
+      wallets(WalletsMethod.ImportKeystore, params)
```
ashchan:
```suggestion
      wallets(WalletsMethod.ImportMnemonic, params)
```

ashchan:
Please also take care of the conflicts.

packages/neuron-wallet/src/controllers/wallets.ts
Keith-CY: That's not very accurate what I've said. I mean in our case, the services just do the work of fetching data from DB or network. Other works like verification, format, should be done in controllers, which precede services.

ashchan: @Keith-CY Makes sense. Thanks for the clarification.

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -30,6 +30,7 @@ export default class WalletService {
     this.create(
       {
         name: 'asw',
+        keystore: '{"master":{"privateKey":"","chainCode":""},"password":"0"}',
```
Keith-CY: why does the keystore have a password property?

BaojunCZ: It`s for checking password for now.It will not be a property after updating the keystore structure.

PR #173, opened 2019-03-27
refactor(neuron-ui): refactor wallet wizard and set its state in Main…

Keith-CY:
…Content


packages/neuron-ui/src/containers/MainContent/state.ts
```diff
@@ -1,6 +1,13 @@
 import { CapacityUnit } from '../../utils/const'
 
 export const initState = {
+  mnemonic: {
```
ashchan: Does this `initState` mean if our app becomes more complicated and has more sub features, we have to store all local data within `MainContent`?

PR #174, opened 2019-03-27
refactor(neuron-ui): refactor UILayer

Keith-CY:
This PR based on https://github.com/nervosnetwork/neuron/pull/173, deal with that first


packages/neuron-ui/src/locales/en.ts
```diff
@@ -33,9 +46,7 @@ export default {
       detail: 'detail',
       explorer: 'explorer',
     },
-    transaction: {
-      goBack: 'go back',
-    },
+    transaction: { goBack: 'go back' },
```
classicalliu: Why not `go-back` 🤔

Keith-CY: previously we use camelCase, several days ago we are going to use hyphen-seperated-case. We'd better to issue a new task to format i18n keys.

PR #176, opened 2019-03-29
feat: use native context menu instead of dropdown

Keith-CY:


Keith-CY:
should the dropdown of the network switch in the header bar also be replaced by native context menu?

ashchan:
> should the dropdown of the network switch in the header bar also be replaced by native context menu?

To make thing simpler:

* Network switch: keep as dropdown
* Other content menu: use Bootstrap popover instead of native content menu

PR #177, opened 2019-03-29
fix(neuron-ui): fix links in wallet setting

Keith-CY:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -39,13 +39,18 @@ export default class WalletService {
...
   }
 
-  public index = (): Wallet[] => {
-    return walletStore.getAllWallets().map(w => ({
-      id: w.id,
-      name: w.name,
-      address: `address of ${w.id}`,
-      publicKey: Buffer.from(`address of ${w.id}`),
-    }))
+  public index = (
```
ashchan: Discussion: since `neuron-wallet` is not UI layer or web service, `index` sounds very odd. Would `list` or something else be a better name even we think this as RESTful resource?

Keith-CY: Agree, @classicalliu do you have any idea about the method names?

classicalliu: `getWallets` ?

Keith-CY: The name of the methods could be determined later.

ashchan:
Oops CI build fails.

Keith-CY:
CI passed

PR #178, opened 2019-03-29
refactor: delete export wallet

BaojunCZ:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -74,6 +74,12 @@ export default class WalletService {
     throw new Error('Failed to create wallet')
   }
 
+  public validate = ({ id, password }: { id: string; password: string }) => {
+    const wallet = walletStore.getWallet(id)
+    Key.checkPassword(wallet.keystore, password)
```
Keith-CY: useless code

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -74,6 +74,12 @@ export default class WalletService {
...
+    return Key.checkPassword(wallet.keystore, password)
```
ashchan: Instead of having a `Key.,checkPassword` static method, it's making more sense to have a `Keystore.checkPassword` instance method here.

BaojunCZ: I watch pr of keystore. Finding `fromKey` missing in mock data.I confused with key and keystore in the past.I refactor this part now.In fact,`fromKeystore` don`t need password.`fromKey` does

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -123,21 +117,18 @@ class WalletsController {
   public static importKeystore = ({
     name,
     password,
-    keystore,
+    key,
```
ashchan: Why renaming this? Doesn't `keystore` look more suitable than `key`?

BaojunCZ: get it

PR #179, opened 2019-03-29
Implement keystore

duanyytop:
* [x] Implement generating and parsing keystore data logic
* [x] Implement searching HD wallet unused address logic
* [x] Implement generating and recommending HD wallet children address logic


duanyytop:
Fix CI fail

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -95,7 +79,7 @@ class WalletsController {
     password: string
     mnemonic: string
   }): ChannelResponse<Wallet> => {
-    const storedKeystore = Key.fromMnemonic(mnemonic, true, password).getKeystoreString()
+    const storedKeystore = Key.fromMnemonic(mnemonic, password, 17, 3).getKeystore()
```
ashchan: What are the magic numbers `17` and `3`?

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -56,28 +56,12 @@ class WalletsController {
...
-  }): ChannelResponse<Wallet> => {
-    const keystore = Key.generateKey(password).getKeystoreString()
-    const wallet = WalletsController.service.create({
-      name,
-      keystore,
-      address,
-      publicKey,
-    })
-    if (wallet) {
+  public static create = (): ChannelResponse<string> => {
```
ashchan: This method doesn't describe what it does. Generating a mnemonic is not a persisting operation and it should not be called `Wallet.create`. No wallet is created when calling this.

Consider renaming it to `generateMnemonic`. Or expose a mnemonic helper class to UI(which is responsible for mnemonic generation, validation, etc).

packages/neuron-wallet/src/keys/address.ts
```diff
@@ -0,0 +1,13 @@
+const Address = {
+  isUsedAddress: (address: string) => {
+    // TODO: check whether the address has history transactions
+    return address.includes('ckb')
+  },
+
+  getAddressFromPrivateKey: (privateKey: string) => {
```
ashchan:
```suggestion
  addressFromPrivateKey: (privateKey: string) => {
```

packages/neuron-wallet/src/keys/hd.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+enum AddressType {
+  Receive = 0,
+  Change = 1,
+}
+
+const HD = {
+  generateReceiveAndChangeAddresses: (keysData: KeysData, receiveNumber: number, changeNumber: number) => {
+    if (receiveNumber < 1 || changeNumber < 1) {
```
ashchan: What does `receiveNumber` mean? And shouldn't we guard program so that we don't allow a big number to pass in and make a long time loop?

packages/neuron-wallet/src/keys/hd.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+    return children
+  },
+
+  getPath: (type: AddressType, index: number) => {
+    return `m/${BIP44Params.Purpose}/${BIP44Params.CoinTypeTestnet}/${BIP44Params.Account}/${type}/${index}`
+  },
+
+  getAddressFromHDIndex: (root: BIP32, index: number, type = AddressType.Receive) => {
```
ashchan:
```suggestion
  addressFromIndex: (root: BIP32, index: number, type = AddressType.Receive) => {
```

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -132,7 +116,7 @@ class WalletsController {
     }
     const wallet = WalletsController.service.create({
       name,
-      keystore,
+      keystore: key.getKeystore()!!,
```
classicalliu: What's the `!!` means ?

packages/neuron-wallet/src/controllers/wallets.ts
duanyytop: Here is controller, not keystore. I just fix it because key's interface changed. So please @Keith-CY and @BaojunCZ refactor it. 

Of course they can tell me the logic, and I refactor it.

packages/neuron-wallet/src/controllers/wallets.ts
duanyytop: 17 and 3 come from 20 addresses in address page.

Maybe they can be set default value or other more better style.

packages/neuron-wallet/src/keys/address.ts
duanyytop: OK

packages/neuron-wallet/src/keys/hd.ts
duanyytop: A new wallet will have 20 addresses which includes 17 receive addresses and 3 change addresses. 

So do you think this action should be done by controller or ui ?

packages/neuron-wallet/src/keys/hd.ts
duanyytop: Get it.

packages/neuron-wallet/src/controllers/wallets.ts
ashchan: Same question here.

duanyytop: It's a mistake.

packages/neuron-wallet/src/keys/hd.ts
```diff
@@ -0,0 +1,112 @@
...
+const MaxAddressNumber = 20
+
+const HD = {
+  generateReceiveAndChangeAddresses: (
+    keysData: KeysData,
+    receiveAddressNumber: number,
+    changeAddressNumber: number,
+  ) => {
+    if (receiveAddressNumber < 1 || changeAddressNumber < 1) {
+      throw new Error('Address number error.')
```
classicalliu: Prefer more clear error message.

packages/neuron-wallet/tests/key.test.ts
```diff
@@ -8,17 +8,33 @@ describe('Key tests', () => {
 
   it('import key from mnemonic', async () => {
     const key = Key.fromMnemonic(mnemonic, '1qaz.2wsx', 17, 3)
-    expect(privateKey).toBe(key.getPrivateKey())
-    expect(key.getAddresses()!.receive.length).toEqual(17)
-    expect(key.getAddresses()!.change.length).toEqual(3)
-    expect(key.getAddresses()!.receive[0]).not.toBe(undefined)
+    if (key.keysData === undefined) {
```
Keith-CY: IMO, throwing errors in tests is unnecessary

duanyytop: OK

packages/neuron-wallet/src/keys/hd.ts
duanyytop: OK

classicalliu: You can do this later in another pr.

duanyytop: After all reviewing, I will update code one time. 

Maybe it doesn't have more questions.

duanyytop:
If it doesn't have any important questions, I suggest we can merge firstly. 

I will update code for the two issues above.

PR #188, opened 2019-04-03
refactor: generate uuid in service

BaojunCZ:


packages/neuron-wallet/tests/store/wallet_store.js
```diff
@@ -1,30 +1,35 @@
...
-    master: {
-      privateKey: 'privateKey2',
-      chainCode: 'chainCode2',
+  const wallet2 = {
+    id: uuid.v4(),
+    name: 'wallet2',
+    keystore: {
+      version: 0,
+      id: '1',
+      crypto: { cipher: 'wallet1', cipherparams: null, ciphertext: 'wallet2', kdf: '2', kdfparams: null, mac: '2' },
```
Keith-CY: `cipher: 'wallet1'` or `cipher: 'wallet2'`

packages/neuron-wallet/tests/store/wallet_store.js
```diff
@@ -33,37 +38,25 @@ describe('wallet store', () => {
...
-      keystore: mKeystore1,
-    })
+    assert.deepEqual(wallet1, wallet1Store)
+    assert.deepEqual(wallet2, wallet2Store)
   })
 
   it('get not exist wallet', () => {
-    walletStore.saveWallet(mWalletName1, mKeystore1)
-    walletStore.saveWallet(mWalletName1, mKeystore1)
+    walletStore.saveWallet(wallet1)
```
Keith-CY: why was the wallet1 saved twice?

BaojunCZ: I have removed.

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -51,18 +50,12 @@ export default class WalletStore {
...
-    const walletId = v4()
+  saveWallet = (walletData: WalletData): string => {
     let idList = this.getIDList()
-    const walletData = {
-      id: walletId,
-      name: walletName,
-      keystore: walletKeystore,
-    }
-    idList = idList.concat(walletId)
+    idList = idList.concat(walletData.id)
```
ashchan: Better extract id list update into a separate method.

BaojunCZ: ok

packages/neuron-wallet/src/store/WalletStore.ts
```diff
@@ -38,8 +38,14 @@ export default class WalletStore {
     return this.walletIDStore.get(WalletIDKey, [])
   }
 
-  private setIDList = (list: string[]) => {
-    this.walletIDStore.set(WalletIDKey, list)
+  private addIDList = (id: string) => {
```
ashchan: Perhaps `addWalletID` and `removeWalletID`? It's not adding/deleting ID list, but ID (single entry).

BaojunCZ: ok

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -55,7 +56,7 @@ export default class WalletService {
   }
 
   public create = ({ name, keystore }: { name: string; keystore: Keystore }): Wallet => {
-    const id = walletStore.saveWallet(name, keystore)
+    const id = walletStore.saveWallet({ id: v4(), name, keystore })
```
ashchan: Now that we pass in pre-generated ID, it should always return a valid ID so no need to check `if (id)`, right?

`saveWallet` doesn't have to return the `ID` anymore.

BaojunCZ: Right

Keith-CY:
CI failed

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -56,15 +56,13 @@ export default class WalletService {
...
-    const id = walletStore.saveWallet({ id: v4(), name, keystore })
-    if (id) {
-      const storedWallet = walletStore.getWallet(id)
-      return {
-        id,
-        name: storedWallet.name,
-      }
+    const id = v4()
+    walletStore.saveWallet({ id, name, keystore })
+    const storedWallet = walletStore.getWallet(id)
```
Keith-CY: The `storedWallet` is equal to `{id, name, keystore}`. Refactor it later.

BaojunCZ: ok

PR #160, opened 2019-03-21
feat: delete and edit wallet

BaojunCZ:

PR #192, opened 2019-04-04
refactor: wallet-store adapt keystore

BaojunCZ:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -28,7 +29,7 @@ class WalletsController {
     this.channel = channel
   }
 
-  public static index = (): ChannelResponse<Wallet[]> => {
+  public static index = (): ChannelResponse<WalletData[]> => {
```
Keith-CY: As #160 in internal discussion, we are going to use method names like `getAll` instead of `index`, `get` instead of `show`

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -82,28 +83,23 @@ class WalletsController {
     mnemonic: string
     receiveAddressNumber: number
     changeAddressNumber: number
```
Keith-CY: just have a question: why does import keystore require receive address number and change address number?

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -1,68 +1,55 @@
...
+  }): WalletData => {
     const id = v4()
-    walletStore.saveWallet({ id, name, keystore })
-    const storedWallet = walletStore.getWallet(id)
-    return {
-      id,
-      name: storedWallet.name,
-    }
+    walletStore.saveWallet({ id, name, keystore, addresses })
+    this.wallets = walletStore.getAllWallets()
```
Keith-CY: If `this.wallets` need to be udpated every time, better to use `getter` method
```javascript
get wallets () {
  return walletStore.getAllWallets()
}
```

packages/neuron-wallet/src/controllers/wallets.ts
BaojunCZ: ok

packages/neuron-wallet/src/controllers/wallets.ts
BaojunCZ: @duanyytop

packages/neuron-wallet/src/controllers/wallets.ts
BaojunCZ: Which pr?

Keith-CY: You can find it [here](https://github.com/nervosnetwork/neuron/blob/52d578f64d60895ae97d2b1ff710e2525b4e2d89/packages/neuron-wallet/src/services/networks.ts)

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -30,7 +30,7 @@ class WalletsController {
   }
 
   public static index = (): ChannelResponse<WalletData[]> => {
```
Keith-CY: It seems that this method has been missed in the refactor.

BaojunCZ: yes

packages/neuron-wallet/src/controllers/wallets.ts
duanyytop: Creating and importing wallet will provide 20(maybe any number) addresses for user. Please reference Neuron Addresses UI.

Keith-CY: The functionality of them is clear, but what add them into the keystore? I mean, is the keystore related to providing 20 addresses?

BaojunCZ: They're parallel in structure

PR #195, opened 2019-04-06
feat(neuron-wallet): import sqlite3 and typeorm

Keith-CY:


packages/neuron-wallet/src/main.ts
```diff
@@ -1,6 +1,8 @@
 import path from 'path'
 import { app, BrowserWindow, Menu } from 'electron'
 import windowStateKeeper from 'electron-window-state'
+// db
```
ashchan:
```suggestion
```

PR #198, opened 2019-04-08
feat: tx service/controller get interface

classicalliu:


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -1,31 +1,65 @@
...
-    if (params) {
-      // TODO
+  public static getAll = (params: TransactionsByLockHashesParam): PaginationResult<Transaction> => {
+    const transaction: Transaction = mockedTransaction
+
+    return {
+      totalCount: params.pageNo * params.pageSize,
+      items: Array.from({
+        length: params.pageSize,
+      }).map(() => transaction),
```
Keith-CY: Transactions have the same hash will make the neuron-ui render only one record.

classicalliu: Ok

packages/neuron-wallet/src/controllers/transactions.ts
```diff
@@ -11,26 +16,31 @@ export default class TransactionsController {
...
-    if (transactions) {
-      if (!params) {
-        return {
-          status: ResponseCode.Success,
-          result: transactions,
-        }
+  public static getAll = (params: TransactionsByLockHashesParam): ChannelResponse<PaginationResult<Transaction>> => {
+    const transactions = TransactionsService.getAll(params)
+
+    if (!transactions) {
```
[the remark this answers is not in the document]
classicalliu: I thinks `if (!transactions)` more like an error break logic, and can follow other break logics, so do something and break(return), return a normal logic at last.

PR #199, opened 2019-04-08
chore: check tx exists by address in tx service

classicalliu:


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -69,6 +69,11 @@ export default class TransactionsService {
     return transactions.find(tx => tx.hash === hash)
   }
 
+  // check weather the address has history transactions
```
Keith-CY: `wether` not `weather`

classicalliu: ...OK

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -69,6 +69,11 @@ export default class TransactionsService {
     return transactions.find(tx => tx.hash === hash)
   }
 
+  // check whether the address has history transactions
+  public static isExist = (_address: string): boolean => {
```
ashchan: `isExist` is incorrect in English grammar.

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -69,6 +69,11 @@ export default class TransactionsService {
...
+    return Math.random() > 0.5
```
ashchan: `>= 0.5` gives us 50% probability, while `> 0.5` doesn't.

packages/neuron-wallet/src/services/transactions.ts
Keith-CY: the meaning of “isExist address” is incorrect, use “hasTransactions” instead

classicalliu:
@Keith-CY call for review

PR #201, opened 2019-04-09
refactor: refactor keys test case and address method

duanyytop:


packages/neuron-wallet/tests/key.test.ts
```diff
@@ -8,33 +8,17 @@ describe('Key tests', () => {
...
-    }
-    if (key.addresses === undefined) {
-      throw new Error('Addresses is undefined.')
-    } else {
-      expect(key.addresses.receive.length).toEqual(17)
-      expect(key.addresses.change.length).toEqual(3)
-      expect(key.addresses.receive[0]).not.toBe(undefined)
-    }
+    expect(privateKey).toBe(key.keysData!.privateKey)
+    expect(key.addresses!.receive.length).toEqual(17)
```
ashchan: Question (again): how  did we come up with the magic number `17` and `3`?

duanyytop: This number can be any value, not just 17 and 3. 

Key support method for controller to determine how many addresses they want to generate.

This 17 and 3 just for test, and maybe using constants replace them will be more better.

ashchan: I see, that makes sense. BTW how about determining the default numbers inside `Key` instead of passing in?

duanyytop: If we don't allow passing parameters to determine the number of addresses, we will refactor key module when the requirement changes.

ashchan: We can keep it to allow number parameter, while the default method assumes a fixed number inside.

BTW I also believe `Key` should take care of key derivation, but not addresses generation.

duanyytop: OK. I see. I will move default number to key.

And I agree with you very much about address. I just reserve a method to generate address from private key, because now we have not determined where should place address generation.(Maybe there is a consensus for it)

ashchan: Nice. We can do that with next tasks.

PR #206, opened 2019-04-10
refactor: Rename receive address to receiving address

ashchan:
Let's standardize it with `Receiving Address`.

### DIFF
--- a/packages/neuron-wallet/src/controllers/wallets.ts
+++ b/packages/neuron-wallet/src/controllers/wallets.ts
@@ -75,17 +75,17 @@ class WalletsController {
     name,
     password,
     mnemonic,
-    receiveAddressNumber = 17,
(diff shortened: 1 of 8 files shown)


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -75,17 +75,17 @@ class WalletsController {
     name,
     password,
     mnemonic,
-    receiveAddressNumber = 17,
+    receivingAddressNumber: receivingAddressNumber = 17,
```
Keith-CY: Type should be named under the rule of PascalCase.

ashchan: Oops, not sure why type was added in the first place. Automatically added by format?

PR #209, opened 2019-04-10
feat: add orm configs and cell/tx entity

classicalliu:


packages/neuron-wallet/src/env.ts
```diff
@@ -19,7 +19,7 @@ const env = {
       type: 1,
     },
   ],
-  dbName: 'neuron.db.local',
+  dbPath: path.join(myApp.getPath('userData'), 'cell.sqlite'),
```
ashchan: Better extract this to DB's own module.

classicalliu: Done

PR #214, opened 2019-04-10
fix: fix tx types to latest version

classicalliu:


classicalliu:
> Question: does this change required SDK update too?

This change not related to SDK, but SDK should update as well

PR #218, opened 2019-04-10
refactor: neuron-wallet jest environment

BaojunCZ:


package.json
```diff
@@ -37,7 +37,7 @@
   },
   "devDependencies": {
     "@cryptape/sdk-ts-config": "0.0.1",
-    "@types/jest": "^24.0.0",
+    "@types/jest": "^23.0.0",
```
Keith-CY: If the neuron-ui uses jest@23 and neuron-wallet uses jest@24, the @types/jest is better to be added in them, respectively.

packages/neuron-wallet/package.json
```diff
@@ -23,19 +23,38 @@
...
+        "moduleFileExtensions": [
+          "ts",
+          "js",
+          "json",
+          "node"
+        ],
+        "setupTestFrameworkScriptFile": "<rootDir>/setupTests.ts"
+      },
+      {
+        "displayName": "node",
```
Keith-CY: Why does the neuron wallet have two projects

package.json
BaojunCZ: They use jest@23 together now.

packages/neuron-wallet/package.json
BaojunCZ: “electron” project need electron environment."node" project doesn`t

Keith-CY: But they cover the same test files

BaojunCZ: different Regex."electron" is *.test.electron.ts,"node" is “.test.ts”

PR #219, opened 2019-04-11
fix: timestamp number => string in tx entity

classicalliu:


packages/neuron-wallet/src/entities/Transaction.ts
```diff
@@ -44,9 +44,9 @@ export default class Transaction extends BaseEntity {
   outputs!: Cell[]
 
   @Column({
-    type: 'bigint',
+    type: 'varchar',
```
Keith-CY: `string` will be cast to `varchar` automatically

classicalliu: Different database will have different behave, like number to int or float? I don't want to depend on default type cast

PR #223, opened 2019-04-11
basic progress of import wallets

Keith-CY:


ashchan:
Please make sure CI pass.

Keith-CY:
> Please make sure CI pass.

CI passed, a file renamed before but seems to be rollbacked.

packages/neuron-ui/src/services/UILayer.ts
```diff
@@ -129,16 +134,34 @@ export const wallets = (
...
+  get: (id: string) => void
   generateMnemonic: () => void
   importKeystore: (params: { name: string; keystore: string; password: string }) => void
   importMnemonic: (params: { name: string; mnemonic: string; password: string }) => void
   update: (params: { id: string; name?: string; password: string; newPassword?: string }) => void
   delete: (params: { id: string; password: string }) => void
   export: (id: string) => void
-  active: () => void
-  setActive: (id: string) => void
+  activeOne: () => void
```
ashchan: Does `activeOne` mean getting current active wallet? If so `getActive` would be easier to understand.

packages/neuron-wallet/src/main.ts
```diff
@@ -76,10 +76,14 @@ function createWindow() {
   i18n.changeLanguage(app.getLocale())
   Menu.setApplicationMenu(mainmenu())
 
-  mainWindow.loadURL(env.mainURL)
+  const url = WalletsController.activeOne().status ? env.mainURL : `${env.mainURL}/wallets/wizard`
```
ashchan: Feels we should have a fixed root URL in `neuron-wallet` and let the UI to determine which actual route/path to display on start up.

Keith-CY: UI will handle the case of an empty wallet too, but it's asynchronous that makes the app launch unnatural(view flash or something)

packages/neuron-ui/src/services/UILayer.ts
Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/223/commits/a2b0393e1cdf7689bde17b2497f31db4cd7e90fd

packages/neuron-wallet/src/main.ts
ashchan: Then what if based on data or logic the start screen could be more than just current `home` or `wizard`?

The nature of the web implementation would force us to have a better approach. Some considerations:

* Make it synchronous for certain calls. For example, checking if a wallet exist perhaps should support synchronous check.
* On the UI layer we show a start up blank screen with loading indicator, and switch to home (active wallet) or wizard or any other views we want to add in the future.

ashchan: Please leave that to a future plan or discussion.

Keith-CY: I'd like not to use a synchronous call cause it makes render blocked(I've tried in sendCapacity, the synchronous call should be put to the end of events queue manually). But a launch screen is a good idea, I'll add it soon.

PR #224, opened 2019-04-11
chore: read tx history from db

classicalliu:


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -51,9 +52,35 @@ export interface PaginationResult<T = any> {
 /* eslint @typescript-eslint/no-unused-vars: "warn" */
 export default class TransactionsService {
   public static getAll = (params: TransactionsByLockHashesParam): PaginationResult<Transaction> => {
+    // TODO: calculate lockHashes when saving transactions
+    let totalCount: number | undefined
+    TransactionEntity.count().then(n => {
```
Keith-CY: The promise might not be resolved when the function returns

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -74,7 +101,11 @@ export default class TransactionsService {
   }
 
   public static get = (hash: string): Transaction | undefined => {
-    return mockedTransactions.find(tx => tx.hash === hash)
+    let transaction
+    TransactionEntity.findOne(hash).then(tx => {
```
Keith-CY: The promise might not be resolved when the function returns

PR #239, opened 2019-04-16
refactor: wallet manager ui

BaojunCZ:


packages/neuron-ui/src/components/Settings/InputWalletPasswordDialog.tsx
```diff
@@ -94,40 +83,46 @@ const InputWalletPasswordDialog = ({
...
+                  })
+                }
+              >
+                {t('common.cancel')}
+              </Button>
+            </ButtonDiv>
+          </Card.Footer>
+        </>
+      ) : (
+        <div>Network not found</div>
```
Keith-CY: `Wallet not found`

packages/neuron-ui/src/services/UILayer.ts
```diff
@@ -140,13 +142,22 @@ export const walletsCall = instantiateMethodCall(wallets) as {
...
   importMnemonic: (params: { name: string; mnemonic: string; password: string }) => void
   update: (params: { id: string; name?: string; password: string; newPassword?: string }) => void
-  delete: (params: { id: string; password: string }) => void
   export: (id: string) => void
   getActive: () => void
   activate: (id: string) => void
   backup: (id: string) => void
 }
 
+export const deleteWallet = (id: string, password: string, handleResult: any) => {
```
Keith-CY: Why move `delete` to an independent channel?

Keith-CY:
CI failed

packages/neuron-ui/src/services/UILayer.ts
BaojunCZ: It should validate password before deleting wallet.If password is wrong,I should notify user in the dialog.It likes sync method.

Keith-CY: In this case, pass id and password to the neuron-wallet together, and keep the dialog open.

If the password is correct, delete the wallet and broadcast to neuron-ui. neuron-ui hdies the dialog.

If the password is incorrect, send error message to neuron-ui, and neuron-ui render it on the dialog.

packages/neuron-ui/src/components/Settings/InputWalletPasswordDialog.tsx
```diff
@@ -94,40 +83,46 @@ const InputWalletPasswordDialog = ({
...
-              })
-            }
-          >
-            {t('common.cancel')}
-          </Button>
-        </ButtonDiv>
-      </Card.Footer>
+      {wallet ? (
+        <>
+          <Card.Header>{`Please Enter ${wallet.name} Password`}</Card.Header>
```
ashchan: Apply i18n for UI text.

packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -3,7 +3,7 @@ import { RouteComponentProps } from 'react-router-dom'
 import { Card, Form, Button, Alert } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 
-import { Routes } from '../../utils/const'
+// import { Routes } from '../../utils/const'
```
ashchan: Please do not keep such comments. Just delete unused code.

packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -92,7 +92,8 @@ export default (props: React.PropsWithoutRef<ContentProps & RouteComponentProps<
           wallet={myWallet}
           dispatch={dispatch}
           checkType={CheckType.EditWallet}
-          handle={() => props.history.push(`${Routes.SettingsWallets}`)}
+          // handle={() => props.history.push(`${Routes.SettingsWallets}`)}
```
ashchan: Delete this instead of comment out as well.

BaojunCZ: fixed

packages/neuron-ui/src/services/UILayer.ts
BaojunCZ: fixed

PR #243, opened 2019-04-16
refactor: edit wallet in neuron-wallet

BaojunCZ:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -15,6 +15,8 @@ export enum WalletsMethod {
   Delete = 'delete',
   GetActive = 'getActive',
   Activate = 'activate',
+  RenameWallet = 'renameWallet',
```
Keith-CY: `rename` and `update addresses` belong to `update`

BaojunCZ: fixed

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -34,33 +34,19 @@ export default class WalletService {
...
-  //       wallet.address = address
-  //     }
-  //     if (publicKey) {
-  //       wallet.publicKey = publicKey
-  //     }
-  //     return true
-  //   }
-  //   return false
-  // }
+  public update = ({ id, name, addresses }: { id: string; name?: string; addresses?: Addresses }): boolean => {
```
Keith-CY: use object combination to update wallet, like the following:
```javascript
const currentWallet = walletStore.get(newWallet.id)
walletStore.update({...currentWallet, ...newWallet})
```

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -34,33 +34,9 @@ export default class WalletService {
...
-  //     }
-  //     if (publicKey) {
-  //       wallet.publicKey = publicKey
-  //     }
-  //     return true
-  //   }
-  //   return false
-  // }
+  public update = (walletId: string, newWallet: WalletData) => {
+    walletStore.update(walletId, newWallet)
```
Keith-CY: It seems that `undefined` not handled in this case.
if `newWallet.name` is undefined, it's current name should be used.

```javascript
{ ...oldWallet, ...newWallet }
```
will cover older wallet except the undefined fields.

BaojunCZ: get it

PR #244, opened 2019-04-16
feat: sync connect status

Keith-CY:


packages/neuron-wallet/src/utils/initApp.ts
```diff
@@ -15,8 +18,25 @@ const initiateNetworks = () => {
...
+    () => {},
+    () => {
+      // TODO: handle complete
+    },
+  )
+}
+
 const initApp = () => {
   nodeService.start()
+  // TODO: this function should be moved to somewhere syncing data
```
ashchan: Before we move this, logically checking should happen after init.

Keith-CY: I'm not sure what does the checking mean, if it refers to error catch, which is handled in [service](https://github.com/nervosnetwork/neuron/pull/244/files#diff-33a5ced4a8dfd71e03e7e1352a3ebbfeR27)

ashchan: Without looking deep into `checkConnectStatus` and `initConnection`, these two methods seems to be related to each other by reading their names. So it's very strange for me to understand why `check` happens before `init`.

Keith-CY: > Without looking deep into `checkConnectStatus` and `initConnection`, these two methods seems to be related to each other by reading their names. So it's very strange for me to understand why `check` happens before `init`.

got the point, will it be better to rename the method to `syncConnectStatus` ?

ashchan: That'll look good.

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/244/commits/5d877b5678ac58547f2a1e59a3de1cb5794573c0

PR #248, opened 2019-04-17
feat(neuron-wallet): save log to user data in prod, save log to proje…

Keith-CY:
…ct in dev

### DIFF
--- a/packages/neuron-wallet/src/utils/logger.ts
+++ b/packages/neuron-wallet/src/utils/logger.ts
@@ -1,4 +1,9 @@
+import { app } from 'electron'
 import winston, { format } from 'winston'
+import env from '../env'
+
+const { isDevMode } = env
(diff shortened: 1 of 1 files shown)


packages/neuron-wallet/src/utils/logger.ts
```diff
@@ -10,13 +15,18 @@ const logger = winston.createLogger({
   format: combine(timestamp(), json()),
   transports: [
     new winston.transports.File({
-      filename: 'combined.log',
+      filename: `${path}combined.log`,
```
ashchan: Question: doesn't app usually write log files to platform's default system log locations?

Keith-CY: This PR is closed for a (new one](https://github.com/nervosnetwork/neuron/pull/249) which saves logs to `app.getPath('logs')`

PR #250, opened 2019-04-18
refactor: edit wallet

BaojunCZ:


packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -1,32 +1,48 @@
...
+import { Routes } from '../../utils/const'
 
-export default (props: React.PropsWithoutRef<ContentProps & RouteComponentProps<{ wallet: string }>>) => {
+export default (props: React.PropsWithoutRef<ContentProps & RouteComponentProps<{ id: string }>>) => {
   const { match, dialog, dispatch } = props
   const { params } = match
   const [t] = useTranslation()
+  const {
+    settings: { wallets },
+    messages: errorMessage,
```
Keith-CY: `errorMessage` => `errorMessages`

packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -1,32 +1,48 @@
...
 
-  const myWallet: Wallet = JSON.parse(params.wallet)
-  const [walletName, setWalletName] = useState(myWallet.name)
+  const myWallet = wallets.find(wallet => wallet.id === params.id)
+  const [walletName, setWalletName] = useState('')
   const [password, setPassword] = useState('')
   const [confirmPassword, setConfirmPassword] = useState('')
-  const [errorMsg, setErrorMsg] = useState('')
+
+  useEffect(() => {
```
Keith-CY: Maybe the dialog is unnecessary. 

Input password with other fields versus a dialog for the password, which one is better?

ashchan: Many wallet would only ask and check password when user clicks 'save', 'send', etc.

ashchan: Also think of this: many devices including macOS computers now can handle password with Touch ID or Face ID. For this, we don't even have a password field for the form, but we'll usually trigger a confirm dialog.

packages/neuron-ui/src/components/WalletEditor/index.tsx
BaojunCZ: fixed

PR #255, opened 2019-04-22
fix: delete active wallet

BaojunCZ:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -177,6 +177,7 @@ class WalletsController {
       const success = WalletsController.service.delete(id)
       if (success) {
         windowManage.broadcast(Channel.Wallets, WalletsMethod.GetAll, WalletsController.getAll())
+        windowManage.broadcast(Channel.Wallets, WalletsMethod.GetActive, WalletsController.getActive())
```
ashchan: Are we 100% sure these two events are broadcast and processed in the exact first `getAll` then `getActive` order?

packages/neuron-wallet/tests/walletservice.test.electronts
```diff
@@ -0,0 +1,78 @@
+import assert from 'assert'
```
ashchan: What does the suffix `electronts` mean?

BaojunCZ: I forget to delete it.It is for wallet service test in my local.

packages/neuron-wallet/src/controllers/wallets.ts
BaojunCZ: I will try to make them to be one

PR #259, opened 2019-04-23
Refactor wallet editor

Keith-CY:


packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -1,65 +1,42 @@
...
-    {
-      label: t('settings.wallet-manager.edit-wallet.confirm-password'),
-      value: confirmPassword,
-      onChange: e => setConfirmPassword(e.currentTarget.value),
-      placeholder: t('settings.wallet-manager.edit-wallet.confirm-password'),
-      inputType: 'password',
-    },
-  ]
+  const inputs: InputProps[] = useInputs(editor)
+  const isParamsValid = useIsParamsValid(editor.name.value, editor.newPassword.value, editor.confirmNewPassword.value)
```
ashchan:
```suggestion
  const areParamsValid = useAreParamsValid(editor.name.value, editor.newPassword.value, editor.confirmNewPassword.value)
```

Kind of difficult to name this 😢

packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -79,36 +68,46 @@ export default (props: React.PropsWithoutRef<ContentProps & RouteComponentProps<
...
+            e.preventDefault()
+            e.stopPropagation()
+          }}
+          style={{
+            width: '40%',
+          }}
+        >
+          <>
+            <Card.Header>
+              {t('settings.wallet-manager.delete-wallet-title', { name: wallet ? wallet.name : '' })}
```
ashchan: Under what circumstances would a passed wallet be `nil`? Is it still making any sense to show the dialog when the wallet is not found?

Keith-CY: it could be removed caused `wallet` check has been executed at https://github.com/nervosnetwork/neuron/pull/259/files/6bc70c4452aa5ae33e8a0147592a2d4a7a340926#diff-764ef9b4eee7844302497a22e71cadadR26

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/259/commits/19789de6f0dcab7327d7c86981bf8f5fa97127e9

PR #261, opened 2019-04-23
chore: reset tx entity to has many inputs and outputs

classicalliu:
Refactor tx entity to has many inputs and outputs so we can search tx by lock hash


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -161,6 +207,12 @@ export default class TransactionsService {
     // TODO: calculate value, sum of not return charge output
     tx.value = Math.round(Math.random() * 10000).toString()
     tx.type = ['send', 'receive', 'unknown'][Math.round(Math.random() * 2)]
+    tx.outputs = tx.outputs!.map(o => {
+      const output = o
+      // TODO: calculate lockHash
```
ashchan: Just a reminder, now that we have SDK 0.9.0 included is it time to calculate the real lock hash?

classicalliu: I'll check it.

classicalliu: Updated

PR #266, opened 2019-04-25
refactor: add address module and refactor key module

duanyytop:
* Add address module for generating and searching
* Refactor key module to make hd more simple

### DIFF
--- a/packages/neuron-wallet/src/address/index.ts
+++ b/packages/neuron-wallet/src/address/index.ts
@@ -3,6 +3,7 @@ import ckbCore from '../core'
 import AddressType from './type'
 import HD from '../keys/hd'
 import { KeysData } from '../keys/keystore'
+import { HDAddress } from '../keys/key'
 
 const MaxAddressNumber = 30
 
@@ -37,13 +38,19 @@ class Address {
     } else if (receivingAddressNumber > MaxAddressNumber || changeAddressNumber > MaxAddressNumber) {
       throw new Error('Address number error.')
     }
-    const receivingAddresses = []
-    const changeAddresses = []
+    const receivingAddresses: HDAddress[] = []
+    const changeAddresses: HDAddress[] = []
     for (let index = 0; index < receivingAddressNumber; index++) {
-      receivingAddresses.push(Address.addressFromHDIndex(keysData, index, AddressType.Receiving))
+      receivingAddresses.push({
+        address: Address.addressFromHDIndex(keysData, index, AddressType.Receiving),
+        path: HD.path(AddressType.Receiving, index),
+      })
     }
     for (let index = 0; index < changeAddressNumber; index++) {
-      changeAddresses.push(Address.addressFromHDIndex(keysData, index, AddressType.Change))
+      changeAddresses.push({
+        address: Address.addressFromHDIndex(keysData, index, AddressType.Change),
+        path: HD.path(AddressType.Change, index),
+      })
     }
     return {
       receiving: receivingAddresses,
--- a/packages/neuron-wallet/src/address/index.ts
+++ b/packages/neuron-wallet/src/address/index.ts
@@ -14,17 +14,22 @@ class Address {
 
   public static addressFromPrivateKey = (privateKey: string) => {
     const account = ckbCore.wallet.accountFromPrivateKey(Buffer.from(ckbCore.utils.hexToBytes(privateKey)))
-    return ckbCore.utils.pubkeyToAddress(account.hexPubKey)
+    return Address.addressFromPublicKey(account.hexPubKey)
+  }
+
+  public static addressFromPublicKey = (publicKey: string) => {
+    return ckbCore.utils.pubkeyToAddress(publicKey)
   }
 
   public static addressFromHDIndex = (keysData: KeysData, index: number, type = AddressType.Receiving) => {
-    const privateKey = HD.privateKeyFromHDIndex(keysData, index, type)
+    const { privateKey } = HD.keyFromHDIndex(keysData, index, type)
     return Address.addressFromPrivateKey(privateKey)
   }
 
   public static latestUnusedAddress = (keysData: KeysData) => {
-    const latestUnusedIndex = HD.searchAddress(keysData, 20)
-    return HD.privateKeyFromHDIndex(keysData, latestUnusedIndex)
+    const latestUnusedIndex = Address.searchAddress(keysData, 20)
+    const { privateKey } = HD.keyFromHDIndex(keysData, latestUnusedIndex, AddressType.Receiving)
+    return Address.addressFromPrivateKey(privateKey)
   }
 
   // Generate both receiving and change addresses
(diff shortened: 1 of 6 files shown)


ashchan:
Let CI pass first ;-p

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    return Address.addressFromPrivateKey(privateKey)
+  }
+
+  public static latestUnusedAddress = (keysData: KeysData) => {
+    const latestUnusedIndex = Address.searchAddress(keysData, 20)
+    const { privateKey } = HD.keyFromHDIndex(keysData, latestUnusedIndex, AddressType.Receiving)
+    return Address.addressFromPrivateKey(privateKey)
+  }
+
+  // Generate both receiving and change addresses
```
Keith-CY: It's weird that one is **doing** and one is **noun**

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
...
+  public static generateAddresses = (
+    keysData: KeysData,
+    receivingAddressNumber: number,
+    changeAddressNumber: number,
+  ) => {
+    if (receivingAddressNumber < 1 || changeAddressNumber < 1) {
```
Keith-CY: use `count` instead of `number`, which is more presentive.

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        path: HD.pathFromIndex(AddressType.Change, index),
+      })
+    }
+    return {
+      receiving: receivingAddresses,
+      change: changeAddresses,
+    }
+  }
+
+  public static searchUsedChildAddresses = (keysData: KeysData) => {
```
Keith-CY: `searchUsedAddresses` is enough, the master key doesn't have an address.

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    const children: HDAddress[] = []
```
Keith-CY: variable name `children` is confusing.

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+      } else {
+        throw new Error('Empty private key')
+      }
+    }
+    return children
+  }
+
+  // TODO: refactor me
+  public static searchAddress = (
```
Keith-CY: This method seems to search an index instead of an address.

packages/neuron-wallet/src/keys/hd.ts
```diff
@@ -10,127 +9,32 @@ enum BIP44Params {
...
-    }
-    for (let index = 0; index < changeAddressNumber; index++) {
-      changeAddresses.push(HD.addressFromHDIndex(root, index, AddressType.Change))
-    }
-    return {
-      receiving: receivingAddresses,
-      change: changeAddresses,
-    }
+  public static pathFromIndex = (type: AddressType, index: number) => {
+    return `m/${BIP44Params.Purpose}/${BIP44Params.CoinTypeTestnet}/${BIP44Params.Account}/${type}/${index}`
```
Keith-CY: Is `CoinTypeTestnet` hard coded here?

packages/neuron-wallet/src/keys/key.ts
```diff
@@ -137,11 +137,18 @@ export default class Key {
 
   public latestUnusedAddress = () => {
     if (this.keysData) {
-      return HD.latestUnusedAddress(this.keysData)
+      return Address.latestUnusedAddress(this.keysData)
     }
     return ''
   }
 
+  public allUsedAddress = () => {
```
Keith-CY: `allUsedAddresses`

packages/neuron-wallet/src/address/index.ts
duanyytop: Receiving can be as a noun.

packages/neuron-wallet/src/keys/key.ts
duanyytop: I will solve all name problems tomorrow. Thanks.

packages/neuron-wallet/src/address/index.ts
ashchan: Here `receiving` and `change` are adj. and they are standard names from Bitcoin.

packages/neuron-wallet/src/keys/hd.ts
ashchan: We'll have a fixed number for `coin type` later, but most likely we'll use the same one for both testnet and mainnet.

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
...
+
+const MaxAddressNumber = 30
+
+export interface HDAddress {
+  address: string
+  path: string
+}
+
+class Address {
+  public static isUsedAddress = (address: string) => {
```
ashchan: Feel like `isAddressUsed` is more descriptive.

packages/neuron-wallet/src/keys/hd.ts
ashchan: In face we can just hard code this like `m/44/${BIP44Params.CoinTypeTestnet}/0/${type}/${index}`. `44` (purpose) and `0` (account) are well defined by BIP and won't change.

duanyytop: `44` (purpose) and `0` (account) are constant. And the reason is that  just a number can't tell others what mean. Maybe `enum` is not a better way to be as a constant. I don't know.

packages/neuron-wallet/src/address/index.ts
```diff
@@ -0,0 +1,124 @@
+import TransactionsService from '../services/transactions'
```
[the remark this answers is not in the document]
duanyytop: I found 'services/transactions' is still in neuron-wallet, and if I updated `import TransactionsService from 'services/transactions'` , it will show `Cannot find module`.

packages/neuron-wallet/src/keys/hd.ts
duanyytop: And we should avoid magic number according to default address number, so I think we should give every constant a more readable name.

ashchan: So `m/44/.../0/` is not a magic number, it's a de fact standard path.

If we need to define `44` etc. as constant and contact them, then why not also make `m` a constant? Naming and variable decision sometimes need certain degree of balance, not stick rules.

Extend this topic, it's fairly common for blockchain developers to search for `m/44` to look at the HD path inside a project's source code - that won't work for our variable names.

duanyytop: OK. I see.

PR #271, opened 2019-04-28
chore: impl generateTx

classicalliu:


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -217,49 +218,75 @@ export default class TransactionsService {
...
-
-    const outputs = [
-      {
-        capacity: '1000',
-        data: '',
+  // system contract info
+  public static contractInfo = async () => {
+    const genesisHash: string = await ckbCore.rpc.getBlockHash(0)
+    const genesisBlock = await ckbCore.rpc.getBlock(genesisHash)
+    const secpTx = genesisBlock.commitTransactions[0]
```
ashchan: secpTx doesn’t describe its meaning well.

classicalliu: renamed to `systemScriptTx`

PR #273, opened 2019-04-28
chore: impl getAll in TransactionsService

classicalliu:
1. implement `getAll`, `getAllByAddresses`, `getAllByPubkeys`
2. add static methods `addressToLockScript` and `addressToLockHash` to `TransactionsService`
3. calculate `capacity` and `lockHash` of `InputEntity` when save
4. remove `value` and `type` from `TransactionEntity`
5. calculate `value` and `type` by `params.lockHashes` when in `getAll`, `getAllByAddresses`, `getAllByPubkeys`

### DIFF
--- a/packages/neuron-wallet/src/services/transactions.ts
+++ b/packages/neuron-wallet/src/services/transactions.ts
@@ -135,8 +135,26 @@ export default class TransactionsService {
   }
 
   // check whether the address has history transactions
-  public static hasTransactions = (_address: string): boolean => {
-    return Math.random() >= 0.5
+  public static hasTransactions = async (address: string): Promise<boolean> => {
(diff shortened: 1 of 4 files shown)


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -89,20 +114,33 @@ export default class TransactionsService {
...
       pageSize: params.pageSize,
-      lockHashes: [],
+      lockHashes,
     })
   }
 
   public static getAllByPubkeys = async (
     params: TransactionsByPubkeysParams,
   ): Promise<PaginationResult<Transaction>> => {
+    const lockHashes: string[] = []
```
ashchan: Does `map` work for this loop (with `await` inside)?

classicalliu: Yes, I'll update it.

classicalliu: Updated

PR #281, opened 2019-04-29
refactor: refactor cells service

classicalliu:
1. remove load cell info and create cell methods, it's useless now
2. `getBalance` and `getLiveCell` should load cells with `status` live


Keith-CY:
conflict

classicalliu:
> conflict

resolved

PR #282, opened 2019-04-29
refactor: move sync blocks to SyncBlocksService

classicalliu:
1. move sync blocks methods to a single service
2. make lockHashes as an object variable


packages/neuron-wallet/src/services/syncBlocks.ts
```diff
@@ -0,0 +1,127 @@
+import { getConnection } from 'typeorm'
+import { Script, OutPoint, Cell } from './cells'
+import TransactionsService, { Input, Transaction } from './transactions'
+import InputEntity from '../entities/Input'
+import ckbCore from '../core'
+
+/* eslint no-await-in-loop: "off" */
+/* eslint no-restricted-syntax: "warn" */
+export default class SyncBlocksService {
+  private lockHashesVar: string[]
```
Keith-CY: The `Var` means `variable`?

classicalliu: yes, can't same as `lockHashes`, can't with `_` prefix

ashchan: If we really can't name it `lockHashes` to avoid name conflict, IMO `lockHashList` is probably better than `lockHashesVar`.

classicalliu: updated to `lockHashList`

PR #290, opened 2019-05-03
refactor: save tx after sent and fetch

classicalliu:


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -197,7 +209,73 @@ export default class TransactionsService {
     return false
   }
 
-  public static create = async (transaction: Transaction): Promise<TransactionEntity> => {
+  /* eslint no-await-in-loop: "off" */
+  /* eslint no-restricted-syntax: "off" */
+  // when after sent:
+  // 1. don't have before, output = sent, input = pending
+  // 2. have before, do nothing
```
ashchan: Suggestions for this comment:

```
// After the tx is sent:
// 1. If the tx is not persisted before sending, output = sent, input = pending
// 2. If the tx is already persisted before sending, do nothing
```

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -197,7 +209,73 @@ export default class TransactionsService {
...
+  public static saveWithSent = async (transaction: Transaction): Promise<TransactionEntity> => {
+    const txEntity: TransactionEntity | undefined = await getConnection()
+      .getRepository(TransactionEntity)
+      .findOne(transaction.hash)
+
+    if (txEntity) {
+      // nothing to do if exist before
```
ashchan:
```suggestion
      // nothing to do if exists already
```

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -197,7 +209,73 @@ export default class TransactionsService {
...
+      return txEntity
+    }
+    return TransactionsService.create(transaction, OutputStatus.Sent, OutputStatus.Pending)
+  }
+
+  // when after fetch:
+  // 1. don't have before, output = live, input = dead
+  // 2. have before, output = live, input = dead
+  public static saveWithFetch = async (transaction: Transaction): Promise<TransactionEntity> => {
```
ashchan:
```
// After the tx is fetched:
// 1. If the tx is not persisted before fetching, output = live, input = dead
// 2. If the tx is already persisted before fetching, output = live, input = dead
```

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -251,19 +339,26 @@ export default class TransactionsService {
...
       txEntities.push(txEntity)
     })
 
     return txEntities
   }
 
   // update previousOutput's status to 'dead' if found
-  public static convertTransactionAndCreate = async (transaction: Transaction): Promise<TransactionEntity> => {
+  // calculate output lockHash, input lockHash and capacity
+  // when sent a transaction, use TxSaveType.Sent
```
ashchan:
```suggestion
  // when send a transaction, use TxSaveType.Sent
```

PR #293, opened 2019-05-04
refactor: auto generate migrations

classicalliu:
1. add generate migration script
2. add `migration` dir to `.eslintignore`, it's probably auto generated files
3. disable auto synchronize
4. enable auto migration
5. generate current entities in migration


package.json
```diff
@@ -27,7 +27,8 @@
     "lint": "lerna run lint",
     "test": "lerna run --parallel test",
     "precommit": "lerna run precommit",
-    "postinstall": "npm rebuild tiny-secp256k1 && lerna run rebuild:nativemodules"
+    "postinstall": "npm rebuild tiny-secp256k1 && lerna run rebuild:nativemodules",
+    "migrate": "ts-node ./node_modules/.bin/typeorm migration:generate"
```
ashchan: Better call it `db:migrate` or something like that.

.gitignore
```diff
@@ -42,4 +42,4 @@ docs
 .idea
 
 # sqlite db
-*.db.*
+cell-dev.sqlite
```
ashchan: Isn't dev db stored under `userData` as well that we don't generate it inside project folder and need to ignore it?

package.json
classicalliu: ok

.gitignore
classicalliu: It's a problem that when I use `typeorm` `cli` to generate migration, it only follows config in `ormconfig.json`, if no `database` here, it will throw error. So it's only use is `cli` need one ..

The real database in our project is under `userData`, it set dynamically in code while initialize connection (should get `userData` dir firstly)

package.json
classicalliu: updated

.gitignore
ashchan: Got it. We should always put DB and other store files under `userData` for all environment, but if it's a limitation of `typeorm cli` for local operation then it's fine.

PR #296, opened 2019-05-04
chore: fix homepage link in neuron-ui/package.json

classicalliu:


packages/neuron-ui/package.json
```diff
@@ -7,7 +7,7 @@
     "email": "dev@nervos.org",
     "url": "https://github.com/nervosnetwork/neuron"
   },
-  "homepage": "./",
+  "homepage": "https://www.nervos.org",
```
Keith-CY: This field sets the root of the web app, not the homepage of the project.

classicalliu: Sounds we can remove this line, no url for desktop app.

ashchan: True

Keith-CY: Not sure, the CRA fetch static files from `/` by default. And the Electron serves static files with the file system. Namely, the assets will be fetched from `file:///styles/xxx.css` which leads to errors. No verified. Please have a try.

PR #299, opened 2019-05-05
refactor: sync transactions in a single renderer process

classicalliu:
1. move `initConnection` to `createLoopTask`
2. add a renderer process to sync data from chain


Keith-CY:
Why using a renderer process instead of a child process to subscribe to new blocks?

classicalliu:
renderer process is most easy to use process, cluster is hard to call electron APIs and service workers can't use native modules

packages/neuron-wallet/src/utils/createLoopTask.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// create a background task to sync transactions
+// this task is a renderer process
+const createLoopTask = () => {
+  let loopWindow: BrowserWindow | null = new BrowserWindow({
+    x: 10,
+    y: 10,
+    show: false,
+    webPreferences: {
+      nodeIntegration: false,
```
Keith-CY: use `nodeIntegrationInWorker` instead of `nodeIntegration` according to https://electronjs.org/docs/tutorial/multithreading#multi-threaded-nodejs, as required in the past, the `nodeIntegration` should be `false`

Keith-CY: According to the doc https://electronjs.org/docs/tutorial/multithreading#multi-threaded-nodejs, the `nodeIntegrationInWorker` should be set to `true`

classicalliu: It seems can set to true, but not must set to true, and why we should set to true

Keith-CY: > It seems can set to true, but not must set to true, and why we should set to true

I'm not sure, doesn't the native module requires node runtime?

classicalliu: It works well, we can add this if we need in the future.

packages/neuron-wallet/src/utils/createLoopTask.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+  })
+
+  loopWindow.loadURL(env.mainURL)
```
ashchan: Must a window (renderer progress) load our main URL?
Can we just load an empty URL or page to avoid unnecessary UI and data loading?

classicalliu: I'll check it.

classicalliu: Replace `env.mainURL` with an empty file.

classicalliu:
> I seriously don't think we should throw app initializing logic into `utils`.

I move this file to `startup`, and I'll move startup logic file to `startup` later in another PR.

PR #300, opened 2019-05-05
feat: use real addresses instead of mock data

Keith-CY:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -209,9 +211,16 @@ class WalletsController {
...
   public static getActive = () => {
     try {
+      const activeWallet = WalletsController.service.getActive()
       return {
         status: ResponseCode.Success,
-        result: WalletsController.service.getActive(),
+        result: {
+          ...activeWallet,
+          addresses: {
+            receiving: activeWallet.addresses.receiving.map(addr => `0x${addr.address}`),
```
duanyytop: `Bech32` format address is not hex string, so `0x` should be removed .

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/300/commits/8db25de39c03e0554e9f30fc2be2d673e1129ee6

PR #303, opened 2019-05-06
feat: broadcast when addresses used

classicalliu:
When find addresses used in fetched transactions, broadcast this info.

### DIFF
--- a/packages/neuron-wallet/src/services/syncBlocks.ts
+++ b/packages/neuron-wallet/src/services/syncBlocks.ts
@@ -223,13 +223,16 @@ export default class SyncBlocksService {
   async resolveTx(transaction: Transaction) {
     const outputs: Cell[] = this.filterOutputs(transaction.outputs!)
     const anyInput: boolean = await SyncBlocksService.anyInput(transaction.inputs!)
-    if (outputs.length > 0 || anyInput) {
+
+    if (outputs.length > 0) {
       // found addresses used
       const addresses: string[] = outputs.map(output => {
         return TransactionsService.lockScriptToAddress(output.lock)
       })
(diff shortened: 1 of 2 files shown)


packages/neuron-wallet/src/services/syncBlocks.ts
```diff
@@ -217,15 +221,22 @@ export default class SyncBlocksService {
 
   // resolve single transaction
   async resolveTx(transaction: Transaction) {
-    const anyOutput: boolean = this.anyOutput(transaction.outputs!)
+    const outputs: Cell[] = this.filterOutputs(transaction.outputs!)
     const anyInput: boolean = await SyncBlocksService.anyInput(transaction.inputs!)
-    if (anyOutput || anyInput) {
-      TransactionsService.saveFetchTx(transaction)
+    if (outputs.length > 0 || anyInput) {
```
Keith-CY: anyInput is useless in this block

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -535,4 +535,17 @@ export default class TransactionsService {
 
     return lockHash
   }
+
+  public static lockScriptToAddress = (lock: Script): string => {
+    const blake160: string = lock.args![0]
+    return TransactionsService.blake160ToAddress(blake160)
+  }
+
+  public static blake160ToAddress = (blake160: string): string => {
```
Keith-CY: The method name seems weird @ashchan

packages/neuron-wallet/src/services/syncBlocks.ts
classicalliu: Also need to save tx if any inputs

PR #304, opened 2019-05-06
feat: add create method in wallet controller and display correct mess…

Keith-CY:
…ages on importing/creating/upda


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -95,6 +96,29 @@ class WalletsController {
     }
   }
 
+  public static create({
+    name,
+    password,
+    mnemonic,
+    receivingAddressNumber = 17,
```
ashchan: I have thought we've changed `17` and `3` already?

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/304/commits/99e22208326928df20032a786382f390d033c283

PR #314, opened 2019-05-09
feat: update depedencies and lock the versions

Keith-CY:


ashchan:
Restarted CI build.

duanyytop:
Keys test case fail because of async

Keith-CY:
CI passed

packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -290,7 +290,7 @@ export default class TransactionsService {
     const previousOutputs: OutputEntity[] = []
     for (const i of transaction.inputs!) {
       const input = new InputEntity()
-      input.outPointHash = i.previousOutput.hash
+      input.outPointTxHash = i.previousOutput.txHash
```
classicalliu: It's this version of SDK is still `outPoint.txHash` or `outPoint.cell.txHash` ?

Keith-CY: > It's this version of SDK is still `outPoint.txHash` or `outPoint.cell.txHash` ?

prerelease of v0.11, which use outpoint of `{ txHash, index }`

packages/neuron-wallet/package.json
```diff
@@ -25,7 +25,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged & yarn test"
```
classicalliu: `&` or `&&`? But I think test can remove from `pre-commit`, CI will check it.

packages/neuron-wallet/src/services/transactions.ts
classicalliu: ok, i c

packages/neuron-wallet/package.json
Keith-CY: `&&` is correct.

CI usually blocked

PR #315, opened 2019-05-10
fix: wallet name can not be same

BaojunCZ:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -77,6 +77,13 @@ class WalletsController {
     changeAddressNumber: number
   }): ChannelResponse<WalletData> => {
     try {
+      const id = WalletsController.service.getAll().findIndex(wallet => wallet.name === name)
+      if (id !== -1) {
+        return {
+          status: ResponseCode.Fail,
+          msg: 'Same name',
```
Keith-CY:
```suggestion
          msg: 'Wallet name existed',
```

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -77,6 +77,13 @@ class WalletsController {
     changeAddressNumber: number
   }): ChannelResponse<WalletData> => {
     try {
+      const id = WalletsController.service.getAll().findIndex(wallet => wallet.name === name)
```
classicalliu: Check this in data level is better, data level should ensure wallet name not duplicated.

BaojunCZ: I will move to wallet services.Store level is atom.

classicalliu: If service throw an error, does controller should catch it and return status and msg ? I'm not sure, @Keith-CY please check it, 3Q

Keith-CY: It will be caught in walletController.create

PR #317, opened 2019-05-10
Refactor wallet store

ashchan:
Remove electron-store and refactor wallet store to separate its main wallet list file and keystore files for each wallet.

## Breaking Notice

It will not load your existing data!!!


Keith-CY:
CI failed

ashchan:
Resolving conflicts...

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -1,66 +1,191 @@
...
+    fs.writeFileSync(this.storeLocation(), JSON.stringify(keystore), { encoding: 'utf8' })
+  }
+
+  deleteKeystore = () => {
+    fs.unlinkSync(this.storeLocation())
+  }
+}
+
+// TODO: Check if '/dev/wallets' path works on Windows
+const defaultStorePath = env.isDevMode ? '/dev/wallets' : '/wallets'
```
classicalliu: maybe can use `path`

Keith-CY: `path` module in Node.js

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -1,66 +1,191 @@
...
 export default class WalletService {
-  public getAll = (): WalletData[] => {
-    return walletStore.getAllWallets()
+  private storePath: string
+  private listStore: Store // Save wallets (meta info except keystore, which is persisted separately)
+  private walletsKey = 'wallets'
+  private currentWalletKey = 'current'
+
+  constructor(storePath: string = defaultStorePath) {
+    this.storePath = `${app.getPath('userData')}/${storePath}`
```
Keith-CY: use `path.join`

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -1,66 +1,191 @@
...
-      throw Error('Wallet name existed')
+  public get = (id: string): Wallet | undefined => {
+    const wallet = this.getAll().find(w => w.id === id)
+    if (wallet) {
+      return FileKeystoreWallet.fromJSON(wallet, this.storePath)
     }
-    const id = v4()
-    walletStore.saveWallet({ id, name, keystore, addresses })
-    return { id, name, keystore, addresses }
+    return undefined
```
Keith-CY: better to use `null` which explicitly means `none`

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -1,66 +1,191 @@
...
   }
 
-  public setActive = (id: string): boolean => {
-    return walletStore.setActiveWallet(id)
+  public getCurrent = (): Wallet | undefined => {
+    const walletId = this.listStore.readSync(this.currentWalletKey) as string
+    if (walletId) {
+      return this.get(walletId)
+    }
+    return undefined
```
Keith-CY: better to use `null`

Keith-CY:
conflicts

packages/neuron-wallet/src/services/wallets.ts
ashchan: Makes sense. Doing that in a future refactoring.

PR #319, opened 2019-05-11
refactor(neuron-wallet): refactor node service

Keith-CY:
1. move ckb core into node service;
2. add network switching on active network updating;
3. check active network on initApp;
4. add basic check on the network url, protocol required;


Keith-CY:
The CI failed due to an import which not exists. Interesting.

PR #320, opened 2019-05-11
refactor(neuron-wallet): refactor node service

Keith-CY:
1. move ckb core into node service;
2. add network switching on active network updating;
3. check active network on initApp;
4. add basic check on the network url, protocol required;


packages/neuron-wallet/src/startup/initApp.ts
```diff
@@ -21,14 +21,24 @@ const syncConnectStatus = () => {
...
-  nodeService.start()
   // TODO: this function should be moved to somewhere syncing data
   syncConnectStatus()
   WalletChannel.start()
   const activeId = await NetworksController.service.activeId()
   if (!activeId) {
     await NetworksController.service.init()
   }
+  if (!nodeService.core.node.url) {
+    const id = await NetworksController.service.activeId()
```
classicalliu: why not use service instead of controller

Keith-CY: > why not use service instead of controller

Yeah, it's indeed the service instance, which has been mounted on the controller.

classicalliu: I means use service directly, not use `controller.service`

Keith-CY: > I means use service directly, not use `controller.service`

The service was instantiated in the Controller, you mean instantiating one more here?

Keith-CY: or just create one somewhere, and controller also use that one

classicalliu: I think it's better to remove controller in logic level, but we can do this later

Keith-CY: make sense. I'm going to use DI in the future.

PR #323, opened 2019-05-13
Fix wallet store

BaojunCZ:
-  check wallet name exist
-  refactor wallet store test
- broadcast active after creating first wallet

### DIFF
--- a/packages/neuron-wallet/src/controllers/wallets.ts
+++ b/packages/neuron-wallet/src/controllers/wallets.ts
@@ -84,6 +84,9 @@ class WalletsController {
         addresses: key.addresses!,
       })
       windowManage.broadcast(Channel.Wallets, WalletsMethod.GetAll, WalletsController.getAll())
+      if (WalletsController.service.getAll().length === 1) {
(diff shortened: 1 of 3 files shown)


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -84,6 +84,9 @@ class WalletsController {
         addresses: key.addresses!,
       })
       windowManage.broadcast(Channel.Wallets, WalletsMethod.GetAll, WalletsController.getAll())
+      if (WalletsController.service.getAll().length === 1) {
```
Keith-CY: the broadcast of active wallet should be invoked when the active wallet updated, nothing related to the import mnemonic

BaojunCZ: I've added active wallet verification.There are two places to set active wallet.So I broadcast in these places.If I broadcast in service with windowManager,the wallet store service test and address service test should add 'electon-enviroment' to test.It will make things complicated.

PR #325, opened 2019-05-13
feat(neuron-wallet): store networks.json to userData/dev/ when env.is…

Keith-CY:
…DevMode is true


Keith-CY:
We're going to config the basePath in the Store module

PR #326, opened 2019-05-13
refactor(neuron-wallet): fix cannot find neuron dir on creating netwo…

Keith-CY:
…rks.json

check the directory on creating networks.json for quick fix, should check the initializing progress to make sure the networks.json will be created after the neuron dir created.

### DIFF
--- a/packages/neuron-wallet/src/utils/store.ts
+++ b/packages/neuron-wallet/src/utils/store.ts
@@ -1,8 +1,7 @@
 import fs from 'fs'
 import path from 'path'
 import EventEmitter from 'events'
-
-import console = require('console')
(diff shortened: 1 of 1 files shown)


packages/neuron-wallet/src/utils/store.ts
```diff
@@ -15,10 +14,17 @@ class Store extends EventEmitter {
 
   constructor(pathname: string, filename: string) {
     super()
+    if (!fs.existsSync(pathname)) {
```
ashchan: Please make sure we remove this after we figure out the loading order issue. This is a potential security issue to create the top level folder.

PR #327, opened 2019-05-13
Remove change password feature

BaojunCZ:


packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -39,31 +34,20 @@ export default ({
...
-  }, [dispatch])
-
   const handleConfirm = useCallback(() => {
     toggleDialog(false)
     dispatch(
       actionCreators.updateWallet({
         id: wallet.id,
-        password: editor.password.value,
-        newPassword: editor.newPassword.value,
+        password: '',
```
Keith-CY: passwords related fields could be removed too.

packages/neuron-ui/src/components/WalletEditor/index.tsx
```diff
@@ -39,31 +34,20 @@ export default ({
...
+        newPassword: '',
```
ashchan: Won't this trigger the `newPassword` check in the controller?

BaojunCZ: yes

PR #329, opened 2019-05-14
feat(neuron-wallet): refactor store module to handle recursive direct…

Keith-CY:
…ories.

create a new file service to handle files accessing, better than the native file system module.


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -93,7 +93,7 @@ export default class WalletService {
   constructor(storePath: string = defaultStorePath) {
     this.storePath = `${app.getPath('userData')}/${storePath}`
     fs.mkdirSync(this.storePath, { recursive: true })
-    this.listStore = new Store(this.storePath, 'wallets.json')
+    this.listStore = new Store('wallet', 'wallets.json')
```
ashchan: Also need to take care of `this.storePath`, which will be used to store keystore files. `wallets.json` and all keystore files should be stored int he same `wallets` (not `wallet`) folder.

packages/neuron-wallet/src/env.ts
```diff
@@ -14,6 +15,7 @@ interface ENV {
 }
 const env: ENV = {
   isDevMode,
+  fileBasePath: path.resolve(app.getPath('userData'), isDevMode ? 'dev/' : ''),
```
ashchan: Check for test env as well?

Keith-CY: > Check for test env as well?

Yes, we might run e2e tests locally, this will avoid erasing production data.

ashchan: OK I didn't describe it clearly. I meant that we should have separate default folder for TEST environment from from development and production. For example:

* Production: ``
* Development: `dev/`
* Test: `test/`

This way test won't affect local development installation.

This might require change of `isDevMode` (to bring back `NODE_ENV` etc.).

packages/neuron-wallet/src/services/wallets.ts
Keith-CY: fixed in https://github.com/nervosnetwork/neuron/pull/329/commits/fdbaf47ab5f7e277dbeb645e2808227217ac8b18

ashchan: On LN 94 `this.storePath` was constructed with the tased in param. Here we need to do to things:

* Support `test/` env and default folder location as I described in the other comment.
* Remove the `storePath` param so wallet list and keystore files use the same `wallets` folder.

packages/neuron-wallet/src/env.ts
Keith-CY: > OK I didn't describe it clearly. I meant that we should have separate default folder for TEST environment from from development and production. For example:
> 
> * Production: ``
> * Development: `dev/`
> * Test: `test/`
> 
> This way test won't affect local development installation.
> 
> This might require change of `isDevMode` (to bring back `NODE_ENV` etc.).

Agree, that's why I added basePath in the config, not in the code

packages/neuron-wallet/src/services/wallets.ts
Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/329/commits/5c18936eac71753db64f1b4bcf35449c970323d1 and https://github.com/nervosnetwork/neuron/pull/329/commits/d69b0c2f5b50b169de76e2d5ab7a67dcd34a6ce7, use NODE_EVN=test for testing which will create /test/ dir in the store dir

PR #330, opened 2019-05-14
Refactor keystore

ashchan:
Fix keystore version to 3 altough we don't guarantee the compatibility with V3 Keystore format used by Ethereum and some other implementations


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -62,13 +58,22 @@ class FileKeystoreWallet implements Wallet {
...
+    const data = new FileService().readFileSync(MODULE_NAME, this.keystoreFileName())
     return JSON.parse(data) as Keystore
   }
 
-  clear = () => {
-    this.keyStore.clear()
+  saveKeystore = (keystore: Keystore) => {
+    const keystoreToSave = keystore
+    keystoreToSave.id = this.id
+    new FileService().writeFileSync(MODULE_NAME, this.keystoreFileName(), JSON.stringify(keystoreToSave))
```
Keith-CY: `fileService` could be imported from `src/startup/fileService`

ashchan: What's the difference? `startup/fileService` is a singleton?

PR #331, opened 2019-05-14
Refactor sync blocks

classicalliu:
1. cache `blockHeaderForCheck`
2. cache `currentBlockNumber`
3. listen `tipBlockNumber` by `nodeService` to replace monitor by self
4. listen to network switch
5. reduce fetch size when failed 3 times in a row

### DIFF
--- a/packages/neuron-wallet/src/services/syncBlocks.ts
+++ b/packages/neuron-wallet/src/services/syncBlocks.ts
@@ -8,7 +8,7 @@ import SyncInfoEntity from '../entities/SyncInfo'
 import nodeService from '../startup/nodeService'
 import { networkSwitchSubject } from './networks'
 
-// FIXME: now have some problem with core, should should update every time network switched
+// FIXME: now have some problem with core, should update every time network switched
 // const { core } = nodeService
 let core: Core
 networkSwitchSubject.subscribe(network => {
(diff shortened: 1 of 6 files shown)


packages/neuron-wallet/src/services/syncBlocks.ts
```diff
@@ -1,12 +1,21 @@
...
 import { Script, OutPoint, Cell } from './cells'
 import TransactionsService, { Input, Transaction } from './transactions'
 import OutputEntity from '../entities/Output'
 import SyncInfoEntity from '../entities/SyncInfo'
 import nodeService from '../startup/nodeService'
-
-const { core } = nodeService
+import { networkSwitchSubject } from './networks'
+
+// FIXME: now have some problem with core, should should update every time network switched
```
ashchan:
```suggestion
// FIXME: now have some problem with core, should update every time network switched
```

packages/neuron-wallet/src/services/syncBlocks.ts
```diff
@@ -31,18 +40,33 @@ export default class SyncBlocksService {
 
   private fetchSize = 128
 
-  private minFetchSize = 2
-
-  private sizeForCheck = 5
+  private minFetchSize = 16
```
ashchan: Now that we have `lines-between-class-members` eslint rule configured, I guess we can remove some blank lines between related members(properties).

classicalliu: updated

PR #335, opened 2019-05-14
refactor: extract types as an individual module

classicalliu:


packages/neuron-wallet/src/appTypes/types.ts
```diff
@@ -0,0 +1,63 @@
+// define types in app
```
Keith-CY: type definition file could be renamed to `xxx.d.ts` and put in `types` dir

packages/neuron-wallet/src/appTypes/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  capacity: string
+  data?: string
+  lock: Script
+  type?: Script | null
+  outPoint?: OutPoint
+  status?: string
+  lockHash?: string
+}
+
+export interface OutPoint {
```
Keith-CY: Better name it `CellOutPoint` to keep consistent with types in CKB

classicalliu: It's still `OutPoint`, and v0.12 will add `CellOutPoint` and update `OutPoint`

PR #345, opened 2019-05-20
refactor: add LockUtils

classicalliu:
Using LockUtils to deal with lock/lockHash/address/blake160 convert

### DIFF
--- a/packages/neuron-wallet/src/utils/lockUtils.ts
+++ b/packages/neuron-wallet/src/utils/lockUtils.ts
@@ -1,0 +1,96 @@
+import nodeService from '../startup/nodeService'
+import { CellOutPoint, OutPoint, Script } from '../appTypes/types'
+
+const { core } = nodeService
+
+export default class LockUtils {
+  static systemScriptInfo: { codeHash: string; outPoint: OutPoint } | undefined
+
+  static async systemScript() {
+    if (this.systemScriptInfo) {
+      return this.systemScriptInfo
+    }
+
+    const genesisHash: string = await core.rpc.getBlockHash('0')
+    const genesisBlock = await core.rpc.getBlock(genesisHash)
+    const systemScriptTx = genesisBlock.transactions[0]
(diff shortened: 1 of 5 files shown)


packages/neuron-wallet/src/utils/lockUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+export default class LockUtils {
+  static systemScriptInfo: { codeHash: string; outPoint: OutPoint } | undefined
+
+  static async systemScript() {
+    if (this.systemScriptInfo) {
+      return this.systemScriptInfo
+    }
+
+    const genesisHash: string = await core.rpc.getBlockHash('0')
+    const genesisBlock = await core.rpc.getBlock(genesisHash)
```
ashchan: We should use `getBlockByNumber('0')` to reduce one RPC call.

packages/neuron-wallet/src/utils/lockUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (typeof data === 'string') {
+      blake2b.update(core.utils.hexToBytes(data))
+    } else {
+      // if Uint8Array
+      blake2b.update(data)
+    }
+    const codeHash: string = blake2b.digest('hex')
+    const cellOutPoint: CellOutPoint = {
+      txHash: systemScriptTx.hash,
+      index: '0',
```
ashchan: If I recall correctly the index should be `1` instead?

packages/neuron-wallet/src/utils/lockUtils.ts
classicalliu: updated, thanks

packages/neuron-wallet/src/utils/lockUtils.ts
classicalliu: updated

Keith-CY:
conflicts

PR #346, opened 2019-05-20
Verify password complexity

BaojunCZ:


packages/neuron-ui/src/utils/validators.ts
```diff
@@ -13,7 +13,7 @@ export const verifyWalletSubmission = ({
   confirmPassword: string
   name: string
 }) => {
-  return password && name && password === confirmPassword
+  return password && name && password === confirmPassword && password.length >= 8 && confirmPassword.length >= 8
```
classicalliu: no need for `confirmPassword.length >= 8` after check `password === confirmPassword`
```suggestion
  return password && name && password === confirmPassword && password.length >= 8
```

BaojunCZ: ok

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -76,6 +76,12 @@ class WalletsController {
     receivingAddressNumber: number
     changeAddressNumber: number
   }): Promise<ChannelResponse<Wallet>> => {
+    if (!WalletsController.verifyPasswordComplexity(password)) {
+      return {
+        status: ResponseCode.Fail,
+        msg: 'At least 8 characters, three types of uppercase and lowercase alphabets, numbers, and special symbols!',
```
ashchan: To make the verification and message more clear, we should separate the results. e.g., check these:

* password !== confirmPassword,
* password length < required length
* password complexity < required conditions

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -178,6 +217,13 @@ class WalletsController {
         }
         if (newPassword) {
           if (WalletsController.service.validate({ id, password })) {
+            if (!WalletsController.verifyPasswordComplexity(newPassword)) {
+              return {
+                status: ResponseCode.Fail,
+                msg:
+                  'At least 8 characters, three types of uppercase and lowercase alphabets, numbers, and special symbols!',
```
ashchan: The message appears in multiple places, and is not localized.

BaojunCZ: ok

packages/neuron-ui/src/utils/validators.ts
```diff
@@ -13,7 +13,7 @@ export const verifyWalletSubmission = ({
   confirmPassword: string
   name: string
 }) => {
-  return password && name && password === confirmPassword
+  return password && name && password === confirmPassword && password.length >= 8
```
Keith-CY: use `MIN_PASSWORD_LENGTH` instead of `8`

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -156,6 +159,32 @@ class WalletsController {
     }
   }
 
+  public static verifyPasswordComplexity = (password: string) => {
+    if (password.length < 8) {
```
Keith-CY: use `MIN_PASSWORD_LENGTH` instead of `8`

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -170,7 +161,7 @@ class WalletsController {
 
   public static verifyPasswordComplexity = (password: string) => {
     if (password.length < 8) {
-      return false
+      throw Error(i18n.t('messages.wallet-password-at-least-8-characters'))
```
ashchan: Please change this to:

1. Use constant for the length (`8`).
2. Use that constant from the `i18n` message.

packages/neuron-wallet/src/locales/en.ts
```diff
@@ -36,6 +36,9 @@ export default {
     messages: {
       'failed-to-load-networks': 'Failed to load networks',
       'Networks-will-be-reset': 'Networks will be reset',
+      'wallet-password-at-least-8-characters': 'Password must be at least 8 characters',
+      'wallet-password-at-least-3-types':
+        'Password must be three types of uppercase and lowercase alphabets, numbers, and special symbols',
```
ashchan:
```suggestion
        'Password must contain a combination of uppercase and lowercase letters, numbers and special symbols.',
```

packages/neuron-wallet/src/locales/en.ts
```diff
@@ -36,6 +36,9 @@ export default {
     messages: {
       'failed-to-load-networks': 'Failed to load networks',
       'Networks-will-be-reset': 'Networks will be reset',
+      'wallet-password-at-least-8-characters': 'Password must be at least 8 characters',
+      'wallet-password-at-least-3-types':
```
ashchan:
```suggestion
      'wallet-password-letter-complexity':
```

Keith-CY:
Should we limit the password up to a maximum length?

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -156,6 +161,32 @@ class WalletsController {
     }
   }
 
+  public static verifyPasswordComplexity = (password: string) => {
+    if (password.length < MIN_PASSWORD_LENGTH) {
+      throw Error(i18n.t('messages.wallet-password-less-than-min-length'))
```
Keith-CY: use the constant in the message means use the i18n interpolation, there's the [guide](https://www.i18next.com/translation-function/interpolation)

ashchan:
> Should we limit the password up to a maximum length?

Perhaps. Make it big enough but still practical, say 50?

packages/neuron-wallet/src/controllers/wallets.ts
ashchan: Indeed.

Keith-CY:
> > Should we limit the password up to a maximum length?
> 
> Perhaps. Make it big enough but still practical, say 50?

Agree

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -18,6 +19,9 @@ export enum WalletsMethod {
   SendCapacity = 'sendCapacity',
 }
 
+export const MIN_PASSWORD_LENGTH = 8
```
Keith-CY: has been defined in `utils/const`

PR #364, opened 2019-05-22
fix(neuron-wallet): fix base path of logger module by path module

Keith-CY:


packages/neuron-wallet/src/utils/logger.ts
```diff
@@ -1,9 +1,10 @@
 import { app } from 'electron'
 import winston, { format } from 'winston'
+import path from 'path'
 import env from '../env'
 
 const { isDevMode } = env
-const path = isDevMode ? '' : `${app.getPath('logs')}/`
+const basePath = isDevMode ? './' : `${app.getPath('logs')}`
```
classicalliu: does `./` works on windows ?

Keith-CY: > does `./` works on windows ?

It will be parsed by `path` module.

PR #367, opened 2019-05-23
refactor(neuron-wallet): refactor helpers controller and networks con…

Keith-CY:
…troller with CatchControllerErr

1. Handle controller errors in CatchControllerError

### DIFF
--- a/packages/neuron-wallet/src/controllers/helpers.ts
+++ b/packages/neuron-wallet/src/controllers/helpers.ts
@@ -12,13 +12,6 @@ export enum HelpersMethod {
  * @description handle messages from helpers channel
  */
 class HelpersController {
-  /**
-   * @method generateMnemonic
-   * @static
-   * @memberof HelpersController
-   * @description generate mnemonic
-   */
-
   @CatchControllerError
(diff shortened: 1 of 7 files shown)


packages/neuron-wallet/src/controllers/helpers.ts
```diff
@@ -1,22 +1,34 @@
...
   GenerateMnemonic = 'generateMnemonic',
 }
+
+/**
+ * @class HelpersController
+ * @description handle messages from helpers channel
+ */
 class HelpersController {
-  public static generateMnemonic = (): Controller.Response<string> => {
+  /**
```
ashchan: I'm not so sure about this but these comments don't really add many values. The scope and function names already describe themselves well enough.

Keith-CY: removed comments on methods of helpers controller and networks controller, but left comments on the controller classes for reminding us the controllers are used to handle messages from channels, not network requests.
https://github.com/nervosnetwork/neuron/pull/367/commits/bf3d7ba826846953acd7a0e2a7546c79944cd26f

PR #369, opened 2019-05-24
Implement BIP32

ashchan:
to:

* Remove `bip32.js` dependency
* Thus remove `tiny-secp256k1` and use `elliptic` (which CKB SDK uses)
  - Resolve build error issue w/ Node 12

## Note

This is not supposed to be a full implementation. As long as it meets need for our daily internal usage, we can skip many features and checks.


ashchan:
Cancelled CI for  b0544e2 as that's known changes breaking tests.

PR #372, opened 2019-05-24
refactor(neuron-wallet): refactor wallets controller with CatchContro…

Keith-CY:
…llerError

1. refactor wallets controller with CatchControllerError
2. move verifyPasswordComplexity to utils.validators
3. add throwing errors in wallets service


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -1,9 +1,11 @@
 import WalletsService, { Wallet, WalletProperties } from '../services/wallets'
 import { ResponseCode } from './index'
 import windowManage from '../utils/windowManage'
-import { Channel, MIN_PASSWORD_LENGTH, MAX_PASSWORD_LENGTH } from '../utils/const'
+import { Channel } from '../utils/const'
 import Key from '../keys/key'
+import { CatchControllerError } from '../utils/decorators'
```
ashchan: Let’s move decorators and validators out of utils. They deserve their own top level folders. This will also do a diet for the utils.

PR #376, opened 2019-05-27
feat(neuron-wallet): reduce wallet info passed to ui

Keith-CY:
filter the path of active wallet addresses, and filter addresses of other wallets

### DIFF
--- a/packages/neuron-ui/src/contexts/NeuronWallet/wallet.ts
+++ b/packages/neuron-ui/src/contexts/NeuronWallet/wallet.ts
@@ -1,8 +1,8 @@
-export interface PlainWallet {
+export interface WalletIdentity {
   name: string
   id: string
 }
-export interface Wallet extends PlainWallet {
+export interface Wallet extends WalletIdentity {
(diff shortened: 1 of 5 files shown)


packages/neuron-ui/src/contexts/NeuronWallet/wallet.ts
```diff
@@ -1,6 +1,8 @@
-export interface Wallet {
+export interface PlainWallet {
```
ashchan: Based on the properties of this new interface, `PlainWallet` might not be suitable as it doesn't really have wallet related properties. Do you think `WalletIdentity` is a better choice?

If we use `WalletIdentity`, then it becomes natural when we load a list of wallet identities instead of `plain` wallets from `Settings` or other places.

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/376/commits/5dd588058d46e30918522306187e4cd65b4a6b21.

PR #381, opened 2019-05-29
Move sync to renderer

classicalliu:
* move sync logic to `syncTask.ts`, it runs in renderer process
* move `NetworksMethod` to service from controller, it will cause some problems in renderer process (service require controller and controller require service)
* also `initConnection` in main process
* using `ncp` for cross platform `cp`


packages/neuron-wallet/package.json
```diff
@@ -81,6 +82,7 @@
     "electron-devtools-installer": "2.2.4",
     "husky": "1.3.1",
     "lint-staged": "8.1.5",
+    "ncp": "^2.0.0",
```
Keith-CY: specify the version explicitly.

packages/neuron-wallet/src/startup/loopTask/syncTask.ts
```diff
@@ -0,0 +1,62 @@
...
+    }),
+  )
+  return lockHashes
+}
+
+// call this after network switched
+// TODO: listen to network switch
+export const switchNetwork = async (networkId: string) => {
+  // stop all blocks service
+  stopLoopSubject.next('stop')
```
Keith-CY: Does `syncBlocksService.stop()` work here?

packages/neuron-wallet/src/services/syncBlocks.ts
```diff
@@ -1,19 +1,23 @@
...
-import { networkSwitchSubject } from './networks'
 import LockUtils from '../utils/lockUtils'
 import TypeConvert from '../appTypes/typeConvert'
 
+const { app }: { app: any } = remote
+const { syncTask } = app
+const { networkSwitchSubject } = syncTask
+
 // FIXME: now have some problem with core, should update every time network switched
 // const { core } = nodeService
```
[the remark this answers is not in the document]
classicalliu: It's not useless, it should be fixed and it's the correct way

packages/neuron-wallet/src/startup/loopTask/create.ts
```diff
@@ -1,78 +1,68 @@
...
+    y: windowState.y,
+    width: windowState.width,
+    height: windowState.height,
+    minWidth: 800,
+    minHeight: 600,
     show: false,
+    backgroundColor: '#e9ecef',
     webPreferences: {
-      nodeIntegration: false,
+      nodeIntegration: true,
```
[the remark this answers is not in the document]
classicalliu: Not web worker

packages/neuron-wallet/src/startup/loopTask/syncTask.ts
classicalliu: yes

ashchan:
@classicalliu Please kindly rebase or merge from upstream to resolve the conflicts.

packages/neuron-wallet/src/startup/loopTask/create.ts
```diff
@@ -1,78 +1,67 @@
...
-    loopWindow!.hide()
-
-    // TODO: add an event on networkId for when it init and changed, it should broadcast a message with networkId
-    networkSwitchSubject.subscribe(async network => {
-      if (network) {
-        await switchNetwork(network.name)
-      }
-    })
+    if (env.isDevMode && process.env.DEV_SYNC_TASK) {
+      loopWindow!.show()
```
ashchan: Is this for local debug purpose? And why do we need to persist the state of the loop window, despite the fact this will be a hidden window for background process?

BTW, `loopWindow` doesn't describe its responsibility (e.g. `txSyncBackgroundWindow`) but its implementation detail (why the outer modules need to know there would be a loop inside?).

classicalliu: Yes, it's for debug.(database log will show in window console)
It's right we should remove state for window.
Maybe `loopWindow` is not a good name, I'll rename it.

PR #382, opened 2019-05-29
refactor(neuron-wallet): refactor networks service

Keith-CY:
1. add validation on parameters;
2. add tests of netowrks service

### DIFF
--- a/packages/neuron-wallet/src/services/wallets.ts
+++ b/packages/neuron-wallet/src/services/wallets.ts
@@ -41,9 +41,9 @@ class FileKeystoreWallet implements Wallet {
   public addresses: Addresses
 
   constructor(id: string, { name, addresses }: WalletProperties) {
-    if (id === undefined) throw new Error(i18n.t('is-required', { field: 'id' }))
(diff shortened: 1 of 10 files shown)


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -150,7 +150,7 @@ export default class WalletService {
   }
 
   public get = (id: string) => {
-    if (id === undefined) throw new Error(i18n.t('is-required', { field: 'id' }))
+    if (id === undefined) throw new Error(i18n.t('messages.is-required', { field: 'id' }))
```
ashchan: Under what circumstance would we pass in `undefined` id?

Keith-CY: When the neuron-ui has bugs that fetching a wallet with `id=undefined`

ashchan: Merging this. But is there any way we can prevent this kind of UI bug from happening?

PR #386, opened 2019-05-30
fix: broadcast addresses used event to main process from renderer pro…

classicalliu:
…cess

### DIFF
--- a/packages/neuron-wallet/src/subjects/addressesUsedSubject.ts
+++ b/packages/neuron-wallet/src/subjects/addressesUsedSubject.ts
@@ -1,0 +1,16 @@
+import { ReplaySubject } from 'rxjs'
+
+// subscribe this Subject to monitor which addresses are used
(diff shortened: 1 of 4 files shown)


packages/neuron-wallet/src/subjects/addressesUsedSubject.ts
```diff
@@ -0,0 +1,16 @@
+import { ReplaySubject } from 'rxjs'
+
+// subscribe this Subject to monitor which addresses are used
+export class AddressesUsedSubject {
```
Keith-CY: The name is a bit confusing, `AddressesUsedSubject` is not a subject, `AddresesUsedSubject.subject` and `AddressesUsedSubject.getSubject()` are subjects.

classicalliu: rename to `AddressesUsedEvent` ? or any advice

PR #392, opened 2019-05-30
feat(neuron-ui): generate wallet with number suffix

Keith-CY:


packages/neuron-ui/src/components/WalletWizard/index.tsx
```diff
@@ -144,14 +145,24 @@ const Submission = ({
...
+  const {
+    settings: { wallets },
+  } = useNeuronWallet()
   const { name, password, confirmPassword, imported } = state
   const [t] = useTranslation()
   const message = 'wizard.set-a-strong-password-to-protect-your-wallet'
 
   useEffect(() => {
+    const genName = (baseNum: number = 0): string => {
+      const walletName = `wallet ${baseNum}`
```
ashchan: Please localized this so that we have `Wallet 1` for English and `钱包 1` for Simplified Chinese.

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/392/commits/1bdff5b68235d9ff22a28928a66ee113c534c55b

PR #395, opened 2019-05-31
refactor(neuron-wallet): move init window method to app controller

Keith-CY:


packages/neuron-wallet/src/controllers/app/index.ts
```diff
@@ -1,9 +1,35 @@
-import { app, dialog, shell, Menu, MessageBoxOptions } from 'electron'
+import { app, dialog, shell, Menu, MessageBoxOptions, BrowserWindow } from 'electron'
 import { Channel, ResponseCode } from '../../utils/const'
 import windowManage from '../../utils/windowManage'
 import { URL, contextMenuTemplate } from './options'
+import WalletsController from '../wallets'
+import NetworksController from '../networks'
 
 export default class AppController {
+  public static initWindow = async (win: BrowserWindow) => {
```
ashchan: Kind of confusing here. It's called `initWindow` but there's much related to initializing the window...

Keith-CY: The name might not be so presentive. 

On the new window ready to pop up, this method used to send the current state to it as the initialized state.

The state includes the current wallet, wallet list, the current network, network list, and the locale information.

PR #399, opened 2019-06-02
chore: Apply small-case naming convention for neuron-wallet

ashchan:
Please stick to small-case naming convention for folders/files inside the `neuron-wallet` package.


packages/neuron-wallet/src/services/sync-blocks.ts
```diff
@@ -2,15 +2,15 @@ import { getConnection } from 'typeorm'
...
 import { Subject, BehaviorSubject } from 'rxjs'
 import Core from '@nervosnetwork/ckb-sdk-core'
-import { Script, OutPoint, Cell, Input, Transaction, Block, BlockHeader } from '../appTypes/types'
+import { Script, OutPoint, Cell, Input, Transaction, Block, BlockHeader } from '../app-types/types'
 import TransactionsService from './transactions'
-import OutputEntity from '../entities/Output'
-import SyncInfoEntity from '../entities/SyncInfo'
+import OutputEntity from '../entities/output'
+import SyncInfoEntity from '../entities/sync-info'
 import nodeService from '../startup/nodeService'
```
Keith-CY: `nodeService` => `node-service`

PR #401, opened 2019-06-03
feat(neuron-ui): hide the pagination when total count is less than or…

Keith-CY:
… equal to the page size

### DIFF
--- a/packages/neuron-ui/src/widgets/Table/index.tsx
+++ b/packages/neuron-ui/src/widgets/Table/index.tsx
@@ -54,15 +54,17 @@ export default ({
         ))}
       </tbody>
     </BsTable>
-    <Row>
-      <Col>
-        <Pagination
-          currentPage={pageNo}
-          pageSize={pageSize}
-          total={totalCount}
-          onChange={page => onPageChange(page, pageSize)}
-        />
-      </Col>
-    </Row>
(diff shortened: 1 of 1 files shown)


packages/neuron-ui/src/widgets/Table/index.tsx
```diff
@@ -54,15 +54,17 @@ export default ({
...
-      <Col>
-        <Pagination
-          currentPage={pageNo}
-          pageSize={pageSize}
-          total={totalCount}
-          onChange={page => onPageChange(page, pageSize)}
-        />
-      </Col>
-    </Row>
+    {totalCount <= pageSize ? null : (
```
ashchan: Does `totalCount > pageSize && (...)` work as this?

Keith-CY: It does, but the expression returns `false`, and the ternary returns `null`, they are slightly different in semantics.

Keith-CY: 1. `A && Component` returns A if A is false-like, which has side-effects in some cases.
E.g. `{items.length && <Items />}`, `Items` is expected to be hidden when `items.length === 0`, but `0`, which is false-like, is rendered.

2. `A ? Component : null` is under controller to show `nothing` or `Component`

PR #402, opened 2019-06-03
refactor: singleton for node & file service

classicalliu:
`startup/node-service` and `startup/file-service` removed


classicalliu:
> Can we have a unified way to define/create singleton services. After this update, will we still have other services' singletons created under `startup`? 😿

It seems `file-service` is also a singleton under startup.

ashchan:
> > Can we have a unified way to define/create singleton services. After this update, will we still have other services' singletons created under `startup`? 😿
> 
> It seems `file-service` is also a singleton under startup.

Right. So I hope we have one way to do the thing to avoid confusing :)

classicalliu:
> > > Can we have a unified way to define/create singleton services. After this update, will we still have other services' singletons created under `startup`? 😿
> > 
> > 
> > It seems `file-service` is also a singleton under startup.
> 
> Right. So I hope we have one way to do the thing to avoid confusing :)

Also apply singleton for `file-service`, no singleton instance under `startup` now.

packages/neuron-wallet/src/services/node.ts
```diff
@@ -55,4 +65,6 @@ class NodeService {
   }
 }
 
+export const nodeService = NodeService.getInstance()
```
Keith-CY: If is it better to use `getter`, and use its instance via `NodeService.instance`

classicalliu: any advice for variable name

Keith-CY: > any advice for variable name

No idea except for `ins` or `instance`

ashchan: As long as we use the same pattern, and have a consistent naming convention. `XxxService.getInstance()` looks just fine.

classicalliu: reset it to `getter` until we know how to name variables .....

Keith-CY: That's fine, but my point is to reference the instance via `NodeService.getInstance()` instead of export one.

But they are almost the same, feel free to merge.

ashchan: I see. I think @classicalliu was looking at `startup/file-service` and did the same export here. Let's change all to `XxxService.getInstance()`.

PR #403, opened 2019-06-03
refactor: Unify singleton usage

ashchan:
Use `XxxService.getInstance()` instead of exposing a singleton constant.


ashchan:
See [discussions](https://github.com/nervosnetwork/neuron/pull/402#discussion_r289689045)

PR #404, opened 2019-06-03
Using `getInstance()`

classicalliu:


classicalliu:
@Keith-CY @ashchan mean this ?

PR #412, opened 2019-06-06
feat(neuron-wallet): enable backup

Keith-CY:
enable backup of keystore


Keith-CY:
> Exporting should require password input. Although keystore is password encrypted it's still sensitive data to export.

add password validation in https://github.com/nervosnetwork/neuron/pull/412/commits/d30c9a8e6dc401e4fc8db469203619bc46daf270

PR #420, opened 2019-06-11
feat: Right align wizard buttons

ashchan:


packages/neuron-ui/src/widgets/ScreenButtonRow/index.tsx
```diff
@@ -1,5 +1,27 @@
...
+  display: flex;
+  padding-top: 30px;
+  justify-content: flex-start;
+
+  button {
+    margin-right: 20px;
+  }
+`
+
+export { RightScreenButtonRow, LeftScreenButtonRow }
```
Keith-CY: These two kinds of layout could be implemented by extending `ScreenButtonRow` with an `align` attribute.

ashchan: Indeed. Will change to that.

PR #433, opened 2019-06-13
New sync blocks module

classicalliu:
* `check-and-save` check a block's txs belongs to me, if true, save to database and notice address used
* `range-for-check` blockHeaders for check fork
* `block-number` get and update currentBlockNumber
* `get-blocks` get blocks by blockNumbers
* `queue` create a queue with startBlockNumber and endBlockNumber, and then sync them
* `block-listener` will listen tipBlockNumber and create queues


packages/neuron-wallet/src/services/sync/renderer-params.ts
```diff
@@ -0,0 +1,5 @@
+import { remote } from 'electron'
+
+const { app }: { app: any } = remote
+const { syncBlockTask } = app
```
ashchan: Instead of defining `syncBlockTask` on `app` object, we should try using `remote.require('sync-block-task')` and see if that works well.

classicalliu: export there params in `sync-block-task/params.ts` and require them by `remote.require`

PR #435, opened 2019-06-14
refactor(neuron-wallet): move init app to app controller

Keith-CY:
and move change-language, set-application-menu from init-window method to init-app method


Keith-CY:
> Why put init app logic in a controller ?

I regard the `ready` signal as a result of user input. Which tells the app to render UI(language, application menu).

Perhaps the `set network` should be put somewhere else.

Keith-CY:
how about setting the network on node service initialization

classicalliu:
I still confusing why `ready` signal is a user input event ?

Keith-CY:
> I still confusing why `ready` signal is a user input event ?

Eventually, the `ready` action is triggered by the user action of opening the app.

PR #439, opened 2019-06-14
feat(neuron-wallet): implement mnemonic module

Keith-CY:
Use the mnemonic module to replace the bip39 package in the future.


packages/neuron-wallet/src/utils/mnemonic/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      })
+    } catch (error) {
+      reject(error)
+    }
+  })
+}
+
+export function mnemonicToEntropy(mnemonic: string = '') {
+  const words = mnemonic.normalize('NFKD').split(' ')
+  if (words.length % 3 !== 0) throw new Error(INVALID_MNEMONIC)
```
ashchan: Practically beside `% 3 === 0`, the length could only be one of `12` - `24`. Would it be necessary to check that as well to catch length validation error early?

Keith-CY: > Practically beside `% 3 === 0`, the length could only be one of `12` - `24`. Would it be necessary to check that as well to catch length validation error early?

updated in https://github.com/nervosnetwork/neuron/pull/439/commits/7cddc8e5b72bb18dd0dcecaae8b301fbc84bb49c

PR #446, opened 2019-06-15
refactor(neuron-ui): remove qrcode.react

Keith-CY:
remove qrcode.react package


packages/neuron-ui/src/widgets/QRCode/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+        return
+      }
+
+      if (cell && start === null) {
+        start = x
+      }
+    })
+  })
+
+  return ops.join('')
```
classicalliu: Why need `join('')` here

Keith-CY: To concatenate sections into a continuous `path` which is acceptable to `svg`

PR #450, opened 2019-06-16
refactor(neuron-wallet): remove the electron-prompt package

Keith-CY:


packages/neuron-wallet/src/utils/prompt/password/index.ts
```diff
@@ -0,0 +1,44 @@
...
+const { language } = window.navigator
+
+const locale = (() => {
+  if (language === 'en-US' || language === 'en') {
+    return 'en-US'
+  }
+  return 'zh-CN'
+})()
+
+const labels: any = {
```
classicalliu: Can we move this to `locales` dir ?

Keith-CY: > Can we move this to `locales` dir ?

updated in https://github.com/nervosnetwork/neuron/pull/450/commits/3d11cad35ac3d1306abce7d8cfe40476c7a7fff0

PR #459, opened 2019-06-17
feat: sign tx with multi private key

classicalliu:
TODO:
- [ ] get address info from address module and get private key from key module (#466)


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -293,64 +285,85 @@ export default class WalletService {
...
+    const message = blake2b.digest()
+    const signature = `0x${addrObj.sign(message)}`
+    const newWitness: Witness = {
+      data: [`0x${addrObj.publicKey}`, signature],
+    }
+    return newWitness
+  }
+
+  public getPrivateKey = (path: string): string => {
+    if (path !== '') {
```
ashchan: Should path always be `` 🤯?

packages/neuron-wallet/src/utils/blake2b.ts
```diff
@@ -0,0 +1,27 @@
+import Core from '@nervosnetwork/ckb-sdk-core'
+import NodeService from '../services/node'
+
+export default class Blake2b {
+  private b: any
```
ashchan: Please give it a longer name.

packages/neuron-wallet/src/services/wallets.ts
classicalliu: It's a mocked path and updated to use real path format.

packages/neuron-wallet/src/utils/blake2b.ts
classicalliu: Renamed to `blake2b`

PR #493, opened 2019-06-21
refactor(neuron-wallet): slice '0x' in privateToPublic

Keith-CY:


ashchan:
Rejecting this as:

1. `Buffer.from` never accepts string with `0x` prefixed hex string (as I observed).
2. 66 is length of string key, not buffer.

Let's keep stripping the `0x` outside.

Will revert this from another PR.

PR #500, opened 2019-06-22
feat(neuron-ui): use shannons in send transaction api

Keith-CY:


packages/neuron-ui/src/containers/MainContent/actionCreators/send.ts
```diff
@@ -32,7 +33,14 @@ export default {
     if (invalid) {
       return errorAction
     }
-    walletsCall.sendCapacity({ id, items })
+    walletsCall.sendCapacity({
+      id,
+      items: items.map(item => ({
+        address: item.address,
+        amount: CKBToShannonFormatter(item.amount, item.unit),
```
classicalliu: controller use capacity, not amount

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/500/commits/0edd739c1173897f4556b8aef76f65a91ef0cd9c

PR #501, opened 2019-06-22
refactor: pass addresses to ui with independent method

Keith-CY:
https://github.com/nervosnetwork/neuron/compare/develop...Keith-CY:display-addresses?expand=1#diff-2873cc75a8b8adbfebbb7b3dc4d3e498R298

A method is required to fetch a bundle of unused addresses. For now only the next one returns. @classicalliu

### DIFF
--- a/packages/neuron-ui/src/components/Addresses/index.tsx
+++ b/packages/neuron-ui/src/components/Addresses/index.tsx
@@ -28,6 +28,10 @@ const headers = [
     key: 'address',
   },
   {
+    label: 'addresses.identifier',
+    key: 'identifier',
+  },
+  {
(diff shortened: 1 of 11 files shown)


classicalliu:
Do you mean this https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/addresses.ts#L149 ?

Keith-CY:
> Do you mean this https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/addresses.ts#L149 ?

I'm not sure if this api returns `m` receiving addresses and `n` change addresses, it seems return all addresses.

PR #502, opened 2019-06-24
feat: add description to address

classicalliu:


packages/neuron-wallet/src/addresses/entities/address.ts
```diff
@@ -49,6 +49,12 @@ export default class Address extends BaseEntity {
   })
   version!: AddressVersion
 
+  @Column({
+    type: 'text',
```
classicalliu: Does this description will be long ? Should I use `text` or `varchar` ? @ashchan

ashchan: Limit to a max length is reasonable.

classicalliu: `varchar` is limited to 255 chars and text is much longer, so should we change to varchar to have a limit ?

ashchan: I believe `varchar` is enough. On UI or app layer, we can limit that to 100 UTF8 characters.

classicalliu: updated to `varchar`

PR #503, opened 2019-06-24
refactor(neuron-ui): handle decimal in ckb formatter

Keith-CY:


packages/neuron-ui/src/utils/formatters.ts
```diff
@@ -50,15 +50,19 @@ export const currencyFormatter = (
...
 
 export const CKBToShannonFormatter = (amount: string, uint: CapacityUnit) => {
+  const [integer = '0', decimal = ''] = amount.split('.')
+  const decimalLength = 10 ** decimal.length
+  const num = integer + decimal
+
   switch (uint) {
     case CapacityUnit.CKB: {
-      return (BigInt(amount) * BigInt(1e8)).toString()
+      return (BigInt(num) * BigInt(1e8 / decimalLength)).toString()
```
classicalliu: Is `1e8 / decimalLength` a `number` and then convert to `bigint` ?

Keith-CY: > Is `1e8 / decimalLength` a `number` and then convert to `bigint` ?

It's a `number` like `10`, `100` and throws exceptions when`1e8/decimalLength` is not an `integer`

classicalliu: ok

PR #508, opened 2019-06-24
feat(neuron-wallet): subscribe to used addresses and send address lis…

Keith-CY:
…t to neuron-ui

### DIFF
--- a/packages/neuron-wallet/src/services/addresses.ts
+++ b/packages/neuron-wallet/src/services/addresses.ts
@@ -7,7 +7,6 @@ import env from '../env'
 import { AddressVersion } from '../addresses/entities/address'
 
 const MAX_ADDRESS_COUNT = 30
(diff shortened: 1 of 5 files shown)


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -164,3 +167,22 @@ export default class AddressService {
...
+AddressesUsedSubject.getSubject().subscribe(async () => {
+  const currentWallet = WalletService.getInstance().getCurrent()
+  if (currentWallet) {
+    const addresses = await AddressService.allAddressesByWalletId(currentWallet.id).then(addrs =>
+      addrs.map(({ address, blake160: identifier, addressType: type, txCount }) => ({
+        address,
+        identifier,
+        type,
+        txCount,
+        description: 'mock description',
```
classicalliu: Can use real description from database now.

PR #512, opened 2019-06-24
fix: tx count by lock hash should include lock hash in input

classicalliu:


classicalliu:
> A question: does this handle the situation under which a single address appears in both inputs and outputs of one transaction?
> 
> For now this might not be a big problem, but in long run we have to consider this. And of course we can have an open discussion about this.
> 
> (For example, our mining and demo examples are using one address for both sending out address and change address; that makes the address appears in inputs/outputs of a single tx.

It returns tx count by unique `txHash`, so if `lockHash` appears in both inputs and outputs, it will get count `1` because of same `hash` of transaction it belongs to.

PR #516, opened 2019-06-25
feat(neuron-ui): add description field and clear button in the send view

Keith-CY:


packages/neuron-ui/src/services/UILayer.ts
```diff
@@ -158,6 +158,7 @@ export const walletsCall = instantiateMethodCall(wallets) as {
       capacity: string
     }[]
     fee: string
+    description: string
```
classicalliu: `description` will be `undefined` if not fill in.

PR #524, opened 2019-06-26
Create models and database folders

ashchan:
For better organization.


packages/neuron-wallet/src/services/sync/check-and-save/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Block } from '../../../app-types/types'
+import { Block } from '../../../types/types'
```
Keith-CY: `types/types` is confusing.

ashchan: The hardest thing in the world.

packages/neuron-wallet/tests/utils/lock-utils.test.ts
```diff
@@ -1,4 +1,4 @@
-import { Script } from '../../src/app-types/types'
+import { Script } from '../../src/types/types'
```
Keith-CY: another `types/types`

PR #525, opened 2019-06-26
refactor(neuron-wallet): send message when addresses update

Keith-CY:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -170,12 +170,14 @@ export default class WalletService {
...
       })
 
-    AddressesUsedSubject.getSubject().subscribe(async () => {
-      const currentWallet = this.getCurrent()
-      if (currentWallet) {
-        broadcastCurrentAddresses(currentWallet.id)
-      }
-    })
+    AddressDbChangedSubject.getSubject()
+      .pipe(debounceTime(DEBOUNCE_TIME))
```
classicalliu: I think this value can be larger than `1000`

Keith-CY: Will the interval of events be larger than 1000ms?

Keith-CY: Will the interval of events be larger than 1000ms?

PR #528, opened 2019-06-26
test(neuron-wallet): add delay in tests of updating network

Keith-CY:


Keith-CY:
> Delaying looks magic if that does solve the random network test failures. Please investigate if that failure doesn't happen again - as if so there must be some issue in the network management flow.

It does look awkward but has passed several times without failure...

PR #538, opened 2019-06-27
fix: fix tx list empty problem

classicalliu:


packages/neuron-wallet/src/controllers/transactions.ts
```diff
@@ -34,9 +34,12 @@ export default class TransactionsController {
...
     const { pageNo, pageSize, addresses = '' } = params
 
-    let searchAddresses = addresses.split(',')
+    let searchAddresses = addresses
+      .split(',')
+      .map(addr => addr.trim())
+      .filter(addr => addr !== '')
 
-    if (!searchAddresses.length) {
+    if (searchAddresses.length <= 0) {
```
ashchan: When will `length` be less than 0?

classicalliu: Equals to `=== 0`

ashchan: <= 0 and === 0 are different.

classicalliu: rollback to `!searchAddresses.length`

PR #539, opened 2019-06-27
refactor(neuron-ui): display type string of transactions

Keith-CY:


packages/neuron-ui/src/locales/zh.json
```diff
@@ -64,8 +64,8 @@
       "date": "时间",
       "amount": "价值(ckb)",
       "transaction-hash": "交易哈希",
-      "sent": "发出",
-      "received": "收入",
+      "send": "发出",
```
ashchan:
```suggestion
      "send": "转账",
```

packages/neuron-ui/src/locales/zh.json
```diff
@@ -64,8 +64,8 @@
...
+      "receive": "收入",
```
ashchan:
```suggestion
      "receive": "收款",
```

packages/neuron-ui/src/contexts/NeuronWallet/chain.ts
```diff
@@ -1,8 +1,8 @@
 import { RawNetwork } from 'components/NetworkEditor'
-import { ConnectStatus, TransactionType } from 'utils/const'
+import { ConnectStatus } from 'utils/const'
 
 export interface Transaction {
-  type: TransactionType
+  type: 'send' | 'receive' | 'other'
```
ashchan: What kind of tx belongs to `other`?

PR #547, opened 2019-06-28
feat: add status to tx list

classicalliu:


packages/neuron-wallet/src/types/cell-types.ts
```diff
@@ -13,6 +13,11 @@ export interface BlockHeader {
   number: string
 }
 
+export enum TransactionStatus {
```
ashchan: Consider using `pending` (or `sending` or `receiving`), `success` and `failed`.

classicalliu: ok

classicalliu: updated

PR #559, opened 2019-06-28
fix: add value to get one transaction

classicalliu:


packages/neuron-wallet/src/controllers/transactions.ts
```diff
@@ -60,6 +61,31 @@ export default class TransactionsController {
...
+        return LockUtils.addressToLockHash(addr)
+      })
+    )
+
+    const outputCapacities: bigint = transaction
+      .outputs!.filter(o => lockHashes.includes(o.lockHash!))
+      .map(o => BigInt(o.capacity))
+      .reduce((result, c) => result + c, BigInt(0))
+    const inputCapacities: bigint = transaction
+      .inputs!.filter(i => {
```
Keith-CY: It seems that the `filter` is unnecessary.

PR #562, opened 2019-07-01
refactor(neuron-wallet): bind the main window to the globalThis object.

Keith-CY:


packages/neuron-wallet/src/startup/create-window.ts
```diff
@@ -1,6 +1,7 @@
 import { BrowserWindow } from 'electron'
 import windowStateKeeper from 'electron-window-state'
 import path from 'path'
+import WindowManager from '../models/window-manager'
```
ashchan: We might need to change `create-window`'s name to reflect the fact that it only handles the main window creation.

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/562/commits/6e86c3d5a613318a7ce168b458b11cf015ccf778

PR #563, opened 2019-07-01
feat: using previous database if no network

classicalliu:
Save latest chain info (previous `genesisBlockHash` and `systemScriptInfo`) to file and load if no network no response when start or switch.


packages/neuron-wallet/src/startup/sync-block-task/init-database.ts
```diff
@@ -1,14 +1,24 @@
...
     await initConnection(hash)
+    const systemScriptInfo = await LockUtils.systemScript()
+    updateMetaInfo({ genesisBlockHash: hash, systemScriptInfo })
   } catch (err) {
-    Utils.sleep(1000)
-    await initDatabase()
+    try {
+      const metaInfo = getMetaInfo()
+      await initConnection(metaInfo.genesisBlockHash)
+      LockUtils.setSystemScript(metaInfo.systemScriptInfo)
```
ashchan: Since `LockUtils` is not a pure utils class any more (it stores its own state), please move it to models or other appropriate module.

classicalliu: Move to `models` and still named `LockUtils`

PR #576, opened 2019-07-03
refactor(refactor networks service): use subject to handle events of networks service

Keith-CY:
1. use network subjects to handle events from networks service.
2. use `current` instead of `active` for consistency with the wallets module.


classicalliu:
CI failed

PR #579, opened 2019-07-04
feat: update tx description

classicalliu:


Keith-CY:
@ashchan how to show the alert of the failure to update transaction descriptions?

classicalliu:
`description` will not show on UI layer (can update, but still show empty) @Keith-CY

Keith-CY:
> `description` will not show on UI layer (can update, but still show empty) @Keith-CY

It is displayed in the transaction list.

classicalliu:
> > `description` will not show on UI layer (can update, but still show empty) @Keith-CY
> 
> It is displayed in the transaction list.

I means `description` in transaction list is still empty, not read the real data.
You can try after this merged.

classicalliu:
I found the problem, I'll fix @Keith-CY

ashchan:
> @ashchan how to show the alert of the failure to update transaction descriptions?

This should not happen in normal use. When does this happen?

PR #586, opened 2019-07-05
Organize scripts

ashchan:


scripts/copy-ui-files.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+./node_modules/.bin/ncp ./packages/neuron-ui/build ./packages/neuron-wallet/dist/neuron-ui
```
classicalliu: Can we use link instead of copy ?

ashchan: I don’t think packaging process can handle link gracefully.

classicalliu: OK, it seems only for package and e2e.

PR #602, opened 2019-07-08
fix: fix tx db changed subject and add tx in event

classicalliu:


packages/neuron-wallet/src/models/subjects/tx-db-changed-subject.ts
```diff
@@ -1,14 +1,20 @@
...
+
+export interface EventAndTx {
+  event: string
+  tx: Transaction
+}
 
 // subscribe this Subject to monitor any transaction table changes
 export class TxDbChangedSubject {
-  static subject = new ReplaySubject<string>(100)
+  static subject = new ReplaySubject<EventAndTx>(100)
```
Keith-CY: The type name is weird

Keith-CY: How about `TransactionMessage`

classicalliu: any advice

classicalliu: Renamed to `TransactionChangedMessage`

PR #603, opened 2019-07-08
refactor: Set text field font size properly

ashchan:


ashchan:
Closing in favor of #604.

PR #608, opened 2019-07-09
fix: showing log in console when run `yarn start`

classicalliu:


classicalliu:
Using `cd` instead of `lerna exec` in `start:ui` and `start:wallet` scripts now.

PR #632, opened 2019-07-11
fix(neuron-ui): use the first unused receiving address for receiving

Keith-CY:


packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -31,7 +31,8 @@ const Receive = ({
   const [showLargeQRCode, setShowLargeQRCode] = useState(false)
 
   const accountAddress = useMemo(
-    () => params.address || (addresses.find(addr => addr.type === 0) || { address: '' }).address || '',
+    () =>
+      params.address || (addresses.find(addr => addr.type === 0 && addr.txCount > 0) || { address: '' }).address || '',
```
ashchan:
```suggestion
      params.address || (addresses.find(addr => addr.type === 0 && addr.txCount === 0) || { address: '' }).address || '',
```

We're looking for unused address.

PR #656, opened 2019-07-15
test(neuron-wallet): add clean after each test case

Keith-CY:


ashchan:
Still failing

Keith-CY:
image

As the log says, this kind of errors comes from `networks.test.ts` [line: 31](https://github.com/nervosnetwork/neuron/blob/c1503cdf7d9f9991ea27802ea7481cd6f20de3db/packages/neuron-wallet/tests/services/networks.test.ts#L31)

Which is
```
service = new NetworksService()
```

and the first error is thrown since the second test case.

So in my opinion, the `beforeEach` and `afterEach` have intervened each other. I'd like to add a time interval between them for trying to solve this problem.

Keith-CY:
I've removed initialization in `beforeEach` cuz `clear` also reset the test environment.

PR #686, opened 2019-07-18
chore: return if tx success when fetch

classicalliu:


packages/neuron-wallet/src/services/transactions.ts
```diff
@@ -324,6 +324,11 @@ export default class TransactionsService {
       .getRepository(TransactionEntity)
       .findOne(transaction.hash, { relations: ['inputs', 'outputs'] })
 
+    // return if success
+    if (txEntity && txEntity.status === TransactionStatus.Success) {
```
ashchan: Does this mean the confirmed tx has already been fetched and saved before so there's no need to persist again?

classicalliu: Yes, if process again, the output's status will be update to `live`, but it may be used before, so it might be `pending` or `dead`, if update, the `balance` and `live cells` would be wrong.

ashchan: I see.

PR #706, opened 2019-07-19
feat(neuron-ui): format difficulty in local number string

Keith-CY:


Keith-CY:
Included in https://github.com/nervosnetwork/neuron/pull/707

PR #709, opened 2019-07-19
feat(neuron-ui): enable copying the pubkey hash of the default address

Keith-CY:


packages/neuron-ui/src/components/Overview/index.tsx
```diff
@@ -311,10 +328,11 @@ const Overview = ({
           </Stack>
         </Callout>
       ) : null}
-      {/* TODO: Implement this */}
       {minerInfoRef.current ? (
         <Callout target={minerInfoRef.current} hidden={!displayMinerInfo} onDismiss={hideMinerInfo} gapSpace={0}>
-          {t('overview.miner-info')}
+          <Stack tokens={{ padding: 15 }}>
```
ashchan: Besides the copy pub key hash button, let's also display these info above the button:

* Address
* Code_hash
* Lock_arg (args's first item) (a.k.a the pubkey hash)

PR #717, opened 2019-07-22
feat: init the app from neuron-ui

Keith-CY:


ashchan:
@Keith-CY My conflict merge from `develop` seems to have broken it, could you take a look? Thanks.

PR #721, opened 2019-07-23
chore(dev-deps): Batch update dev deps

ashchan:


packages/neuron-ui/src/components/Settings/index.tsx
```diff
@@ -59,7 +59,9 @@ const Settings = ({
           </IconButton>
         </Stack.Item>
         <Stack.Item align="center">
-          <Text variant="xLarge" as="h1">{t('navbar.settings')}</Text>
+          <Text variant="xLarge" as="h1">
```
ashchan: Change caused by local lint run.

PR #727, opened 2019-07-23
fix: wrapper async.queue and rewrite kill

classicalliu:


packages/neuron-wallet/src/services/queue-adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export default class QueueAdapter {
+  private q: any
+
+  constructor(worker: async.AsyncWorker<any, Error>, concurrency?: number | undefined) {
+    this.q = async.queue(worker, concurrency)
+  }
+
+  public push = (value: any) => {
+    this.innerPush(value)
```
ashchan: Does `inner` mean internal implementation? I feel calling `this.q.push(value)` is direct and enough assuming `kill` still works as the `innerPush`.

classicalliu: The original `q.kill()` seems strange, it still can push value to queue, and I wrapper this method and reset push method to an empty method, the `innerPush` is for easy to recover it (`push`). But seems it's no use now.

PR #730, opened 2019-07-24
feat(neuron-ui): use go back instead of go to overview in the setting…

Keith-CY:
…s view


Keith-CY:
> I don't think going back would work in some cases.
> 
> For example, go to Preferences, go to Network, then add a new network.
> 
> After adding the new network, if we click the back button the app would go back to the add network view again, but not exit the settings.

check current wallet id on leaving the settings view https://github.com/nervosnetwork/neuron/pull/730/commits/b33a238f45f8072c63d26503a8f8b3230fef946f

PR #732, opened 2019-07-24
feat(neuron-ui): only support en and zh languages

Keith-CY:


packages/neuron-ui/src/utils/i18n.ts
```diff
@@ -7,7 +7,15 @@ import en from 'locales/en.json'
 i18n.use(initReactI18next).init({
   resources: {
     en,
+    'en-AU': en,
```
ashchan: Do we really need to map all possible `zh-` and `en-` locales?

A simpler solution: if system language is `zh` or `zh-CN`, select `zh`, otherwise select `en` (even for `zh-TW`).

Keith-CY: redundant resources have been removed https://github.com/nervosnetwork/neuron/pull/732/commits/d1212feb4d608b5d32d769f76cde41379b322e1b,

PR #733, opened 2019-07-24
use remote module instead of channels

Keith-CY:
Removed `Channel Messages`, and use the `remote` module for calling methods in neuron-wallet


Keith-CY:
@classicalliu Plz check the modification on `sync-block`

classicalliu:
> @classicalliu Plz check the modification on `sync-block`

I haven't see any changes about `sync-block`

PR #755, opened 2019-07-27
fix: move update address to sync process and buffer it

classicalliu:
update address will block main process now

### DIFF
--- a/packages/neuron-wallet/src/listeners/address.ts
+++ b/packages/neuron-wallet/src/listeners/address.ts
@@ -1,12 +1,28 @@
+import { remote } from 'electron'
+import { ReplaySubject } from 'rxjs'
+import { bufferTime } from 'rxjs/operators'
 import AddressesUsedSubject from '../models/subjects/addresses-used-subject'
 import AddressService from '../services/addresses'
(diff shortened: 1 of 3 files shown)


packages/neuron-wallet/src/listeners/address.ts
```diff
@@ -1,12 +1,28 @@
...
+addressesUsedSubject.subscribe((addresses: string[]) => {
+  bridge.next(addresses)
+})
+
 // update txCount when addresses used
 export const register = () => {
-  AddressesUsedSubject.getSubject().subscribe(async (addresses: string[]) => {
-    const addrs = await AddressService.updateTxCountAndBalances(addresses)
+  bridge.pipe(bufferTime(1000)).subscribe(async (addressesList: string[][]) => {
+    const addresses = addressesList.reduce((acc, val) => acc.concat(val), [])
```
Keith-CY:
```suggestion
    const addresses = addressesList.flat()
```

classicalliu: `flat()` should enable module in tsconfig, and it may cause new problems

PR #760, opened 2019-07-27
feat(neuron-wallet): remove spend-from menuitem

Keith-CY:


packages/neuron-wallet/src/controllers/app/options.ts
```diff
@@ -157,12 +157,15 @@ export const contextMenuTemplate: {
           AppController.navTo(`/receive/${address}`)
         },
       },
-      {
-        label: i18n.t('contextMenu.spend-from'),
-        click: () => {
-          AppController.navTo(`/send/${address}`)
-        },
-      },
+      /**
```
ashchan: Please just remove it. (No comment left)

Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/760/commits/6805fa2300610a88e16379ac46bd86fb5eb7e3bd

PR #770, opened 2019-07-30
feat: Check for updates

ashchan:
Add menu item allowing user to check updates.

### DIFF
--- a/packages/neuron-wallet/src/controllers/app/index.ts
+++ b/packages/neuron-wallet/src/controllers/app/index.ts
@@ -1,5 +1,5 @@
 import path from 'path'
-import { dialog, shell, Menu, MessageBoxOptions, SaveDialogOptions, BrowserWindow } from 'electron'
+import { dialog, shell, Menu, MenuItem, MessageBoxOptions, SaveDialogOptions, BrowserWindow } from 'electron'
 import { take } from 'rxjs/operators'
 import app from '../../app'
 import { URL, contextMenuTemplate } from './options'
@@ -9,6 +9,7 @@ import NetworksService from '../../services/networks'
 import WalletsService from '../../services/wallets'
 import WalletsController from '../wallets'
 import SyncInfoController from '../sync-info'
+import UpdateController from '../update'
 
 import { Controller as ControllerDecorator } from '../../decorators'
 import { Channel, ResponseCode } from '../../utils/const'
@@ -153,6 +154,10 @@ export default class AppController {
       cancelId: 0,
     }
     AppController.showMessageBox(options)
+  }
+
+  public static checkUpdates(menuItem: MenuItem) {
+    UpdateController.checkUpdates(menuItem)
   }
 
   public static openWebsite() {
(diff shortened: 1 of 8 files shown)


Keith-CY:
CI failed

ashchan:
> CI failed

Checking that

packages/neuron-wallet/src/controllers/app/index.ts
```diff
@@ -83,6 +87,9 @@ export default class AppController {
       connectionStatus,
       codeHash,
     }
+
+    AppController.handleWalletListChange()
```
Keith-CY: `getInitState` would be called multiple times, which makes the subscription repeated.

packages/neuron-wallet/src/controllers/app/index.ts
```diff
@@ -155,6 +162,10 @@ export default class AppController {
     AppController.showMessageBox(options)
   }
 
+  public static checkUpdates(menuItem: MenuItem) {
+    UpdateController.checkUpdates(menuItem)
```
Keith-CY: The name `update controller checks updates menu item` is confusing.

ashchan: Indeed. What's your suggestion? We do need to pass the menu item for disabling/enabling.

packages/neuron-wallet/src/controllers/app/index.ts
ashchan: Where's the appropriate place to initialize this?

packages/neuron-wallet/src/controllers/app/index.ts
ashchan: `checkUpdates(sender: MenuItem)`, `checkUpdates(sender: any)`?

Keith-CY: If [Menu.getApplicationMenu](https://electronjs.org/docs/api/menu#menugetapplicationmenu) works in the update controller, the `menuitem` parameter is redundant.

ashchan: We might trigger this by other UI component, eg from a button. `sender: any` is a common signature when programming macOS/iOS app.

ashchan: > If [Menu.getApplicationMenu](https://electronjs.org/docs/api/menu#menugetapplicationmenu) works in the update controller, the `menuitem` parameter is redundant.

Not exactly. We can't be sure the menu item triggering it is from the main menu. (But if we have multiple entries then we do need to disable/enable all).

I'd say we leave this as. If we add more action entries then we refactor/rename this.

Keith-CY: LGTM

packages/neuron-wallet/src/controllers/app/index.ts
Keith-CY: We can subscribe to the subjects just on the app launched.

ashchan: I see. Thanks.

ashchan: @Keith-CY Moved that to `app.ready` event block. Please let me know if there's a better launch entry.

PR #775, opened 2019-07-30
[ᚬmaster] Rc/v0.17.0 alpha.1

ashchan:


Keith-CY:
`lerna.json` is not updated.

ashchan:
> `lerna.json` is not updated.

Nice catch!

PR #784, opened 2019-07-31
feat: Only bundle en and zh_CN language folders for mac

ashchan:


packages/neuron-wallet/electron-builder.yml
```diff
@@ -58,6 +58,7 @@ mac:
   gatekeeperAssess: false
   entitlements: assets/entitlements.plist
   entitlementsInherit: assets/entitlements.plist
+  electronLanguages: ["en", "zh", "zh_CN"]
```
Keith-CY: What's the difference between `zh` and `zh-CN`

Keith-CY: I've found that.

PR #789, opened 2019-07-31
feat(neuron-ui): add notification panel

Keith-CY:
show all warnings and alerts in the notification panel

### DIFF
--- a/packages/neuron-ui/src/components/NetworkEditor/hooks.ts
+++ b/packages/neuron-ui/src/components/NetworkEditor/hooks.ts
@@ -1,7 +1,7 @@
 import { useState, useEffect, useMemo, useCallback } from 'react'
 
-import { AppActions, StateDispatch } from 'states/stateProvider/reducer'
-import { createNetwork, updateNetwork } from 'states/stateProvider/actionCreators'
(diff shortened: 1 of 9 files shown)


packages/neuron-ui/src/components/NetworkEditor/hooks.ts
```diff
@@ -126,72 +120,54 @@ export const useHandleSubmit = (
...
-      return dispatch({
-        type: AppActions.AddNotification,
-        payload: {
-          ...warning,
-          content: i18n.t(Message.LengthOfNameShouldBeLessThanOrEqualTo, {
-            length: MAX_NETWORK_NAME_LENGTH,
-          }),
-        },
-      })
+      return addNotification({
```
ashchan: When validating a form, if an filed has error, do we show any error style for that field? For example, if name's length exceeds limit, beside the error message (notification), do we style the name field so it stands out (e.g., its border is rendering as red)?

If not, and if some form errors hide in the pane, the user would be confused why we don't let them submit.

Keith-CY: The error messages will be appended to the field in the future

PR #790, opened 2019-07-31
refactor: clean and refactor queue

classicalliu:


packages/neuron-wallet/src/services/sync/block-listener.ts
```diff
@@ -44,46 +45,37 @@ export default class BlockListener {
     if (restart) {
       await this.currentBlockNumber.updateCurrent(BigInt(0))
     }
-    while (this.queue !== null) {
+    while (!this.stopFlag) {
```
ashchan: We would rename `stopFlag` to `stopped` or `isStopped`.

classicalliu: updated

PR #796, opened 2019-08-01
fix(neuron-ui): remove the quotation mark around error message

Keith-CY:


packages/neuron-ui/src/states/stateProvider/actionCreators/wallets.ts
```diff
@@ -113,7 +113,8 @@ export const sendTransaction = (params: Controller.SendTransaction) => (dispatch
       if (res.status) {
         history.push(Routes.History)
       } else {
-        addNotification({ type: 'alert', content: JSON.stringify(res.message.title) })(dispatch)
+        // TODO: the pre-treat is unnecessary once the error code is implemented
```
ashchan: What is the error code change is expected to finish? If that's very soon we can combine these two changes together.

Keith-CY: The CKB team is going to add error codes in the document, but the deadline is not determined.

ashchan: I see.

PR #817, opened 2019-08-04
Baseurl setup for neuron-wallet

ashchan:
So that we can use absolute import path.


ormconfig-address.json
```diff
@@ -5,13 +5,13 @@
   "logging": true,
   "database": "address-dev.sqlite",
   "entities": [
-     "packages/neuron-wallet/src/database/address/entities/**/*.ts"
+     "packages/neuron-wallet/dist/database/address/entities/**/*.js"
```
ashchan: @classicalliu Could you take a look and test if with these changes our db cli tasks still work? Thanks!

classicalliu: It works, but in my mind, we should run `yarn build` before `db:chain` and `db:address`

packages/neuron-wallet/package.json
```diff
@@ -64,6 +64,8 @@
     "lint-staged": "9.2.0",
     "neuron-ui": "0.17.0-alpha.5",
     "rimraf": "2.6.3",
-    "spectron": "7.0.0"
+    "spectron": "7.0.0",
+    "ts-transformer-imports": "^0.4.3",
```
Keith-CY: Do not use version range

ashchan: Which rule claims that? Without any param like `exact` lerna add creates this by default. So this is really not a human mistake.

ashchan: I'll merge and confirm this. If we do have an explicit rule for this we'll have another PR. There're other dependencies with `^` it worthy a separate change.

Keith-CY: I've encountered a package who broken the semver rules and made everything down.

ashchan: Will send another PR updating that. But I don't think this rule should be required.

PR #820, opened 2019-08-05
refactor(neuron-ui): add missing dependency cachedNetwork in network …

Keith-CY:
…editor


ashchan:
Closed as duplicated with #819.

PR #846, opened 2019-08-09
chore: set yield time to reduce CPU

classicalliu:


packages/neuron-wallet/src/services/sync/queue.ts
```diff
@@ -48,19 +50,26 @@ export default class Queue {
...
       try {
         this.inProcess = true
+
         const current: bigint = await this.currentBlockNumber.getCurrent()
         const startNumber: bigint = current + BigInt(1)
         const endNumber: bigint = current + BigInt(this.fetchSize)
         const realEndNumber: bigint = endNumber < this.endBlockNumber ? endNumber : this.endBlockNumber
 
+        if (realEndNumber >= this.endBlockNumber) {
+          this.yieldTime = 1000
```
ashchan: Is this `1000ms`?

classicalliu: Yes

PR #850, opened 2019-08-10
fix(neuron-ui): handle overflow on recent activities

Keith-CY:
image

The scrollbar is hidden by chromium automatically, try to find a better indicator in the future.

UPDATED: add a custom scrollbar on the list
image


Keith-CY:
image
image
I think the layout does the right thing now. @ashchan

PR #852, opened 2019-08-11
chore: Add wallet e2e test

cezres:
Test create/import/switch/delete/edit wallet


packages/neuron-wallet/tests-e2e/wallet.test.ts
```diff
@@ -0,0 +1,224 @@
...
+      return app.stop()
+    }
+  })
+
+  it('create wallet', async () => {
+    const { client } = app
+    await client.waitUntilWindowLoaded()
+
+    // Click create wallet
+    const createWalletButton = await getElementByTagName(client, 'button', '创建新钱包')
```
ashchan: Do we set language to Chinese explicitly or is this just an assumption which matches local development environment? If so we should use English as that's the default language (and as fallback language), and I believe our CI has that as default locale as well.

cezres: Ok, I have added the locale parameter `--lang=en`. https://github.com/nervosnetwork/neuron/pull/852/commits/914890aa7154755da4316f4d3516dc6c14a3efba

PR #858, opened 2019-08-12
fix(neuron-ui): replace delimiters in error messages

Keith-CY:
replace the '.' and ':' in error messages which make the i18n work improperly.

With `.` and `:` in the messages, the alert is not rendered correctly. 
image
image

The `i18n` recognizes the `127.0.0.1:8114` as 
```json
{
  "127": {
    "0": {
      "0": {
        "1": {
          "8114": undefined
        }
      }
    }
  }
}
```
After the replacement, the alert is rendered correctlly:
image

### DIFF
--- a/packages/neuron-ui/src/services/remote/controllerMethodWrapper.ts
+++ b/packages/neuron-ui/src/services/remote/controllerMethodWrapper.ts
@@ -66,7 +66,7 @@ export const controllerMethodWrapper = (controllerName: string) => (
   }
   return {
     status: 0,
-    message: { title },
+    message: { title: title.replace(/(\.|:)/g, '_') },
   }
 }
 


Keith-CY:
> May I ask why `127.0.0.1:8114` is a nested JSON object in the first place?

Actually the most outside key is `connect ECONNREFUSED 127`, and the `i18next` works in this way
image

Keith-CY:
Maybe change `keyseparator` and `nsseparator` is a better solution, I'll have a try.

Keith-CY:
> Maybe change `keyseparator` and `nsseparator` is a better solution, I'll have a try.

It requires a bulk of replacement on existing messages.

Keith-CY:
I18n should be used on pushing messages

PR #859, opened 2019-08-12
refactor(neuron-ui): tag a tx as sending or receiving if its status i…

Keith-CY:
…s pending

### DIFF
--- a/packages/neuron-ui/src/components/Overview/index.tsx
+++ b/packages/neuron-ui/src/components/Overview/index.tsx
@@ -34,16 +34,26 @@ import { backToTop } from 'utils/animations'
 const TITLE_FONT_SIZE = 'xxLarge'
 export type ActivityItem = State.Transaction & { confirmations: string; typeLabel: string }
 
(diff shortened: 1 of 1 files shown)


packages/neuron-ui/src/components/Overview/index.tsx
```diff
@@ -34,16 +34,26 @@ import { backToTop } from 'utils/animations'
...
-const genTypeLabel = (type: 'send' | 'receive', confirmationCount: number) => {
+const genTypeLabel = (
+  type: 'send' | 'receive',
+  confirmationCount: number,
+  status: 'pending' | 'success' | 'failed'
+) => {
   switch (type) {
     case 'send': {
-      if (confirmationCount < CONFIRMATION_THRESHOLD) {
+      if (status === 'failed') {
```
ashchan: This seems to be the same as the fallback (last line of return).

Keith-CY: All tx are remarked as pending without this branch when the tip block number is less than or equal to the confirmations threshold.

PR #861, opened 2019-08-13
chore: don't show error when network off

classicalliu:


packages/neuron-wallet/src/services/sync/queue.ts
```diff
@@ -67,7 +67,11 @@ export default class Queue {
           await this.pipeline(rangeArr)
         }
       } catch (err) {
-        logger.error(`sync error:`, err)
+        if (err.message.startsWith('connect ECONNREFUSED')) {
+          logger.debug(`sync error:`, err)
+        } else {
+          logger.error(`sync error:`, err)
```
ashchan: Does this bubble up to exception handler and display on UI? If not I don't see any difference by changing the log level.

classicalliu: No, but debug level will not write to log file when packaged.

ashchan: I see. If the title had been 'don't log error' it wouldn't be misleading.

PR #863, opened 2019-08-13
feat: sync by indexer RPCs

classicalliu:
* impl sync by indexer RPCs
* auto use indexer mode when indexer RPCs is available


ashchan:
Holding this to prevent from merging while we review and test this feature.

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  public processFork = async () => {
+    while (!this.stopped) {
+      try {
+        const tip = this.tipBlockNumber
+        const txs = await IndexerTransaction.txHashes()
+        for (const tx of txs) {
+          const result = await this.getBlocksService.getTransaction(tx.hash)
+          if (!result) {
```
shaojunda: Under what conditions will tx not exist?

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -0,0 +1,221 @@
...
+      page += 1
+    }
+  }
+
+  public stop = () => {
+    this.tipNumberListener.unsubscribe()
+    this.stopped = true
+  }
+
+  public waitForDrained = async (timeout: number = 5000) => {
```
shaojunda: what does `waitForDrained` do?

packages/neuron-wallet/src/startup/sync-block-task/indexer.ts
```diff
@@ -0,0 +1,55 @@
...
+  const addresses: string[] = (await AddressService.allAddresses()).map(addr => addr.address)
+  const lockHashes: string[] = await LockUtils.addressesToAllLockHashes(addresses)
+  return lockHashes
+}
+
+// call this after network switched
+let indexerQueue: IndexerQueue | undefined
+export const switchNetwork = async (nodeURL: string) => {
+  // stop all blocks service
+  if (indexerQueue) {
```
shaojunda: Does the `indexerQueue` have a value when setup?

packages/neuron-wallet/src/services/sync/get-blocks.ts
```diff
@@ -55,6 +55,16 @@ export default class GetBlocks {
     return block
   }
 
+  public getTransaction = async (hash: string): Promise<CKBComponents.TransactionWithStatus> => {
```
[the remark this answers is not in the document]
classicalliu: No need, it's block info too.

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -0,0 +1,221 @@
...
+          this.resetFlag = false
+        }
+        const { lockHashes } = this
+        const currentBlockNumber: bigint = await this.blockNumberService.getCurrent()
+        if (!this.indexed || currentBlockNumber !== this.tipBlockNumber) {
+          if (!this.indexed) {
+            await this.indexLockHashes(lockHashes)
+            this.indexed = true
+          }
+          const minBlockNumber = await this.getCurrentBlockNumber(lockHashes)
```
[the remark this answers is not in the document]
classicalliu: `minBlockNumber` means the min block number in `get_lock_hash_index_states` of all lock hashes.

packages/neuron-wallet/src/services/indexer/queue.ts
classicalliu: The block of tx not in the longest chain now.

packages/neuron-wallet/src/services/indexer/queue.ts
classicalliu: Ensure this queue stopped after call `stop`, then you can run a new queue.

packages/neuron-wallet/src/startup/sync-block-task/indexer.ts
classicalliu: When you call switch network multi times.

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -0,0 +1,221 @@
...
+        let txPoint: CKBComponents.TransactionPoint | null = null
+        if (type === TxPointType.CreatedBy) {
+          txPoint = tx.createdBy
+        } else if (type === TxPointType.ConsumedBy) {
+          txPoint = tx.consumedBy
+        }
+
+        if (
+          txPoint &&
+          (BigInt(txPoint.blockNumber) >= startBlockNumber || this.tipBlockNumber - BigInt(txPoint.blockNumber) < 1000)
```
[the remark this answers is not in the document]
classicalliu: No need, only use once time.

packages/neuron-wallet/src/services/sync/utils.ts
```diff
@@ -48,4 +48,20 @@ export default class Utils {
     }
     return result
   }
+
+  public static min = (array: bigint[]): bigint | undefined => {
```
ashchan: Why would a numeric operation belong to sync/utils?

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -0,0 +1,221 @@
...
+import IndexerTransaction from 'services/tx/indexer-transaction'
+
+import IndexerRPC from './indexer-rpc'
+
+enum TxPointType {
+  CreatedBy = 'createdBy',
+  ConsumedBy = 'consumedBy',
+}
+
+export default class Queue {
```
ashchan: While I understand we have different folders - to some extend they act like namespaces, having multiple `Queue` classes worries me that it could confuse reading and understanding our design and implementation.

packages/neuron-wallet/src/services/sync/utils.ts
classicalliu: Utility methods of sync module.

packages/neuron-wallet/src/services/indexer/queue.ts
classicalliu: How about `IndexerQueue`

ashchan: That's better. We have two queue implementations for two different modules, then we name them differently for clarification. Please do so.

packages/neuron-wallet/src/startup/sync-block-task/task.ts
```diff
@@ -1,86 +1,45 @@
...
-
-  blockListener.start()
 }
 
 export const run = async () => {
   await initAddressConnection()
   databaseInitSubject.subscribe(async (network: NetworkWithID | undefined) => {
     if (network) {
-      await switchNetwork()
+      const indexerEnable = await testIndexer(network.remote)
```
ashchan:
```suggestion
      const indexerEnabled = await testIndexer(network.remote)
```

packages/neuron-wallet/src/startup/sync-block-task/task.ts
```diff
@@ -1,86 +1,45 @@
...
+      if (indexerEnable) {
```
ashchan:
```suggestion
      if (indexerEnabled) {
```

packages/neuron-wallet/src/startup/sync-block-task/task.ts
```diff
@@ -1,86 +1,45 @@
...
-  return lockHashes
-}
-
-// call this after network switched
-let blockListener: BlockListener | undefined
-export const switchNetwork = async () => {
-  // stop all blocks service
-  if (blockListener) {
-    await blockListener.stopAndWait()
+export const testIndexer = async (url: string): Promise<boolean> => {
```
ashchan: It'd be more clearer to name it such as `isIndexerEnabled` or something like that.

classicalliu: ok

PR #864, opened 2019-08-13
feat(neuron-ui): check the JSON format before keystore validation

Keith-CY:
Check the JSON format before the Keystore validation
image

### DIFF
--- a/packages/neuron-wallet/src/exceptions/common.ts
+++ b/packages/neuron-wallet/src/exceptions/common.ts
@@ -40,6 +40,12 @@ export class ShouldBeTypeOf extends Error {
   }
 }
 
+export class InvalidJSON extends Error {
+  constructor() {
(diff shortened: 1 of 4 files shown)


packages/neuron-wallet/src/exceptions/common.ts
```diff
@@ -40,11 +40,18 @@ export class ShouldBeTypeOf extends Error {
   }
 }
 
+export class InvalidJSON extends Error {
+  constructor() {
+    super(i18n.t('messages.invalid-JSON'))
```
ashchan: We can use `small-case` key even if `JSON` is a proper noun.

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/864/commits/227448f041be587a3df49849fd77e0833ee4b9c7

PR #868, opened 2019-08-15
feat(neuron-ui): update the cycles on the outputs changing

Keith-CY:
@classicalliu please implement the `calculateCycles` method in the wallet controller


ashchan:
Holding this til we have a clear solution to this feature.

ashchan:
Please un-hold this when we have the cycle related calculation (just works is enough, no need to be perfect) implemented.

Keith-CY:
TODO: format methods require more tests

Keith-CY:
@ashchan @classicalliu Ready for review

PR #869, opened 2019-08-16
feat: Add report issue menu item

ashchan:
Also move external URLs out to put them to `utils/const`.


ashchan:
Integration (e2e) tests are failing and under investigation. Don't think that's related to these changes.

PR #870, opened 2019-08-16
feat(neuron-ui): display the absolute value of a transaction as its a…

Keith-CY:
…mount


Keith-CY:
> Overview activities show absolute amount, but tx history list should show original amount (negative when sending out).

updated in https://github.com/nervosnetwork/neuron/commit/57c13c9f24afb000f891d1ac5969c4ece797627c

PR #872, opened 2019-08-16
feat: compute cycles

classicalliu:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -371,10 +372,31 @@ export default class WalletService {
...
+    }
+
+    const addressInfos = await this.getAddressInfos(walletID)
+
+    const addresses: string[] = addressInfos.map(info => info.address)
+
+    const lockHashes: string[] = await LockUtils.addressesToAllLockHashes(addresses)
+
+    const { inputs } = await CellsService.gatherInputs(capacities, lockHashes, '0')
+    const cycles = BigInt(1387008) * BigInt(inputs.length)
```
ashchan: Have we test different txs to see if the magic number fit? For example, more inputs/outputs, different amount (not necessarily to be related to the final cycles, just to cover more cases).

If this really works, we should still assign it to a variable, and comment for the rationale and how we get this exact number.

PR #879, opened 2019-08-17
feat: check tx success regular intervals

classicalliu:


ashchan:
What do `regular intervals` in the commit message mean?

classicalliu:
> What do `regular intervals` in the commit message mean?

every 5 seconds.

ashchan:
> > What do `regular intervals` in the commit message mean?
> 
> every 5 seconds.

I meant the commit message mentioned intervals but I didn't see anything related to the interval from the changes.

PR #880, opened 2019-08-17
chore(e2e): Add transaction e2e test

cezres:


packages/neuron-wallet/tests-e2e/operations/sendTransaction.ts
```diff
@@ -0,0 +1,105 @@
...
+  expect(passwordInputElement.value).not.toBeNull()
+  await app.setElementValue('//BODY/DIV[4]/DIV/DIV/DIV/DIV[2]/DIV[2]/DIV/DIV[1]//INPUT', password)
+  // Confirm
+  const confirmElement = await app.element('//BODY/DIV[4]/DIV/DIV/DIV/DIV[2]/DIV[2]/DIV/DIV[2]/BUTTON[2]')
+  expect(confirmElement.value).not.toBeNull()
+  await client.elementIdClick(confirmElement.value.ELEMENT)
+  app.waitUntilLoaded()
+
+  // Transaction element
+  const transactionElementPath = `//MAIN/DIV/DIV[2]/DIV/DIV/DIV[2]/DIV/DIV/DIV/DIV/DIV/DIV[1]/DIV/DIV/DIV[2]/DIV/DIV/DIV[1]`
```
ashchan: This won't be easy to maintain in the future. There should be semantic class or id or other properties we can refer to UI elements. We can add that to the app if necessary.

packages/neuron-wallet/tests-e2e/env.ts
```diff
@@ -0,0 +1,23 @@
+export interface TransactionSendTo {
+  address: string
+  amount: number
+}
+
+export interface TransactionTestEnv {
+  // Used to test sending transactions, please make sure there is enough balance.
```
ashchan: Send txs requires connecting to a real CKB node. At the moment that might be too much work to setup. We can focus on other fundamental UI and features, and lower priority of on chain operations for now.

cezres: It is designed to make it easier for testers to send multiple receive address transactions.

packages/neuron-wallet/tests-e2e/operations/sendTransaction.ts
cezres: Ok, the element path is cumbersome to write and difficult to maintain after the change. I will add an id to the element for the search.

PR #883, opened 2019-08-20
refactor(neuron-ui): verify the addresses before submission

Keith-CY:
Use SDK to verify the addresses in the `verifyAddress` method.

### DIFF
--- a/packages/neuron-ui/src/locales/en.json
+++ b/packages/neuron-ui/src/locales/en.json
@@ -272,9 +272,7 @@
       "rpc-url-should-have-no-whitespaces": "The RPC URL should have no whitespaces",
       "is-required": "{{field}} is required",
       "is-used": "{{field}} is used",
-      "invalid-address": "{address}} is an invalid address",
-      "invalid-checksum": "{{address}} has an invalid checksum",
-      "address-length-error": "{{address}} has an invalid length"
+      "invalid-address": "{{address}} is an invalid address"
     },
     "sync": {
       "syncing": "Syncing",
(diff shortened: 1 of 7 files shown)


packages/neuron-ui/src/locales/en.json
```diff
@@ -272,7 +271,10 @@
       "rpc-url-should-have-protocol": "The RPC URL should start with http(s)://",
       "rpc-url-should-have-no-whitespaces": "The RPC URL should have no whitespaces",
       "is-required": "{{field}} is required",
-      "is-used": "{{field}} is used"
+      "is-used": "{{field}} is used",
+      "invalid-address": "{address}} is an invalid address",
+      "invalid-checksum": "{{address}} has an invalid checksum",
```
ashchan: We can use `invalid-address` translation for checksum too. People usually don't understand what checksum is.

packages/neuron-ui/src/locales/en.json
```diff
@@ -272,7 +271,10 @@
...
+      "address-length-error": "{{address}} has an invalid length"
```
ashchan: Same as above, although we have different kinds of invalid reasons, show something like "The address you input is invalid" or so is more friendly and generically enough.

packages/neuron-ui/src/locales/en.json
Keith-CY: `invalid-address` and `invalid-checksum` has been removed in https://github.com/nervosnetwork/neuron/pull/883/commits/bba929219e50cf1eef300fde610d7291aedbaf88

PR #884, opened 2019-08-20
refactor(neuron-ui): add + sign ahead of amount values in the History…

Keith-CY:
… View

image

### DIFF
--- a/packages/neuron-ui/src/utils/formatters.ts
+++ b/packages/neuron-ui/src/utils/formatters.ts
@@ -99,12 +99,17 @@ export const CKBToShannonFormatter = (amount: string = '0', unit: CapacityUnit) 
   }
 }
 
-export const shannonToCKBFormatter = (shannon: string = '0') => {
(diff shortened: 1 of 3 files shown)


packages/neuron-ui/src/utils/formatters.ts
```diff
@@ -81,8 +81,13 @@ export const CKBToShannonFormatter = (amount: string, uint: CapacityUnit) => {
   }
 }
 
-export const shannonToCKBFormatter = (shannon: string) => {
-  const sign = shannon.startsWith('-') ? '-' : ''
+export const shannonToCKBFormatter = (shannon: string, showSign?: boolean) => {
```
ashchan: Is `showPositiveSign` more precise?

Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/884/commits/e40a276ed25b333a3098ad1b02467c9a1455cdd1

PR #885, opened 2019-08-20
feat: not start from zero when create wallet

classicalliu:


packages/neuron-wallet/src/controllers/wallets/index.ts
```diff
@@ -72,6 +72,7 @@ export default class WalletsController {
       name,
       password,
       mnemonic,
+      isImport: true,
```
ashchan:
```suggestion
      isImporting: true,
```

classicalliu: `isImport` means this wallet is generate by import or create,  I'm not sure `isImporting` is better to describe this ?

classicalliu: `isImport` means this wallet is generate by import or create,  I'm not sure `isImporting` is better to describe this ?

ashchan: Then perhaps `isImported` is more appropriate.

classicalliu: `isImported` is more like the `import` action done or not, what do you think?

classicalliu: Renamed to `isImporting`

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -5,9 +5,15 @@ import LockUtils from 'models/lock-utils'
 import AddressDao, { Address as AddressInterface } from 'database/address/dao'
 import env from 'env'
 import AddressEntity, { AddressVersion } from 'database/address/entities/address'
+import AddressCreatedSubject from 'models/subjects/address-created-subject'
 
 const MAX_ADDRESS_COUNT = 30
 
+export interface AddressWithWay {
```
Keith-CY: The name is confusing, how about `AddressAndIsImporting`

classicalliu: `isImporting` moves to `Address` structure, so this struct deleted

PR #888, opened 2019-08-21
feat(neuron-ui): add verification on updating amounts

Keith-CY:
disable the amount to be updated if the value is not a number or has more than 8 digits decimal

### DIFF
--- a/packages/neuron-ui/src/components/Send/hooks.ts
+++ b/packages/neuron-ui/src/components/Send/hooks.ts
@@ -4,14 +4,12 @@ import { IDropdownOption } from 'office-ui-fabric-react'
 import { AppActions, StateDispatch } from 'states/stateProvider/reducer'
 import { calculateCycles } from 'services/remote/wallets'
 
-import { Message } from 'utils/const'
+import { Message, MAX_DECIMAL_DIGITS } from 'utils/const'
 import { verifyAddress, verifyAmountRange } from 'utils/validators'
 import { outputsToTotalCapacity } from 'utils/formatters'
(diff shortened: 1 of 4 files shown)


Keith-CY:
image
image

Do these messages look better? @ashchan

ashchan:
> image
> image
> 
> Do these messages look better? @ashchan

Yes definitely. Two small changes:

1. `Please set a value no less than 61 CKB`.
2. `Please enter an amount with no more than 8 decimal places`.  - This is to emphasis `8`. The message from the above screenshot could very easily give an impression that we have 9 decimals.

Keith-CY:
> > image
> > image
> > Do these messages look better? @ashchan
> 
> Yes definitely. Two small changes:
> 
> 1. `Please enter an amount no less than 61 CKB`.
> 2. `Please enter an amount with no more than 8 decimal places`.  - This is to emphasis `8`. The message from the above screenshot could very easily give an impression that we have 9 decimals.

Updated in https://github.com/nervosnetwork/neuron/commit/e38c3f86afe081a92bebaa0e117947c411c291b0

```json
{
      "amount-decimal-exceed": "The amount {{amount}} CKB is invalid, please enter an amount with no more than 8 decimal places",
      "amount-decimal-exceed": "金额 {{amount}} 是个无效的值, 其小数位应不超过 8 位",
      "amount-too-small": "The amount {{amount}} CKB is too small, please enter an amount no less than 61 CKB",
      "amount-too-small": "金额 {{amount}} 太小, 请输入一个不小于 61 CKB 的值"

}

PR #889, opened 2019-08-21
chore: balance = capacity of live cells + sent cells

classicalliu:


classicalliu:
> Can we add a test for this?

Tests added.

Keith-CY:
CI failed

packages/neuron-wallet/tests/database/address/balance.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  it('balance = live + sent - pending', async () => {
+    const address = generateAddress('1000', '100', '300')
+    const addrs: AddressEntity[] = await AddressDao.create([address])
+    const addr = addrs[0]
+    expect(addr.balance()).toEqual((1000 + 100).toString())
+  })
+
+  it('balance is right in interface', async () => {
```
Keith-CY:
```suggestion
  it('balance returned by the interface method', async () => {
```

packages/neuron-wallet/tests/database/address/balance.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    ]
+    const addrs: AddressEntity[] = await AddressDao.create(addresses)
+    const balance: bigint = addrs
+      .map(addr => BigInt(addr.balance()))
+      .reduce((result, c) => result + c, BigInt(0))
+
+    expect(balance).toEqual(BigInt(1000))
+  })
+
+  it('sent to others with 10shannon fee', async () => {
```
Keith-CY:
```suggestion
  it('sent to others with 10 shannon fee', async () => {
```

packages/neuron-wallet/tests/database/address/balance.test.ts
classicalliu: This test this focus on `balance` is right or not, not returned or not

Keith-CY: > This test this focus on `balance` is right or not, not returned or not

The description is a bit confusing, `balance is right in interface` means `the balance field is listed in the interface exactly`, try to use a better description, such as `the balance returned by the interface method is correct`

PR #890, opened 2019-08-22
Set cycles empty if tx invalid

Keith-CY:


Keith-CY:
It seems that the rebase doesn't work very well, I'd like to create another PR.

PR #903, opened 2019-08-30
feat: skip cells which has data or type when calculate balance

classicalliu:
Now the `balance` not include cells which has data or type, and `totalBalance` includes.
Has a setting to set can use cells with data/type or not.


packages/neuron-wallet/src/services/skip-data-and-type.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  static getInstance(): SkipDataAndType {
+    if (!SkipDataAndType.instance) {
+      SkipDataAndType.instance = new SkipDataAndType()
+    }
+
+    return SkipDataAndType.instance
+  }
+
+  // open means can use cells with data and type
```
Keith-CY: Where is the `open` used?

packages/neuron-wallet/src/services/skip-data-and-type.ts
```diff
@@ -0,0 +1,51 @@
...
+      JSON.stringify({
+        skip,
+      })
+    )
+    // cache this variable
+    this.skip = skip
+  }
+
+  public get(): boolean {
+    // if cached, don't to read file
```
Keith-CY:
```suggestion
    // if cached, don't read file
```

packages/neuron-wallet/src/services/skip-data-and-type.ts
classicalliu: Renamed to `skip` ...

packages/neuron-wallet/src/database/address/dao.ts
```diff
@@ -67,9 +70,12 @@ export default class AddressDao {
         const addressEntity = entity
         addressEntity.txCount = txCount
         const lockHashes: string[] = await LockUtils.addressToAllLockHashes(addressEntity.address)
-        addressEntity.liveBalance = await CellsService.getBalance(lockHashes, OutputStatus.Live)
-        addressEntity.sentBalance = await CellsService.getBalance(lockHashes, OutputStatus.Sent)
-        addressEntity.pendingBalance = await CellsService.getBalance(lockHashes, OutputStatus.Pending)
+        addressEntity.liveBalance = await CellsService.getBalance(lockHashes, OutputStatus.Live, true)
```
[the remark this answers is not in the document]
classicalliu: No need, it's not slow and avoid database timeout ...

packages/neuron-wallet/src/services/skip-data-and-type.ts
```diff
@@ -0,0 +1,51 @@
+import FileService from './file'
+
+export default class SkipDataAndType {
+  private static moduleName = ''
+  private static fileName = 'skip-data-and-type.json'
```
ashchan: How about a global `settings.json` for all related persisted app configuration? I don't see any benefit maintaining a single file for each feature.

classicalliu: Add a `BaseSettings` to read/update settings and read/update `SkipDataAndType` via this.

PR #904, opened 2019-09-02
feat: add an edit button by the side of description

Keith-CY:
Add an edit button aside the description field
image
image


packages/neuron-ui/src/utils/hooks.ts
```diff
@@ -1,6 +1,7 @@
 import { useState, useCallback } from 'react'
 import { updateTransactionDescription, updateAddressDescription } from 'states/stateProvider/actionCreators'
 import { StateDispatch } from 'states/stateProvider/reducer'
+import { AppActions } from '../states/stateProvider/reducer'
```
ashchan: Can we keep consistence for importing (See line 3 and 4)?

ashchan:
```suggestion
import { AppActions } from 'states/stateProvider/reducer'
```

PR #907, opened 2019-09-03
feat(neuron-ui): move the miner info from overview to wallet list

Keith-CY:
image
image
image
image


Keith-CY:
@ashchan update the trigger of miner info to texts instead of icon buttons

PR #908, opened 2019-09-03
refactor: adapt to mainnet in wallet

classicalliu:


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -165,7 +165,8 @@ export default class AddressService {
       AddressPrefix.Mainnet
     ).address
 
-    const blake160: string = LockUtils.addressToBlake160(testnetAddress)
+    const addressToParse = env.testnet ? testnetAddress : mainnetAddress
```
ashchan: Any plan to move this out of `Env`, placing into its own storage/class?

classicalliu: Will do this later, don't know how to judge mainnet now.

ashchan: got that.

PR #925, opened 2019-09-11
Improve qr scanner

Keith-CY:
According to [some tutorials](https://www.qrcode-monkey.com/6-reasons-why-your-qr-code-is-not-working)
>6. Leave a border space around your QR Code
Always leave a space between your QR code and the content or design around it. The QR Code should always stand alone on the background and the foreground should not touch other elements. It is important for QR code scanner apps to detect the corner elements as stand-alone objects.

I've added a blank border around the QR Code like this:
image
While an indent between the Image and buttons is introduced on the Receive View
image

I prefer to keep it since the blank margin is a part of the QR Code and not aberrant.

### DIFF
--- a/packages/neuron-ui/src/components/Receive/index.tsx
+++ b/packages/neuron-ui/src/components/Receive/index.tsx
@@ -63,6 +63,7 @@ const Receive = ({
             onQRCodeClick={() => setShowLargeQRCode(true)}
             size={256}
             exportable
+            includeMargin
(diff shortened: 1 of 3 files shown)


Keith-CY:
> Does that improve the situation under which I download the QR image to my phone's camera roll and scan that image in full screen mode?

Yes, it works well when the QR Code is displayed in full-screen mode.

PR #926, opened 2019-09-11
chore: deindex lock hashes when delete wallet

classicalliu:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -263,7 +266,30 @@ export default class WalletService {
 
     this.listStore.writeSync(this.walletsKey, newWallets)
     wallet.deleteKeystore()
-    AddressService.deleteByWalletId(id)
+    this.deleteAddressesAndDeindex(id)
+  }
+
+  private deleteAddressesAndDeindex = async (walletID: string) => {
```
ashchan: Functions with naming like `doSomethingAndSomething` usually indicate they're doing more than one thing. If we have a `deindexAddresses` and pass `addresses` returned from `AddressService.deleteByWalletId` to that function, then we have a function that does one thing only. Something like:

```javascript
addressInterfaces = await AddressService.deleteByWalletId(walletID)
this.deindexAddresses(addressInterfaces)
```

classicalliu: updated

PR #929, opened 2019-09-11
fix(neuron-ui): fix the transaction detail view

Keith-CY:
1. hide the deprecated block hash
2. fix the display of cell out point

### DIFF
--- a/packages/neuron-ui/src/components/Transaction/index.tsx
+++ b/packages/neuron-ui/src/components/Transaction/index.tsx
@@ -16,23 +16,25 @@ const inputColumns: IColumn[] = [
     key: 'lockHash',
     name: 'Lock Hash',
     minWidth: 100,
-    maxWidth: 525,
-  },
-  {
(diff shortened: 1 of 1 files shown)


packages/neuron-ui/src/components/Transaction/index.tsx
```diff
@@ -16,23 +16,25 @@ const inputColumns: IColumn[] = [
...
+    maxWidth: 200,
+    onRender: (item: any) => (
+      <span title={item.lockHash || 'none'} className="text-overflow">
+        {item.lockHash || 'none'}
+      </span>
+    ),
   },
   {
     key: 'outPointCell',
     name: 'OutPoint Cell',
```
classicalliu: I think `outPoint` is enough.

Keith-CY: For now, I just keep it working in the desired way. We can remove it in the next version.

PR #931, opened 2019-09-12
feat: Improve localization translations

ashchan:


packages/neuron-ui/src/locales/en.json
```diff
@@ -210,18 +209,6 @@
     "footer": {
       "fail-to-fetch-tip-block-number": "Cannot fetch tip block number"
     },
-    "menuitem": {
```
ashchan: @Keith-CY I guess we don't use these any more. Can you check and confirm?

Keith-CY: Yes, they are not used any more.

ashchan: Nice. Thanks!

packages/neuron-ui/src/locales/en.json
```diff
@@ -243,10 +230,10 @@
     "messages": {
       "error": "Error",
       "unknown-error": "Unknown error",
-      "update-wallet-successfully": "Update the wallet successfully",
-      "delete-wallet-successfully": "Delete the wallet successfully",
-      "create-network-successfully": "Create a network successfully",
-      "update-network-successfully": "Update the network successfully",
+      "update-wallet-successfully": "The wallet was updated",
+      "delete-wallet-successfully": "The Wallet was deleted",
```
classicalliu: `Wallet` or `wallet` ?

Keith-CY: I think it should be `wallet`

PR #939, opened 2019-09-17
feat(neuron-ui): add mainnet/testnet addresses toggle

Keith-CY:
image
image

And disable context menu when the mainnet addresses are displayed.

### DIFF
--- a/packages/neuron-ui/src/components/Addresses/index.tsx
+++ b/packages/neuron-ui/src/components/Addresses/index.tsx
@@ -1,16 +1,20 @@
-import React, { useEffect, useMemo } from 'react'
+import React, { useState, useEffect, useMemo } from 'react'
 import { RouteComponentProps } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import {
+  Stack,
   ShimmeredDetailsList,
   TextField,
   IColumn,
   CheckboxVisibility,
+  DefaultButton,
   IconButton,
+  Text,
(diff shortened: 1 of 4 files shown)


packages/neuron-ui/src/components/Addresses/index.tsx
```diff
@@ -181,18 +186,44 @@ const Addresses = ({
...
-        items={addresses}
+        items={addresses.map(addr => ({
+          ...addr,
+          address: showMainnetAddress
+            ? ckbCore.utils.bech32Address(addr.identifier, {
+                prefix: ckbCore.utils.AddressPrefix.Mainnet,
+                type: ckbCore.utils.AddressType.HashIdx,
+                codeHashIndex: '0x00',
+              }) || ''
+            : addr.address,
```
ashchan: Is this assuming current network is always `testnet`? It should check current network. (I know `network.type` is something not implemented, but want to know that this would get updated and delegate to that property).

ashchan: Same for `showMainnetAddress`'s initial state.

Keith-CY: I've added a condition to determine whether the toggle and caution should be displayed in https://github.com/nervosnetwork/neuron/pull/939/commits/5be03351d1c28b90c88b887024aa0a8a4c2ad156

PR #945, opened 2019-09-19
fix: create tx in several SQLs

classicalliu:


packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -157,10 +158,44 @@ export class TransactionPersistor {
...
+      await queryRunner.commitTransaction();
+    } catch (err) {
+      await queryRunner.rollbackTransaction()
+    } finally {
+      await queryRunner.release()
+    }
     return tx
   }
 
+  private static waitTransaction = async(queryRunner: QueryRunner, timeout: number = 5000) => {
```
ashchan: What does this mean? Wait for a transaction to do what?

classicalliu: Wait transaction end.

Keith-CY: It seems to be `waitUntilTransactionFinished`

classicalliu: updated

PR #946, opened 2019-09-19
Optimize list with target property

Keith-CY:
1. use data-* as the item identifier
2. align the QR scanner video to the center of the dialog

### DIFF
--- a/packages/neuron-ui/src/components/Addresses/index.tsx
+++ b/packages/neuron-ui/src/components/Addresses/index.tsx
@@ -112,12 +112,14 @@ const Addresses = ({
           return item ? (
             <>
               <TextField
+                data-description-key={item.address}
+                data-description-value={item.description}
                 borderless
                 title={item.description}
                 value={isSelected ? localDescription.description : item.description || ''}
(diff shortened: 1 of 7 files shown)


Keith-CY:
> Just out of curiosity, by `Optimize` do we mean this would make the list more performant? If so do we have some tweak data to compare before & after?

Sorry for deficiency of data, in this update, the event handler is changed from

```javascript
const eventHandler = (key: string) => (e, value) => { update({key, value}) }
```

To

```javascript
const eventHandler = (e, value) => { update({ key: e.target.dataset.key, value }) }
```

An extra cost from creating function is reduced.

ashchan:
> > Just out of curiosity, by `Optimize` do we mean this would make the list more performant? If so do we have some tweak data to compare before & after?
> 
> Sorry for deficiency of data, in this update, the event handler is changed from
> 
> ```js
> const eventHandler = (key: string) => (e, value) => { update({key, value}) }
> ```
> 
> To
> 
> ```js
> const eventHandler = (e, value) => { update({ key: e.target.dataset.key, value }) }
> ```
> 
> An extra cost from creating function is reduced.

I see. Thanks.

PR #952, opened 2019-09-23
feat: bump sdk to v0.21.0

classicalliu:


packages/neuron-wallet/src/utils/hex.ts
```diff
@@ -0,0 +1,9 @@
+export default class HexUtils {
+  public static toDecimal(hex: string): string {
```
Keith-CY: It should be named as `toInt`

classicalliu: It's a convert between `decimal` and `hex`, not `Integer` and  `String`

PR #970, opened 2019-10-07
feat: bump sdk to v0.22.0

classicalliu:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -443,19 +443,15 @@ export default class WalletService {
     return addr!.address
   }
 
-  public signWitness = (witness: Witness, privateKey: string, txHash: string): Witness => {
+  public signWitness = (witness: string, privateKey: string, txHash: string): string => {
```
Keith-CY: If it is possible to use the SDK to sign witnesses
https://github.com/nervosnetwork/ckb-sdk-js/blob/develop/packages/ckb-sdk-core/src/index.ts#L158

classicalliu: updated

PR #975, opened 2019-10-09
chore: catch `updateMetaInfo` error when init database

classicalliu:


packages/neuron-wallet/src/startup/sync-block-task/init-database.ts
```diff
@@ -9,8 +9,13 @@ export const initDatabase = async () => {
   try {
     const hash = await genesisBlockHash()
     await initConnection(hash)
-    const systemScriptInfo = await LockUtils.systemScript()
-    updateMetaInfo({ genesisBlockHash: hash, systemScriptInfo })
+
```
Keith-CY: It seems to be nested in a try block.

classicalliu: Yes, get systemScriptInfo failed should not bother initConnection.
But if `initConnection` failed, should not execute those codes.

Keith-CY: If it is possible to run these two try blocks sequentially.

classicalliu: > But if `initConnection` failed, should not execute those codes.

If in two try blocks, no matter `initConnection` success or not, the second try block always be executed.

PR #976, opened 2019-10-09
Dismiss alerts if correct actions are taken

Keith-CY:


packages/neuron-wallet/tests-e2e/tests/notification.ts
```diff
@@ -11,6 +11,7 @@ import { sleep } from '../application/utils'
  * 6. check the notification, it should have two messages
  *   1. incorrect PasswordRequest
  *   2. disconnected to the network
+ * 6. password-incorrect alerts should be dismissed once a correct one is inputted
```
classicalliu:
```suggestion
 * 7. password-incorrect alerts should be dismissed once a correct one is inputted
```

PR #977, opened 2019-10-11
feat(neuron-ui): update the epoch index

Keith-CY:


packages/neuron-ui/src/utils/parsers.ts
```diff
@@ -23,4 +23,12 @@ export const prompt = (search: string) => {
 }
 export const queryParsers = { history, prompt }
 
-export default { queryParsers }
+/* eslint-disable no-bitwise */
```
ashchan: For such rule why don't we simply disable it from config?

I see no reason to allow bitwise operations in some places and disallow in some others.

Keith-CY: This rule is removed in https://github.com/nervosnetwork/neuron/pull/977/commits/c5c51012e1dd2c5c96e74d789e0aeee5eadfae4c

PR #987, opened 2019-10-15
feat(neuron-ui): display address field of input on the transaction de…

Keith-CY:
…tail view

image

### DIFF
--- a/packages/neuron-ui/src/components/Transaction/index.tsx
+++ b/packages/neuron-ui/src/components/Transaction/index.tsx
@@ -58,7 +58,7 @@ const Transaction = () => {
               return null
             }
             if (!input.lock) {
-              return t('transaction.cell-from-coinbase')
+              return t('transaction.cell-from-cellbase')
             }
             try {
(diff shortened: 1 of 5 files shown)


packages/neuron-ui/src/components/Transaction/index.tsx
```diff
@@ -27,25 +41,42 @@ const Transaction = () => {
...
+          key: 'address',
+          name: t('transaction.address'),
+          minWidth: 200,
+          maxWidth: 500,
+          onRender: (input?: State.DetailedInput, _index?: number, column?: IColumn) => {
+            if (!input) {
+              return null
+            }
+            if (!input.lock) {
+              return t('transaction.cell-from-coinbase')
```
classicalliu: `cellbase` or `coinbase` ?

Keith-CY: fixed in https://github.com/nervosnetwork/neuron/pull/987/commits/27df45391736ff96612ab85f898784dce7da34e3

PR #995, opened 2019-10-17
Friendly output name

ashchan:
* QR image file: replace `Receive.png` with `ckb-address.png`.
* Keystore file: use `[wallet name].json` instead of `[wallet name]`.

### DIFF
--- a/packages/neuron-ui/src/widgets/QRCode/index.tsx
+++ b/packages/neuron-ui/src/widgets/QRCode/index.tsx
@@ -121,7 +121,7 @@ const QRCode = ({
     }
     const dataURL = canvasRef.current.toDataURL('image/png')
     const downloadLink = document.createElement('a')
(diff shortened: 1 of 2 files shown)


packages/neuron-ui/src/widgets/QRCode/index.tsx
```diff
@@ -121,7 +121,7 @@ const QRCode = ({
     }
     const dataURL = canvasRef.current.toDataURL('image/png')
     const downloadLink = document.createElement('a')
-    downloadLink.download = 'Receive'
+    downloadLink.download = 'dkb-address'
```
Keith-CY: Do you mean `ckb`?

ashchan: Oh shit. Auto correction.

PR #997, opened 2019-10-17
refactor: Reduce nesting levels, move app/content menu

ashchan:
construction to app controller

The overview purpose of this refactoring and more coming ones is to reduce files/objects and complexity.


ashchan:
Fixing e2e tests

PR #999, opened 2019-10-18
chore: check is mainnet or not

classicalliu:


packages/neuron-wallet/src/models/chain-info.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  public setChain = (chain: string) => {
+    this.chain = chain
+  }
+
+  public getChain = (): string => {
+    return this.chain
+  }
+
+  public isMainnet = (): boolean => {
```
ashchan: This could be as simple as `return this.chain === 'ckb'`.

classicalliu: updated

PR #1000, opened 2019-10-18
test(e2e): refactor goto setting page from menu based on os platform

kellyshang:
for macOS and windows.

### DIFF
--- a/packages/neuron-wallet/tests-e2e/application/index.ts
+++ b/packages/neuron-wallet/tests-e2e/application/index.ts
@@ -222,7 +222,7 @@ export default class Application {
 
   // goto Setting page from menu according to OS platform
   async gotoSettingPageFromMenu() {
-    if(this.osPlatform.includes("darwin")) {
+    if (this.osPlatform.includes("darwin")) {
       await this.clickMenu(['Electron', 'Preferences...'])
     } else if(this.osPlatform.includes("win32")) {
       this.wait(300)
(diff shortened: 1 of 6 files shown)


packages/neuron-wallet/tests-e2e/application/index.ts
```diff
@@ -211,4 +212,21 @@ export default class Application {
...
+  getOSplatform():string {
+    let os = require('os');
+    let platform = os.platform();
+    console.log(`current os platform is ${platform}`)
+    return platform;
+  }
+
+  // goto Setting page from menu according to OS platform
+  async gotoSettingPageFromMenu() {
+    if(this.osPlatform.includes("darwin")) {
```
ashchan:
```suggestion
    if (this.osPlatform.includes("darwin")) {
```

packages/neuron-wallet/tests-e2e/application/index.ts
```diff
@@ -211,4 +212,21 @@ export default class Application {
...
+      await this.clickMenu(['Electron', 'Preferences...'])
+    } else if(this.osPlatform.includes("win32")) {
```
ashchan: On both Windows and Linux `Settings` is under `Help` so the check for `win32` could be removed.

ashchan: Not a big problem for now.

PR #1007, opened 2019-10-21
Api controller

ashchan:
* Add a new `ApiController` which replaces `AppController`
* neuron-ui talks to remote/api instead of remote/app
* AppController is used as instance, not static functions object
* Detach menu from app controller
* Remove controller and UI action dependency from services


ashchan:
Oops forgot to update e2e tests.

PR #1008, opened 2019-10-21
feat: Expose a flat API controller from neuron wallet to UI layer

ashchan:
Neuron UI should require this api controller only.

### DIFF
--- a/packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
+++ b/packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
@@ -21,7 +21,7 @@ export const RemoteNotLoadError = {
   },
 }
 
-export const apiMethodWrapper = () => (
+export const apiMethodWrapper = (
   callControllerMethod: (
     controller: any
   ) => (
--- a/packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
+++ b/packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
@@ -21,14 +21,7 @@ export const RemoteNotLoadError = {
   },
 }
 
-export const controllerNotLoaded = (controllerName: string) => ({
-  status: 0 as 0,
-  message: {
-    content: `${controllerName} controller not loaded`,
-  },
-})
-
-export const controllerMethodWrapper = (controllerName: string) => (
+export const apiMethodWrapper = () => (
   callControllerMethod: (
     controller: any
   ) => (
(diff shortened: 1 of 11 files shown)


packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
```diff
@@ -42,13 +35,18 @@ export const controllerMethodWrapper = (controllerName: string) => (
   if (!window.remote) {
     return RemoteNotLoadError
   }
-  const controller = window.remote.require(`./controllers/${controllerName}`).default
+  const controller = window.remote.require('./controllers/api').default
   if (!controller) {
```
Keith-CY: The validation should include the error of `./controllers/api is not found`

packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
```diff
@@ -75,6 +73,5 @@ export const controllerMethodWrapper = (controllerName: string) => (
 
 export default {
   RemoteNotLoadError,
-  controllerNotLoaded,
-  controllerMethodWrapper,
+  controllerMethodWrapper: apiMethodWrapper,
```
Keith-CY: `controllerMethodWrapper` should be renamed `apiMethodWrapper`

packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
```diff
@@ -21,14 +21,7 @@ export const RemoteNotLoadError = {
...
 
-export const controllerNotLoaded = (controllerName: string) => ({
-  status: 0 as 0,
-  message: {
-    content: `${controllerName} controller not loaded`,
-  },
-})
-
-export const controllerMethodWrapper = (controllerName: string) => (
+export const apiMethodWrapper = () => (
```
Keith-CY: The most outside wrapper is redundant now.

packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
ashchan: What did you mean by including the error of that? I moved that `const` of controller not loadable to inner object and return that.

BTW `RemoteNotLoadError` was not used as well - but it's exported.

packages/neuron-ui/src/services/remote/apiMethodWrapper.ts
ashchan: Updated w/ 18e7a327.

PR #1010, opened 2019-10-22
chore: bump sdk to v0.23.0

classicalliu:
1. bump SDK to v0.23.0
2. split `sendCapacity` to `sendTx` and `generateTx`
3. add `calculateFee`
4. support generate transaction with `feeRate`
5. set default fee rate to 1000 if fee not set in wallets controller `sendCapacity`

### DIFF
--- a/packages/neuron-wallet/src/controllers/wallets.ts
+++ b/packages/neuron-wallet/src/controllers/wallets.ts
@@ -349,6 +349,49 @@ export default class WalletsController {
   }
 
   @CatchControllerError
+  public static async sendCapacity(params: {
+    id: string
+    walletID: string
+    items: {
+      address: string
+      capacity: string
+    }[]
+    password: string
+    fee: string
+    description?: string
+  }) {
+    if (!params) {
+      throw new IsRequired('Parameters')
+    }
+    // set default feeRate
+    let feeRate = '0'
+    if (!params.fee || params.fee === '0') {
+      feeRate = '1000'
+    }
+    try {
+      const walletsService = WalletsService.getInstance()
+      const hash = await walletsService.sendCapacity(
+        params.walletID,
+        params.items,
+        params.password,
+        params.fee,
+        feeRate,
+        params.description
+      )
+      return {
+        status: ResponseCode.Success,
+        result: hash,
+      }
+    } catch (err) {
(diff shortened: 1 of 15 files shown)


ashchan:
Too quick to click approve button. Sorry.

Keith-CY:
ckb-sdk-js@v0.23.1 has been published

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -348,35 +349,91 @@ export default class WalletsController {
...
         params.items,
-        params.password,
         params.fee,
-        params.description
+        params.feeRate,
       )
       return {
         status: ResponseCode.Success,
-        result: hash,
+        result: JSON.stringify(tx),
```
Keith-CY: Why use `stringify` here?

classicalliu: Just a serialization here, can do it or not.

ashchan: Let’s do that only to the api controller later.

classicalliu: Removed now.

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -143,7 +161,22 @@ export default class CellsService {
     return {
       inputs,
       capacities: inputCapacities.toString(),
+      needFee: needFee.toString(),
+    }
+  }
+
+  public static everyInputFee = (feeRate: bigint): bigint => {
+    /*
+    * every input need 44 Bytes
```
ashchan:
```suggestion
    * every input needs 44 Bytes
```

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -143,7 +161,22 @@ export default class CellsService {
...
+    * every input need 1 witness signed by secp256k1, with 65 Bytes data, serialized in 69 Bytes, add extra 4 Bytes when add to transaction.
```
ashchan:
```suggestion
    * every input needs 1 witness signed by secp256k1, with 65 Bytes data, serialized in 69 Bytes, add extra 4 Bytes when add to transaction.
```

PR #1011, opened 2019-10-22
feature: Allow creating LockUtils instance

ashchan:
Eliminate nested async calls and maps. Other than loading the system script,
lock utils should be able to handle batch of addresses <=> lock hashes
in synchronous manner.

### DIFF
--- a/packages/neuron-wallet/src/database/address/dao.ts
+++ b/packages/neuron-wallet/src/database/address/dao.ts
@@ -69,11 +69,11 @@ export default class AddressDao {
       TransactionStatus.Pending,
       TransactionStatus.Success,
     ], url)
+    const lockUtils = new LockUtils(await LockUtils.systemScript(url))
     const entities = await Promise.all(
       addressEntities.map(async entity => {
         const addressEntity = entity
         addressEntity.txCount = txCount
-        const lockUtils = new LockUtils(await LockUtils.loadSystemScript(url))
         const lockHashes: string[] = lockUtils.addressToAllLockHashes(addressEntity.address)
         addressEntity.liveBalance = await CellsService.getBalance(lockHashes, OutputStatus.Live, true)
         addressEntity.sentBalance = await CellsService.getBalance(lockHashes, OutputStatus.Sent, true)
--- a/packages/neuron-wallet/src/database/address/dao.ts
+++ b/packages/neuron-wallet/src/database/address/dao.ts
@@ -73,7 +73,8 @@ export default class AddressDao {
       addressEntities.map(async entity => {
         const addressEntity = entity
         addressEntity.txCount = txCount
-        const lockHashes: string[] = await LockUtils.addressToAllLockHashes(addressEntity.address, url)
+        const lockUtils = new LockUtils(await LockUtils.loadSystemScript(url))
+        const lockHashes: string[] = lockUtils.addressToAllLockHashes(addressEntity.address)
         addressEntity.liveBalance = await CellsService.getBalance(lockHashes, OutputStatus.Live, true)
         addressEntity.sentBalance = await CellsService.getBalance(lockHashes, OutputStatus.Sent, true)
         addressEntity.pendingBalance = await CellsService.getBalance(lockHashes, OutputStatus.Pending, true)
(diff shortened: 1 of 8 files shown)


packages/neuron-wallet/src/startup/sync-block-task/sync.ts
```diff
@@ -44,19 +44,18 @@ export const switchNetwork = async (url: string, genesisBlockHash: string, chain
...
-          const hashes: string[] = await LockUtils.addressToAllLockHashes(addr.address)
-          // undefined means false
-          const isImporting: boolean = addr.isImporting === true
-          return hashes.map(h => {
-            return {
-              lockHash: h,
-              isImporting,
-            }
-          })
+      const lockUtils = new LockUtils(await LockUtils.loadSystemScript(url))
```
Keith-CY: These parts look like the same as https://github.com/nervosnetwork/neuron/pull/1011/files#diff-1449216a4a57add5c2a9df562c148952R45-R56

ashchan: Duplicated but not a big issue. `sync` is for non-indexer module and `indexer` is for indexer module. I'd say it's ok to have them not reusable to isolate the implementations.

BTW reusing code or logic is not concern of this PR.

ashchan:
Please do NOT merge until @classicalliu gives it a look.

packages/neuron-wallet/src/database/address/dao.ts
```diff
@@ -73,7 +73,8 @@ export default class AddressDao {
       addressEntities.map(async entity => {
         const addressEntity = entity
         addressEntity.txCount = txCount
-        const lockHashes: string[] = await LockUtils.addressToAllLockHashes(addressEntity.address, url)
+        const lockUtils = new LockUtils(await LockUtils.loadSystemScript(url))
```
classicalliu: I think here should use `LockUtils.systemScript(url)` to adapt offline mode.

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -306,7 +307,7 @@ export class TransactionsService {
     status: TransactionStatus[],
     url: string = NodeService.getInstance().core.rpc.node.url
   ): Promise<number> => {
-    const lockHashes: string[] = await LockUtils.addressToAllLockHashes(address, url)
+    const lockHashes: string[] = new LockUtils(await LockUtils.loadSystemScript(url)).addressToAllLockHashes(address)
```
classicalliu: Also here.

packages/neuron-wallet/src/startup/sync-block-task/sync.ts
```diff
@@ -20,9 +20,9 @@ export interface LockHashInfo {
 // maybe should call this every time when new address generated
 // load all addresses and convert to lockHashes
 export const loadAddressesAndConvert = async (nodeURL: string): Promise<string[]> => {
-  const addresses: string[] = (await AddressService.allAddresses()).map(addr => addr.address)
-  const lockHashes: string[] = await LockUtils.addressesToAllLockHashes(addresses, nodeURL)
-  return lockHashes
+  const lockUtils = new LockUtils(await LockUtils.loadSystemScript(nodeURL))
```
classicalliu: Also here.

packages/neuron-wallet/src/startup/sync-block-task/sync.ts
classicalliu: And here.

ashchan: After changes from ed71600 the only places having `loadSystemScript` is `sync-block-task/indexer`. Should that use `systemScript(url)` instead as well?

classicalliu: If in offline mode, it will switch to normal sync, so when switch to indexer, the network would be available, so I think replace or not is not important.

ashchan: I see. Thanks.

PR #1015, opened 2019-10-23
feat(neuron-ui): update the Send View according to the new transactio…

Keith-CY:
…n fee model.

image

### DIFF
--- a/packages/neuron-ui/src/components/TransactionFeePanel/index.tsx
+++ b/packages/neuron-ui/src/components/TransactionFeePanel/index.tsx
@@ -4,30 +4,24 @@ import { useTranslation } from 'react-i18next'
 
 interface TransactionFee {
   fee: string
-  cycles: string
   price: string
   onPriceChange: any
 }
 
 const calculateSpeed = (price: number) => {
-  if (price >= 160) {
-    return '180'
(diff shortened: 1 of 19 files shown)


packages/neuron-ui/src/components/TransactionFeePanel/index.tsx
```diff
@@ -79,7 +84,7 @@ const TransactionFee: React.FunctionComponent<TransactionFee> = ({
             <Label>{t('send.price')}</Label>
           </Stack.Item>
           <Stack.Item grow>
-            <TextField value={price} onChange={onPriceChange} aria-label="price" />
+            <TextField value={price} onChange={onPriceChange} aria-label="price" suffix="shannons/kB" />
```
classicalliu:
```suggestion
            <TextField value={price} onChange={onPriceChange} aria-label="price" suffix="shannons/KB" />
```

Keith-CY: I'd like to say, `kB` is more precise.
image

classicalliu: ok

PR #1019, opened 2019-10-25
feat: display disconnection errors and dismiss them on getting connec…

Keith-CY:
…ted.

UI will display the alerts of disconnection from neuron-wallet, and clear them once UI find that connection is built.
There may be some status conflicts between UI and wallet, but it's acceptable.


ashchan:
Please check CI status.

Keith-CY:
> Please check CI status.

Fixed

PR #1029, opened 2019-10-28
refactor: remove some useless db query

classicalliu:


packages/neuron-wallet/src/services/tx/indexer-transaction.ts
```diff
@@ -61,10 +61,11 @@ export default class IndexerTransaction {
       .where({
         outPointTxHash: txHash,
         outPointIndex: index,
+        status: Not(OutputStatus.Dead)
```
ashchan: This might be lower since we don't have index on `[outPointTxHash, outPointIndex, status]` or `[status]`.

`(outPointTxHash, outPointIndex)` is primary key, I believe the previous query before this change is faster and cost less (without having to create a new index).

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -30,37 +30,56 @@ export class TransactionPersistor {
...
+      .findOne(transaction.hash)
 
     // return if success
     if (txEntity && txEntity.status === TransactionStatus.Success) {
       return txEntity
     }
 
     if (txEntity) {
+      // lazy load inputs / outputs
+      const inputEntities = await connection
```
ashchan: Please run and see if loading tx, inputs, outputs separately as three queries are actually faster than loading them as a join query `{ relations: ['inputs', 'outputs'] }`.

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -122,6 +141,7 @@ export class TransactionPersistor {
         const previousOutput: OutputEntity | undefined = await connection.getRepository(OutputEntity).findOne({
           outPointTxHash: outPoint.txHash,
           outPointIndex: outPoint.index,
+          status: Not(inputStatus),
```
ashchan: As commented above.

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -133,28 +153,26 @@ export class TransactionPersistor {
...
-        const data = outputsData[index]
-        if (data && data !== '0x') {
-          output.hasData = true
-        } else {
-          output.hasData = false
-        }
-        return output
-      })
-    )
+    const outputs: OutputEntity[] = transaction.outputs!.map((o, index) => {
```
ashchan: 👍

packages/neuron-wallet/src/services/tx/indexer-transaction.ts
classicalliu: I think we should test it with / without index, to consider rollback or add an index.
If add `status` to query conditions, it will not load data to memory if `status != dead`, it will save some costs.

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
classicalliu: See the code above it, the main reason is only need to load `inputs` and `outputs` if `txEntity.status` not `success`

ashchan: I see. That makes sense.

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -199,13 +199,15 @@ export default class IndexerQueue {
...
           if (!txEntity || !txEntity.blockHash) {
-            for (const input of transaction.inputs!) {
-              const previousTxWithStatus = await this.getBlocksService.getTransaction(input.previousOutput!.txHash)
-              const previousTx = TypeConvert.toTransaction(previousTxWithStatus.transaction)
-              const previousOutput = previousTx.outputs![+input.previousOutput!.index]
-              input.lock = previousOutput.lock
-              input.lockHash = LockUtils.lockScriptToHash(input.lock)
-              input.capacity = previousOutput.capacity
+            if (!txEntity) {
+              for (const input of transaction.inputs!) {
```
Keith-CY: You can use `Promise.all` to handle async calls in a loop

packages/neuron-wallet/src/services/tx/indexer-transaction.ts
classicalliu: I have tested in `test` env ( in a test case).

I created 10k `outputs` in sqlite memory db. (half `live` and half `dead`)

`with status Not(dead)` | `with status Not(dead) and without index` | `only primary key` | `add union index and with status` 
--- | --- | --- | ---
query 1k times (shuffled and first 1k txs) | `360ms` | `318ms` | `291ms`
query 10k times (shuffled and load each tx) | `2667ms` | `2278ms` | `2249ms`
query 100k times (shuffled and load loop 10 times) | `21083ms` | `22047ms` | `20227ms`

And tested created 10k Outputs

`without index` | `with index`
--- | ---
`3192ms` | `3189ms`

classicalliu: Only use primary key is faster than with status but no index, add an index to `outPointTxHash, outPointIndex, status` seems make limited promotion

I'll rollback to query only using `outPointTxHash, outPointIndex`

packages/neuron-wallet/src/services/indexer/queue.ts
classicalliu: `Promise.all` cause problems usually, should use it very carefully.
It not the focus of this PR, will check later.

PR #1031, opened 2019-10-28
perf: cache getTransaction result when sync in indexer mode

classicalliu:


packages/neuron-wallet/src/services/indexer/transaction-cache.ts
```diff
@@ -0,0 +1,36 @@
+import { TransactionWithStatus } from "types/cell-types"
+
+export class TransactionCache {
```
Keith-CY: This `TransactionCache` looks like a Map has specified size`.

If so, you can use a proxy to trap the set method of a Map as follows

```javascript
const map = new Map()
map.set = new Proxy(map.set, {
  apply(target, thisArg, args) {
    if (target.name === 'set') {
      //pretreat
      console.log(args)
    }
    target.apply(thisArg, args)
  }
})
```

packages/neuron-wallet/tests/services/indexer/transaction-cache.test.ts
```diff
@@ -0,0 +1,74 @@
+import { TransactionCache } from '../../../src/services/indexer/transaction-cache'
```
Keith-CY: The `import` path is too long.

Keith-CY: The `import` path could be concise.

packages/neuron-wallet/src/services/indexer/transaction-cache.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  public push(value: TransactionWithStatus) {
+    const key = value.transaction.hash
+
+    if (this.store.has(key)) {
+      return
+    }
+    this.store.set(key, value)
+    if (this.store.size > this.limit) {
+      const firstKey = Array.from(this.store.keys())[0]
```
Keith-CY: You can get the first key by `this.store.keys().next().value`

packages/neuron-wallet/src/services/indexer/queue.ts
```diff
@@ -43,6 +44,7 @@ export default class IndexerQueue {
   private resetFlag = false
 
   private latestCreatedBy: TxUniqueFlagCache = new TxUniqueFlagCache(100)
+  private txCache: TransactionCache = new TransactionCache(100)
```
ashchan: I don't see many benefits to introduce a specific cache implementation. Is using a `Map` directly not sufficient?

When a `limit` is set and the cache reaches that limit, pushing a new element always trigger the `find first and delete that one` action. Why is that even necessary.  (Another action to consider - just an example - when the limit is reached just clear the first 30 or so elements)

But my real argument here is we can just use a `Map` directly - not very necessary to have a thin layer wrapper of that.

ashchan: BTW, do we have a before and after time comparison for the Elon test wallet?

ashchan: One run on my machine. Before v.s. after: `50min` v.s. `25min`. Not bad.

packages/neuron-wallet/src/services/indexer/transaction-cache.ts
classicalliu: OK, I mistakenly thought the `value` is the value of a `Map`, not the key .... 😢

packages/neuron-wallet/src/services/indexer/queue.ts
classicalliu: Not tested on the whole sync process, but I found `getTransaction` always need about `50ms` usually, if cached, it would take about `0~1ms` usually.
The loop would be more faster.

packages/neuron-wallet/src/services/indexer/transaction-cache.ts
classicalliu: Worth considering

PR #1039, opened 2019-11-01
fix(neuron-ui): remove length limitation of password, but should the …

Keith-CY:
…alert instead

image


ashchan:
@Keith-CY Please check this into key manager too.

PR #1042, opened 2019-11-02
chore: bump SDK to v0.24.0

classicalliu:


Keith-CY:
CI failed

classicalliu:
> CI failed

Fixed

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -489,10 +489,15 @@ export default class WalletService {
   }
 
   public signWitness = (witness: string, privateKey: string, txHash: string): string => {
+    const witnessArg: CKBComponents.WitnessArgs = {
+      lock: witness,
```
ashchan: Are names for `lock` and `witness` still appropriate?

classicalliu: What's your point, I don't quite understand what you mean.

PR #1043, opened 2019-11-02
feat: Remove miner info feature

ashchan:


packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -102,7 +86,7 @@ export default class ApiController {
 
     const initState = {
       currentWallet,
-      wallets: [...wallets.map(({ name, id }, idx: number) => ({ id, name, minerAddress: minerAddresses[idx] }))],
+      wallets: [...wallets.map(({ name, id }) => ({ id, name }))],
```
Keith-CY: The `[...wallets]` here is superfluous

ashchan: Leave that for future refactoring.

PR #1055, opened 2019-11-07
Multi script group witness signing

classicalliu:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -371,18 +371,55 @@ export default class WalletService {
...
+        witness: '',
+        blake160,
+      }
+    })
+
+    const lockHashes = new Set(witnessesWithLockHashes.map(w => w.lockHash))
+
+    for (const lockHash of lockHashes) {
+      const firstIndex = witnessesWithLockHashes.findIndex(w => w.lockHash === lockHash)
+      const witnessesArgsWithBlake160 = witnessesWithLockHashes
```
Keith-CY: Now there are `witness`, `witnesses`, `witnessArgs`, `witnessesArgs`, which is bewildering.

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -372,38 +372,54 @@ export default class WalletService {
...
+        .filter(w => w.lockHash === lockHash)
+        .map(w => ({args: w.witnessArgs, blake160: w.blake160}))
       // A 65-byte empty signature used as placeholder
-      witnesses[0].lock = "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
-      const signedWitness = core.signWitness(privateKey)({
+      witnessesArgsWithBlake160[0].args.lock = '0x' + '0'.repeat(130)
+
+      const blake160: string = witnessesArgsWithBlake160[0].blake160
```
ashchan: I feel we should extract this to a function to isolate the private key fetch.

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -371,18 +371,55 @@ export default class WalletService {
     const paths = addressInfos.map(info => info.path)
     const pathAndPrivateKeys = this.getPrivateKeys(wallet, paths, password)
 
-    const witnesses: string[] = inputs!.map((input: Input) => {
+    const witnessesWithLockHashes = inputs!.map((input: Input) => {
```
ashchan: `witnessesWithLockHashes` is really a bad name. If I understand how it was named previously, it should be called `witnessesWithLockHashesAndWitnessArgsAndBlake160` now?

packages/neuron-wallet/src/services/wallets.ts
ashchan: `witnessesWithLockHashes` becomes a bad name now that there're more members.. If I understand how it was named previously, it should be called `witnessesWithLockHashesAndWitnessArgsAndBlake160` now? :P

xxuejie: Or just witnessGenerationData for short

PR #1056, opened 2019-11-07
Skip get previous tx when cellbase

classicalliu:


packages/neuron-wallet/src/services/sync/get-blocks.ts
```diff
@@ -46,7 +46,12 @@ export default class GetBlocks {
         const checkTx = new CheckTx(tx, this.url)
         const addresses = await checkTx.check(lockHashes)
         if (addresses.length > 0) {
-          for (const input of tx.inputs!) {
+          const inputs = tx.inputs!
+          for (let i = 0; i < inputs.length; ++i) {
```
Keith-CY: The loop can start from 1.

packages/neuron-wallet/src/services/sync/get-blocks.ts
```diff
@@ -46,7 +46,12 @@ export default class GetBlocks {
...
+            if (i === 0) {
+              continue
```
ashchan: For what reason the first input should be skipped?

Keith-CY: It's a cellbase and has no previous output.

packages/neuron-wallet/src/services/sync/get-blocks.ts
```diff
@@ -46,7 +46,12 @@ export default class GetBlocks {
         const checkTx = new CheckTx(tx, this.url)
         const addresses = await checkTx.check(lockHashes)
         if (addresses.length > 0) {
-          for (const input of tx.inputs!) {
+          const inputs = tx.inputs!
+          for (let i = 0; i < inputs.length; ++i) {
+            if (i === 0) {
```
ashchan: That's the first tx of a block, not the first input of a tx.

classicalliu: fixed

PR #1065, opened 2019-11-09
feat: update the address prefix and the explorer url according to the chain type

ashchan:
## TODO
- [x] `isMainnet` pending implementation.


Keith-CY:
Updated

PR #1066, opened 2019-11-10
fix(neuron-ui): fix the address prefix on tx view according to the ch…

Keith-CY:
…ain type.


Keith-CY:
I'll fix it in https://github.com/nervosnetwork/neuron/pull/1065

PR #1071, opened 2019-11-11
feat: verify address according to chain type

Keith-CY:


packages/neuron-ui/src/components/Send/index.tsx
```diff
@@ -63,6 +65,8 @@ const Send = ({
   const labelWidth = '140px'
 
   const isAffordable = verifyTotalAmount(totalAmount, fee, balance)
+  const network = networks.find(n => n.id === networkID)
+  const isMainnet = (network && network.chain === 'ckb') || false // TODO: shoudl be replaced after the mainnet tag is introduced
```
classicalliu:
```suggestion
  const isMainnet = (network && network.chain === 'ckb') || false // TODO: should be replaced after the mainnet tag is introduced
```

classicalliu:
Conflicts and CI failed.

PR #1093, opened 2019-11-13
fix(neuron-ui): add decimal validation on deposit value

Keith-CY:
image

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAO/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAO/index.tsx
@@ -9,7 +9,7 @@ import { updateNervosDaoData, clearNervosDaoData } from 'states/stateProvider/ac
 
 import calculateFee from 'utils/calculateFee'
 import { shannonToCKBFormatter, CKBToShannonFormatter } from 'utils/formatters'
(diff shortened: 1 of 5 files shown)


packages/neuron-ui/src/components/NervosDAO/index.tsx
```diff
@@ -56,27 +57,34 @@ const NervosDAO = ({
...
+        setErrorMessage('')
+        clearGeneratedTx()
+
+        const verifyRes = verifyAmount(value)
+        if (verifyRes !== true) {
+          setErrorMessage(t(`messages.codes.${verifyRes.code}`, { fieldName: 'deposit' }))
+          return
+        }
+
         if (+value < MIN_DEPOSIT_AMOUNT) {
```
classicalliu: Please using BigInt for `value` and `MIN_DEPOSIT_AMOUNT`

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/1093/commits/f96c1543d5f881a2822931f8f27291c48dfd0312

PR #1097, opened 2019-11-13
fix: current block number should be -1 if not start

classicalliu:


Keith-CY:
Will this have any effects on UI?

classicalliu:
> Will this have any effects on UI?

Didn't see

PR #1101, opened 2019-11-14
[HOLD] feat(neuron-ui): rename interest to allowance.

Keith-CY:
`APY` will be updated later.


Keith-CY:
Has confirmed to use `compensation and APC`

PR #1119, opened 2019-11-15
fix(neuron-ui): cache the genesis block timestamp in the nervos dao c…

Keith-CY:
…omponent instead of in global

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAO/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAO/index.tsx
@@ -15,7 +15,7 @@ import { MIN_DEPOSIT_AMOUNT, MEDIUM_FEE_RATE, SHANNON_CKB_RATIO, MAX_DECIMAL_DIG
 import { verifyAmount } from 'utils/validators'
 
 import { generateDepositTx, generateWithdrawTx, generateClaimTx } from 'services/remote'
-import { ckbCore } from 'services/chain'
+import { ckbCore, getBlockByNumber } from 'services/chain'
 import { epochParser } from 'utils/parsers'
 
 import DAORecord from 'components/CustomRows/DAORecordRow'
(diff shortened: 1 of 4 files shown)


packages/neuron-ui/src/components/NervosDAO/index.tsx
```diff
@@ -98,6 +99,9 @@ const NervosDAO = ({
   useEffect(() => {
     updateNervosDaoData({ walletID: wallet.id })(dispatch)
     updateDepositValue(`${MIN_DEPOSIT_AMOUNT}`)
+    getBlockByNumber('0x0')
```
ashchan: Why not calling `get_header_by_number` to only fetch the genesis block header, as the block is big and slow to fetch?

packages/neuron-ui/src/utils/calculateGlobalAPC.ts
```diff
@@ -6,16 +6,11 @@ const DAYS_PER_PERIOD = 365 * 4 * 1
...
-export default async (checkPointTimestamp: number, initialTimestamp: number | undefined = cachedGenesisTimestamp) => {
+export default async (checkPointTimestamp: number, initialTimestamp?: number | undefined) => {
   let genesisTimestamp = initialTimestamp
   if (genesisTimestamp === undefined) {
     genesisTimestamp = await getBlockByNumber('0x0')
-      .then(b => {
-        cachedGenesisTimestamp = +b.header.timestamp
-        return cachedGenesisTimestamp
-      })
+      .then(b => +b.header.timestamp)
```
ashchan: Same as above

packages/neuron-ui/src/components/NervosDAO/index.tsx
Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/1119/commits/b8d44731887efed8042469265abdbad5d671ac27

PR #1126, opened 2019-11-15
feat(neuron-ui): add hint for synchronization not started.

Keith-CY:
image

### DIFF
--- a/packages/neuron-ui/src/containers/Footer/index.tsx
+++ b/packages/neuron-ui/src/containers/Footer/index.tsx
@@ -31,7 +31,7 @@ export const SyncStatus = ({
     return <Text variant="xSmall">{t('footer.fail-to-fetch-tip-block-number')}</Text>
   }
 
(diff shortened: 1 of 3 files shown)


packages/neuron-ui/src/containers/Footer/index.tsx
```diff
@@ -31,6 +31,10 @@ export const SyncStatus = ({
     return <Text variant="xSmall">{t('footer.fail-to-fetch-tip-block-number')}</Text>
   }
 
+  if (+syncedBlockNumber < 0) {
```
classicalliu: Please using bigint for `blockNumber: u64`

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/1126/commits/cc3cd039c832bdccbe5b7d9a770fd78b5255f34f

PR #1138, opened 2019-11-16
Clean cache feature

ashchan:


packages/neuron-ui/src/states/stateProvider/actionCreators/app.ts
```diff
@@ -60,6 +60,7 @@ export const initAppState = () => (dispatch: StateDispatch, history: any) => {
     })
 }
 
+// text: an i18n key under `messages`
```
Keith-CY:
```suggestion
```

packages/neuron-wallet/src/database/chain/cleaner.ts
```diff
@@ -0,0 +1,21 @@
...
+  // Delete all sqlite files under cells folder.
+  // It doesn't handle error or throw exception when deleting fails.
+  public static clean() {
+    const folder = path.join(env.fileBasePath, 'cells')
+    fs.readdir(folder, (err, files) => {
+      if (err) {
+        return
+      }
+
+      for (const file of files.filter(f => { return f.endsWith('sqlite') })) {
```
Keith-CY: Better to use `path.extname(filename) === '.sqlite'` to filter the files.

PR #1139, opened 2019-11-17
feat: add generateDepositAllTx

classicalliu:


packages/neuron-wallet/src/services/cells.ts
```diff
@@ -193,20 +193,49 @@ export default class CellsService {
...
+          lockHash: In(lockHashes),
+          status: OutputStatus.Live,
+          hasData: false,
+          typeScript: null,
+        },
+      })
+
+    const inputs: Input[] = cellEntities.map(cell => {
+      return {
+        previousOutput: cell.outPoint(),
```
Keith-CY: It's `previous output` or `previous outpoint`

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -193,20 +193,49 @@ export default class CellsService {
...
-    const base = BigInt(4 + 44 + 89) * feeRate
+    const base = BigInt(CellsService.everyInputSize()) * feeRate
     const fee = base / ratio
     if (fee * ratio < base) {
       return fee + BigInt(1)
     }
     return fee
   }
 
+  public static everyInputSize = (): number => {
```
Keith-CY: I think `inputSize` is a better name

Keith-CY:
Conflicts

PR #1140, opened 2019-11-17
fix: set restart start number to -1

classicalliu:


ashchan:
Related to #1141. This must be the root source.

PR #1147, opened 2019-11-18
feat(neuron-ui): add copy address and copy tx hash context menus on t…

Keith-CY:
…he tx detail view.

image
image
image

### DIFF
--- a/packages/neuron-ui/src/locales/en.json
+++ b/packages/neuron-ui/src/locales/en.json
@@ -241,7 +241,9 @@
       "toggle": {
         "on": "On",
         "off": "Off"
-      }
+      },
+      "copy-tx-hash": "Copy transaction hash",
+      "copy-address": "Copy address"
     },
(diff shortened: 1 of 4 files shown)


packages/neuron-ui/src/locales/en.json
```diff
@@ -241,7 +241,9 @@
       "toggle": {
         "on": "On",
         "off": "Off"
-      }
+      },
+      "copy-tx-hash": "Copy transaction hash",
```
ashchan:
```suggestion
      "copy-tx-hash": "Copy Transaction Hash",
```

packages/neuron-ui/src/locales/en.json
```diff
@@ -241,7 +241,9 @@
...
+      "copy-address": "Copy address"
```
ashchan:
```suggestion
      "copy-address": "Copy Address"
```

PR #1152, opened 2019-11-22
Clean lock info when switch

classicalliu:


packages/neuron-wallet/src/models/dao-utils.ts
```diff
@@ -45,6 +45,10 @@ export default class DaoUtils {
     return DaoUtils.daoScriptInfo
   }
 
+  static cleanInfoWhenSwitchNetwork(): void {
```
Keith-CY: I think `cleanInfo` is enough for this method

packages/neuron-wallet/src/models/lock-utils.ts
```diff
@@ -75,6 +63,10 @@ export default class LockUtils {
     return LockUtils.systemScriptInfo
   }
 
+  static cleanInfoWhenSwitchNetwork(): void {
```
Keith-CY: `cleanInfo` is enough

packages/neuron-wallet/src/models/dao-utils.ts
classicalliu: updated

PR #1155, opened 2019-11-23
fix: initialize NetworksService in renderer process

classicalliu:
Reduce about `1100 ~ 1200` `TIME_WAIT` connections to about `20 ~ 30`


ashchan:
Fantastic!

PR #1162, opened 2019-11-27
feat: enable deposit all to nervos dao

Keith-CY:
…it's close to max balance.

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAO/DepositDialog.tsx
+++ b/packages/neuron-ui/src/components/NervosDAO/DepositDialog.tsx
@@ -25,7 +25,7 @@ interface DepositDialogProps {
   onChange: any
   onSubmit: any
   onSlide: any
-  balance: string
+  maxDepositAmount: bigint
   isDepositing: boolean
   errorMessage: string
 }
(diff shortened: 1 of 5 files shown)


packages/neuron-ui/src/components/NervosDAO/DepositDialog.tsx
```diff
@@ -52,7 +65,7 @@ const DepositDialog = ({
     ),
     []
   )
-  const maxValue = +(BigInt(balance) / BigInt(SHANNON_CKB_RATIO)).toString()
+  const maxValue = +(maxDepositAmount / BigInt(SHANNON_CKB_RATIO)).toString()
```
classicalliu: Why convert this value to number ?

Keith-CY: > Why convert this value to number ?

The component of Fabric UI only accepts number type, neither bigint nor string.

classicalliu: OK, please make sure it has no precision problems.

PR #1169, opened 2019-11-30
feat: Add generateDepositAllTx API

classicalliu:


packages/neuron-wallet/src/services/cells.ts
```diff
@@ -193,20 +193,49 @@ export default class CellsService {
...
-    const base = BigInt(4 + 44 + 89) * feeRate
+    const base = BigInt(CellsService.everyInputSize()) * feeRate
     const fee = base / ratio
     if (fee * ratio < base) {
       return fee + BigInt(1)
     }
     return fee
   }
 
+  public static everyInputSize = (): number => {
```
ashchan: I really can't get this name 😿. How about renaming it to `inputSize`? (change `everyInputFee` as well).

Issue #1170, opened 2019-11-30
Crash when copying address tx hash from history view


ashchan:
Crash log: https://gist.github.com/Aimeedeer/60ab3fbd9a4529d9d43f51d3b359b8e6

Keith-CY:
@kellyshang Please close this issue after verification.

kellyshang:
> @kellyshang Please close this issue after verification.

OK get.

kellyshang:
Passed verification on Mac with develop branch(commit id: 2582a9b1dca3c856f86a205fcce45886e4342886 ). Tried copy tx hash on History page and address/tx hash on detail page will not crash now. 
BTW, seems I have no permission to close this issue. @Keith-CY you can close it if available.

Keith-CY:
> Passed verification on Mac with develop branch(commit id: [2582a9b](https://github.com/nervosnetwork/neuron/commit/2582a9b1dca3c856f86a205fcce45886e4342886) ). Tried copy tx hash on History page and address/tx hash on detail page will not crash now.
> BTW, seems I have no permission to close this issue. @Keith-CY you can close it if available.

Sure

PR #1171, opened 2019-12-01
Fix copy tx hash

Keith-CY:
Ref: https://github.com/nervosnetwork/neuron/issues/1170

It crashes for invalid reference from MenuEventHandler, the menu object is garbage collected improperly due to a known bug in Electron[https://github.com/electron/electron/pull/21169]

This PR removes the reference between ui and wallet, and use a react hook to cache the menu object.


packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -158,6 +156,11 @@ export default class ApiController {
     return WalletsController.getAllAddresses(id)
   }
 
+  @MapApiResponse
+  public static async requestPassword({ walletID, action }: { walletID: string, action: 'delete-wallet' | 'backup-wallet' }) {
```
ashchan: If I understand this correctly this doesn't need the backend to do any actual work, except dispatching `delete-wallet` or `backup-wallet` action to the UI? If so why not just remove this passing around and call UI action directly?

ashchan: Reason to keep this: Menu triggered action would go through this API from background as well.

PR #1173, opened 2019-12-03
Refactor transaction size and fee calculate logic

classicalliu:
* Add `TransactionSize` for transaction size calculate
* Add `TransactionFee` for transaction fee calculate
* Adjust `gatherInputs` to check tx fee include fee for change output size or not
* Adjust `gatherInputs` to calculate the fee of whole transaction
* Adjust `gatherInputs` to estimate `witnesses` size
* Move `startWithdrawFromDao` main logic to `TransactionGenerator`
*  Add more test cases for generate transactions, especially 2 keys & 3 inputs.


packages/neuron-wallet/src/models/transaction-size.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public static input(): number {
+    return 44
+  }
+
+  public static output(output: Cell): number {
+    const sdkOutput = ConvertTo.toSdkOutput(output)
+    const bytes = serializeOutput(sdkOutput)
+    return Buffer.byteLength(HexUtils.removePrefix(bytes), 'hex') + TransactionSize.SERIALIZED_OFFSET_BYTESIZE
```
Keith-CY: The method of getting buffer.size has been defined in `HexUtils.byteLength`.

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -6,6 +6,9 @@ import { OutputStatus } from './tx/params'
 import FeeMode from 'models/fee-mode'
 import { TransactionStatus } from 'types/cell-types'
 import TransactionEntity from 'database/chain/entities/transaction'
+import { WitnessArgs } from '../types/cell-types';
```
Keith-CY:
```suggestion
```

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -6,6 +6,9 @@ import { OutputStatus } from './tx/params'
 import FeeMode from 'models/fee-mode'
 import { TransactionStatus } from 'types/cell-types'
```
Keith-CY:
```suggestion
import { TransactionStatus, WitnessArgs } from 'types/cell-types'
```

packages/neuron-wallet/src/models/transaction-size.ts
classicalliu: Updated

ashchan:
Let's do a manual test against dev net to make sure this works as expected. cc @kellyshang

kellyshang:
> Let's do a manual test against dev net to make sure this works as expected. cc @kellyshang

OK, in progress now.

PR #1199, opened 2019-12-09
fix: react-hooks/exhaustive-deps warning

fundon:


ashchan:
We are leaving this on purpose as we have a performance issue with this hook dep to deal with later.

PR #1209, opened 2019-12-11
Reset addresses txCount and balance info when clear cache

classicalliu:


packages/neuron-wallet/src/database/chain/cleaner.ts
```diff
@@ -3,10 +3,12 @@ import InputEntity from './entities/input'
 import OutputEntity from './entities/output'
 import TransactionEntity from './entities/transaction'
 import SyncInfoEntity from './entities/sync-info'
+import AddressDao from 'database/address/address-dao'
 
 // Clean local sqlite storage
 export default class ChainCleaner {
   public static async clean() {
+    AddressDao.resetAddresses()
```
ashchan: It should not be placed here as this method is inside chain db section. Instead, call resetAddresses where we call clean.

classicalliu: Updated

PR #1223, opened 2019-12-16
Stop sync if change network and genesis info not match

classicalliu:


packages/neuron-wallet/src/startup/sync-block-task/index.ts
```diff
@@ -46,6 +46,13 @@ const networkChange = async (network: NetworkWithID) => {
     DatabaseInitSubject.getSubject().next(databaseInitParams)
     // re init txCount in addresses if switch network
     await updateAllAddressesTxCount(network.remote)
+  } else {
+    // if killed, should stop previous one
+    DatabaseInitSubject.getSubject().next({
+      network,
+      genesisBlockHash: 'stop',
```
ashchan: Is an empty string even better?

packages/neuron-wallet/src/startup/sync-block-task/task.ts
```diff
@@ -22,7 +22,7 @@ const testIndexer = async (url: string): Promise<boolean> => {
 const run = async () => {
   DatabaseInitSubject.getSubject().subscribe(async (params: DatabaseInitParams) => {
     const { network, genesisBlockHash, chain } = params
-    if (network && genesisBlockHash.startsWith('0x')) {
+    if (network && (genesisBlockHash.startsWith('0x') || genesisBlockHash === 'stop')) {
```
ashchan: Why it should go through to start the queue when `stopped`?

packages/neuron-wallet/src/startup/sync-block-task/index.ts
classicalliu: Means '0x' here ?

packages/neuron-wallet/src/startup/sync-block-task/task.ts
classicalliu: The stop queue logic is in ` syncSwitchNetwork`.

ashchan:
Let's leave it as. My refactoring on sync would mostly wipe these out ;P

ashchan:
Closing this as we're having dramatic changes to this module.

PR #1227, opened 2019-12-17
Change tsconfig target to es2017 to support optional chaining

ashchan:


ashchan:
Oops, forgot we needed `esnext` for `BigInt` literals.

PR #1230, opened 2019-12-17
Add new error message for live capacities not enough

classicalliu:


packages/neuron-wallet/src/locales/en.ts
```diff
@@ -88,6 +88,7 @@ export default {
       'unsupported-cipher': 'Unsupported cipher.',
       'capacity-not-enough': 'Insufficient balance.',
       'capacity-not-enough-for-change': "You need more capacities for change (at least 61 CKBytes), or click 'Max' button to send all your balance.",
+      'live-capacity-not-enough': 'Unavailable balance found, please try again when last transaction has been confirmed.',
```
Keith-CY: I think `Unavailable balance found` could be changed to `Insufficient available balance`

classicalliu: @CipherWang

classicalliu: updated

PR #1236, opened 2019-12-18
Move tx-related parts from wallet service to tx wallet service

classicalliu:


packages/neuron-wallet/src/services/tx-wallet.ts
```diff
@@ -0,0 +1,460 @@
...
+import { CellIsNotYetLive, TransactionIsNotCommittedYet } from 'exceptions/dao'
+import TypeConvert from 'types/type-convert'
+import FeeMode from 'models/fee-mode'
+import DaoUtils from 'models/dao-utils'
+import TransactionSize from 'models/transaction-size'
+import TransactionFee from 'models/transaction-fee'
+import logger from 'utils/logger'
+import Keychain from 'models/keys/keychain'
+
+export default class TxWalletService {
```
ashchan: `TxWallet` is confusing. Please consider a better name such as `TransactionService`.

classicalliu: A class already named `TransactionService`, how about `TransactionSender` ? It's a class combine transaction and wallet (to sign transactions).

ashchan: `TransactionSender` looks good to me 👍

classicalliu: updated

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -0,0 +1,460 @@
...
+import TransactionFee from 'models/transaction-fee'
+import logger from 'utils/logger'
+import Keychain from 'models/keys/keychain'
+
+export default class TransactionSender {
+  private static instance: TransactionSender
+
+  private walletService: WalletService
+
+  public static getInstance = () => {
```
ashchan: I feel singleton is unnecessary for this class.

classicalliu: Removed

PR #1240, opened 2019-12-19
feat: Refine the layout of sidebar

Keith-CY:


packages/neuron-ui/src/styles/layout.scss
```diff
@@ -1,5 +1,5 @@
 :root{
-  --sidebar-width: 150px;
+  --sidebar-width: 158px;
```
ashchan: What's the sidebar width from the design? A bit wider?

packages/neuron-ui/src/locales/en.json
```diff
@@ -11,7 +11,9 @@
       "history": "History",
       "addresses": "Addresses",
       "settings": "Settings",
-      "nervos-dao": "Nervos DAO"
+      "nervos-dao": "Nervos DAO",
+      "fail-to-fetch-tip-block-number": "Connection error",
+      "sync-not-start": "Synchronization not started yet"
```
ashchan: Will this fit as one line within the sidebar?

packages/neuron-ui/src/styles/layout.scss
Keith-CY: > What's the sidebar width from the design? A bit wider?

It's from the last design in zeplin.

packages/neuron-ui/src/locales/en.json
Keith-CY: > Will this fit as one line within the sidebar?

This message is still wrapped in the sidebar

ashchan: I see. Let's make it shorter.

ashchan:
```suggestion
      "sync-not-start": "Sync not started yet"
```

PR #1241, opened 2019-12-20
Generate receiving/change addresses alone

classicalliu:


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -17,6 +17,9 @@ export interface AddressMetaInfo {
 }
 
 export default class AddressService {
+  // <= 3
+  private static generateCheckDividingLine: number = 3
```
ashchan: Threashold for `Dividing line`?

classicalliu: Means if available addresses count <= 3, then generate new addresses.

classicalliu: renamed

PR #1243, opened 2019-12-20
Update current network info (chain, genesis hash) when connected

ashchan:
Whenever connecting to a new network or current one, it should update genesis info and chain type.

### DIFF
--- a/packages/neuron-wallet/src/main.ts
+++ b/packages/neuron-wallet/src/main.ts
@@ -7,7 +7,7 @@ import { changeLanguage } from 'utils/i18n'
 import env from 'env'
 import { register as registerListeners } from 'listeners/main'
 
-const appController = new AppController()
+let appController = new AppController()
 
 app.on('ready', async () => {
   changeLanguage(app.getLocale())
@@ -20,7 +20,7 @@ app.on('ready', async () => {
 
   WalletService.getInstance().generateAddressesIfNecessary()
 
-  appController.openWindow()
+  appController.start()
 })
 
 app.on('before-quit', async () => {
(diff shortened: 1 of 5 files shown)


packages/neuron-wallet/src/main.ts
```diff
@@ -7,7 +7,7 @@ import { changeLanguage } from 'utils/i18n'
 import env from 'env'
 import { register as registerListeners } from 'listeners/main'
 
-const appController = new AppController()
+let appController = new AppController()
```
classicalliu: What's the purpose here ?

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -100,24 +91,7 @@ export default class NetworksService extends Store {
...
-      }
-
-      const genesisHash = await core.rpc
-        .getBlockHash('0x0')
-        .catch(() => EMPTY_GENESIS_HASH)
-      if (genesisHash !== EMPTY_GENESIS_HASH) {
-        network.genesisHash = genesisHash
-      }
-    }
+    Object.assign(await this.refreshChainInfo(network))
```
classicalliu: Only one argument here ?

ashchan: Oops

packages/neuron-wallet/src/main.ts
ashchan: fixed.

packages/neuron-wallet/src/services/networks.ts
ashchan: Updated.

PR #1260, opened 2019-12-24
feat: Update the UI of transaction detail

Keith-CY:
image


packages/neuron-wallet/src/controllers/app/show-window.ts
```diff
@@ -4,7 +4,9 @@ import env from 'env'
 
 const showWindow = (url: string, title: string): BrowserWindow => {
 	const win = new BrowserWindow({
-		width: 1200,
+    width: 1200,
```
classicalliu: Indent ?

Keith-CY: > Indent ?

Replaced tabs with spaces in https://github.com/nervosnetwork/neuron/pull/1260/commits/664bde7d32aaf2a59c32143db279e2f6858b8f08

PR #1298, opened 2020-01-06
Build classes for chain types

classicalliu:


ashchan:
This is huge 🍻

packages/neuron-wallet/src/block-sync-renderer/sync/check-and-save/tx.ts
```diff
@@ -50,20 +53,19 @@ export default class CheckTx {
   }
 
   public filterOutputs = (lockHashes: string[]) => {
-    const cells: Cell[] = this.tx.outputs!.map((output, index) => {
+    const cells: Output[] = this.tx.outputs!.map((output, index) => {
```
ashchan: Looking at `filterOutputs` and `filterInputs`, a reader would expect them to return `Output`s and `Inputs`s. 

Question one: is it possible to rename so we know exactly what are these methods doing?
Question two: `filterOuputs` returns outputs, then why the var is called `cells`, not `outputs` (or a generic name as `results`)?

packages/neuron-wallet/src/models/chain/block-header.ts
```diff
@@ -0,0 +1,78 @@
+import TypeChecker from "utils/type-checker"
+
+export interface BlockHeaderInterface {
```
ashchan: I see no reason to have a `XxxInterface` layer.

My suggestion:

* Remove all these `Interfaces`.
* Expose class properties (read & write) by default.
* Comment or mark any properties that should have a setter to warn direct assignment.

packages/neuron-wallet/src/block-sync-renderer/sync/check-and-save/tx.ts
classicalliu: 1. How about `selectInputsOfMine` / `selectOutputsOfMine` ?
2. I'll change them to `outputs`

packages/neuron-wallet/src/models/chain/transaction-point.ts
```diff
@@ -0,0 +1,23 @@
+import TypeChecker from "utils/type-checker"
+
+export default class TransactionPoint {
+  constructor(
+    public blockNumber: string,
+    public txHash: string,
+    public index: string
+  ) {
+    this.blockNumber = BigInt(blockNumber).toString()
```
ashchan: `blockNumber` and `index` are modified when assigned, thus constructor assignment shorthand is not shorthand anymore. How about declaring them as public instance properties explicitly?

packages/neuron-wallet/src/models/chain/script.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default class Script {
+  constructor(
+    public codeHash: string,
+    public args: string,
+    public hashType: ScriptHashType
+  ) {
+    this.args = args
+    this.codeHash = codeHash
+    this.hashType = hashType
```
ashchan: See blow. It looks more clearer to explicitly declare properties when we're not using constructor assignment shorthand (empty `{}`).

packages/neuron-wallet/src/block-sync-renderer/sync/check-and-save/tx.ts
```diff
@@ -49,25 +52,24 @@ export default class CheckTx {
...
       }
       return false
-    }).filter(cell => !!cell) as Cell[]
-    return cells
+    }).filter(output => !!output) as Output[]
+    return outputs
   }
 
-  public filterInputs = async (lockHashes: string[]): Promise<string[]> => {
+  public selectInputsOfWallets = async (lockHashes: string[]): Promise<string[]> => {
```
ashchan: Does it actually select addresses instead of inputs? If so it should be called `selectAddressesOfWallets`.

packages/neuron-wallet/src/controllers/dao.ts
```diff
@@ -1,13 +1,15 @@
...
 import CellsService from 'services/cells'
 import LockUtils from 'models/lock-utils'
 import TransactionSender from 'services/transaction-sender'
+import OutPoint from 'models/chain/out-point'
+import Output from 'models/chain/output'
+import Transaction from 'models/chain/transaction'
 
 export default class DaoController {
-  public async getDaoCells(params: Controller.Params.GetDaoCellsParams): Promise<Controller.Response<Cell[]>> {
+  public async getDaoCells(params: Controller.Params.GetDaoCellsParams): Promise<Controller.Response<Output[]>> {
```
ashchan: Get cells or outputs?

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -42,7 +44,7 @@ export default class CellsService {
 
   public static getDaoCells = async (
     lockHashes: string[],
-  ): Promise<Cell[]> => {
+  ): Promise<Output[]> => {
```
ashchan: Please change name accordingly now that it has a different return type.

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -71,23 +73,23 @@ export default class CellsService {
...
+          output.setDepositTimestamp(tx.timestamp)
         }
       }
     }
 
     return cells
   }
 
-  public static getLiveCell = async (outPoint: OutPoint): Promise<Cell | undefined> => {
+  public static getLiveCell = async (outPoint: OutPoint): Promise<Output | undefined> => {
```
ashchan: As above, please change name accordingly now that it has a different return type.

packages/neuron-wallet/src/models/chain/witness-args.ts
```diff
@@ -0,0 +1,45 @@
+export default class WitnessArgs {
+  public static EMPTY_LOCK = '0x' + '0'.repeat(130)
+
+  constructor(
+    public lock?: string,
+    public inputType?: string,
+    public outputType?: string
+  ) {
+    this.outputType = outputType
```
ashchan: Why do we need to assign this property manually?

packages/neuron-wallet/src/models/chain/transaction-point.ts
classicalliu: OK，you means 
```
public blockNumber: string

constructor(blockNumber: string)
```
 ?

packages/neuron-wallet/src/models/chain/script.ts
classicalliu: I forgot to delete those lines.

packages/neuron-wallet/src/block-sync-renderer/sync/check-and-save/tx.ts
classicalliu: Maybe `selectInputAddressesOfWallets` ?

packages/neuron-wallet/src/controllers/dao.ts
classicalliu: Cell is alias to outputs, and CKB also using them mixed.

packages/neuron-wallet/src/services/cells.ts
classicalliu: `import Cell from xxx` ?

packages/neuron-wallet/src/models/chain/witness-args.ts
classicalliu: I forgot to delete this line too.

packages/neuron-wallet/src/block-sync-renderer/sync/check-and-save/tx.ts
ashchan: That sounds good to me.

packages/neuron-wallet/src/controllers/dao.ts
ashchan: Huh, CKB using them mixed doesn't mean we have to follow that misleading style 😹

packages/neuron-wallet/src/models/chain/transaction-point.ts
classicalliu: All shorthand assignments are deleted

packages/neuron-wallet/src/block-sync-renderer/sync/check-and-save/tx.ts
classicalliu: Renamed

packages/neuron-wallet/src/controllers/dao.ts
classicalliu: Rename to `Cell` in this file.

packages/neuron-wallet/src/services/cells.ts
classicalliu: Renamed to `Cell` in this file.

PR #1312, opened 2020-01-10
refactor: inject the state into sub componnets of the main component …

Keith-CY:
…and remove the state from hoc

### DIFF
--- a/packages/neuron-ui/src/stories/Overview.stories.tsx
+++ b/packages/neuron-ui/src/stories/Overview.stories.tsx
@@ -1,14 +1,14 @@
 import React from 'react'
-import { Route } from 'react-router-dom'
 import { storiesOf } from '@storybook/react'
 import { withKnobs, text, boolean } from '@storybook/addon-knobs'
 import StoryRouter from 'storybook-react-router'
 import { action } from '@storybook/addon-actions'
 import Overview from 'components/Overview'
 import initStates from 'states/initStates'
-import { StateWithDispatch } from 'states/stateProvider/reducer'
 import transactions from './data/transactions'
 import addresses from './data/addresses'
+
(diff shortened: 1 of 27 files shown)


packages/neuron-ui/src/stories/Overview.stories.tsx
```diff
@@ -61,14 +61,11 @@ const states = {
...
 
-const OverviewWithRouteProps = (props: StateWithDispatch) => (
-  <Route path="/" render={routeProps => <Overview {...routeProps} {...props} />} />
-)
-
 const stories = storiesOf(`Overview`, module).addDecorator(StoryRouter())
 
 Object.entries(states).forEach(([title, props]) => {
-  stories.add(title, () => <OverviewWithRouteProps {...props} />)
+  console.info(props)
```
classicalliu: Is this log expected ?

packages/neuron-ui/src/stories/Overview.stories.tsx
```diff
@@ -120,5 +117,6 @@ stories.addDecorator(withKnobs).add('With knobs', () => {
       ],
     },
   }
-  return <OverviewWithRouteProps {...props} />
+  console.info(props)
```
classicalliu: Also here.

packages/neuron-ui/src/stories/Overview.stories.tsx
Keith-CY: Yes, it's used to print the expected state of the component in the storybook. Due to the `useContext` hook used in these components, the states cannot be observed directly.

classicalliu: ok

Keith-CY: BTW, all the codes in `/stories` are just for developing and testing.

PR #1313, opened 2020-01-10
Fix import wallet syncing in default network

classicalliu:


ashchan:
We might want to allow connecting to different chains even from the default network.

PR #1322, opened 2020-01-14
Fix daoCellOutPoints sql too long error

classicalliu:


packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -179,15 +179,14 @@ export class TransactionsService {
       .map(i => i.outPointTxHash)
       .filter(h => !!h) as string[]
 
-    const daoCellOutPoints: { txHash: string, index: string }[] = await getConnection()
+    const daoCellOutPoints: { txHash: string, index: string }[] = (await getConnection()
```
Keith-CY: Using promise or separate the filter out makes it more readable.

PR #1325, opened 2020-01-14
Fix multiple keystore file dialog bug

ashchan:


packages/neuron-ui/src/services/remote/index.ts
```diff
@@ -40,6 +40,14 @@ export const showOpenDialog = (options: { title: string; message?: string }) =>
   return window.remote.require('electron').dialog.showOpenDialog(options)
 }
 
+export const showOpenDialogModal = (options: { title: string; message?: string }) => {
```
Keith-CY: What's the difference between `showOpenDialogModal` and `showOpenDialog`?

ashchan: The modal override passes in the current window. On macOS open file dialog usually appears as a modal sub window.

PR #1326, opened 2020-01-15
Move synced block number access out of sync queue

ashchan:
* Block number DB write should happen in main process (TODO: update indexer queue)
* Add Sync API controller to handle ipc invoke from sync renderer process


packages/neuron-wallet/src/block-sync-renderer/index.ts
```diff
@@ -38,11 +38,13 @@ const syncNetwork = async (rescan = false) => {
   LockUtils.cleanInfo()
   DaoUtils.cleanInfo()
 
-  const blockNumber = new BlockNumber()
+  const blockNumber = new SyncedBlockNumber()
   if (rescan) {
     await blockNumber.reset()
   }
-  SyncedBlockNumberSubject.getSubject().next((await blockNumber.getCurrent()).toString())
+  const startBlockNumber = (await blockNumber.getNextBlock()).toString()
```
classicalliu: Does `getNextNumber` or `getNextBlockNumber` a better name ?

ashchan: Not a big problem to me, as class name `SyncedBlockNumber` already tells this is `block number` related.

classicalliu: ok

PR #1329, opened 2020-01-16
Define address property from useRouteMatch params

ashchan:


packages/neuron-ui/src/components/Receive/index.tsx
```diff
@@ -15,14 +15,13 @@ const Receive = () => {
...
+  } = useRouteMatch()
 
   const accountAddress = useMemo(
-    () =>
-      params.address ||
-      (addresses.find(addr => addr.type === 0 && addr.txCount === 0) || { address: '' }).address ||
-      '',
-    [params, addresses]
+    () => address || (addresses.find(addr => addr.type === 0 && addr.txCount === 0) || { address: '' }).address || '',
+    [addresses]
```
Keith-CY: The `address` is missing in the hook dependencies.

PR #1346, opened 2020-01-29
Indexer sync debug and fix

ashchan:


ashchan:
Closing this as we're going to remove indexer module support.

PR #1348, opened 2020-01-31
Remove indexer module

ashchan:
Current indexer module of CKB doesn't offer flexible indexing feature, we're removing it and stick with block (scan) sync.


ashchan:
> I think can also check and delete `services/tx/indexer-transaction.ts`, and indexer part of `rpc-service` and `models/chain`

Good catch.

PR #1352, opened 2020-02-04
Feature for importing extended public key

ashchan:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -209,14 +209,52 @@ export default class WalletsController {
     })
   }
 
+  public async importXPubkey(): Promise<Controller.Response<Wallet>> {
+    return new Promise((resolve, reject) => {
```
Keith-CY: The `Promise` is redundant.

packages/neuron-wallet/src/controllers/app/menu.ts
```diff
@@ -135,8 +135,18 @@ const updateApplicationMenu = (mainWindow: BrowserWindow | null) => {
           {
             id: 'import-with-keystore',
             label: i18n.t('application-menu.wallet.import-keystore'),
-            click: () => { navigateTo(URL.ImportKeystore )},
+            click: () => { navigateTo(URL.ImportKeystore )}
```
classicalliu:
```suggestion
            click: () => { navigateTo(URL.ImportKeystore) }
```

PR #1357, opened 2020-02-05
Support full address format when send transaction

classicalliu:


packages/neuron-wallet/src/services/tx/transaction-generator.ts
```diff
@@ -124,21 +127,24 @@ export class TransactionGenerator {
...
 
-      // skip last output
-      if (BigInt(capacity) < minCellCapacity && index !== targetOutputs.length - 1) {
-        throw new CapacityTooSmall()
+      let lockScript: Script | undefined
+      if (FullAddress.isFullFormat(address)) {
+        lockScript = FullAddress.parse(address)
+      } else {
+        const blake160: string = LockUtils.addressToBlake160(address)
+        lockScript = new Script(codeHash, blake160, hashType)
```
ashchan: Can we extract this to something as `parseAddressToScript`?

classicalliu: ok

classicalliu: Updated, using `AddressParser` to parse all addresses types

PR #1367, opened 2020-02-10
feat: display full address in tx detail

Keith-CY:


packages/neuron-ui/src/components/Transaction/index.tsx
```diff
@@ -173,22 +171,30 @@ const Transaction = () => {
           address = t('transaction.cell-from-cellbase')
         } else {
           try {
-            address = ckbCore.utils.bech32Address(cell.lock.args, {
-              prefix: addressPrefix,
-              type: ckbCore.utils.AddressType.HashIdx,
-              codeHashOrCodeHashIndex: '0x00',
-            })
+            if (cell.lock.codeHash === systemCodeHash) {
```
classicalliu: Also need `hashType` match

Keith-CY: updated in https://github.com/nervosnetwork/neuron/pull/1367/commits/3140d092ea837714a710a9baf246ee5dcd1e10b2

PR #1381, opened 2020-02-19
Sync single multi sign transactions

classicalliu:
1. sync single multi sign transactions
2. generate transaction with single multi sign inputs
3. sign transaction with single multi sign input and send

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
@@ -35,13 +35,18 @@ export default class Queue {
 
   private yieldTime = 1
 
-  constructor(url: string, lockHashes: string[], startBlockNumber: bigint) {
+  private multiSignCodeHash: string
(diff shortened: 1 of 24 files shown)


packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -35,13 +35,18 @@ export default class Queue {
 
   private yieldTime = 1
 
-  constructor(url: string, lockHashes: string[], startBlockNumber: bigint) {
+  private multiSignCodeHash: string
+  private multiSignBlake160s: string[]
+
+  constructor(url: string, lockHashes: string[], startBlockNumber: bigint, multiSignCodeHash: string, multiSignBlake160s: string[]) {
```
ashchan: Params order matters (to group things together). Could we move `startBlockNumber` to the end, or before `lockHashes`?

classicalliu: Moved to the end

PR #1383, opened 2020-02-20
Fix an issue that SyncInfoEntry would write too frequently

ashchan:


packages/neuron-wallet/src/models/synced-block-number.ts
```diff
@@ -4,64 +4,54 @@ import SyncedBlockNumberSubject from "models/subjects/node"
...
-      return
-    }
+      this.#lastSavedBlock = current
 
-    if (!blockNumberEntity) {
-      blockNumberEntity = new SyncInfoEntity()
-      blockNumberEntity.name = SyncInfoEntity.CURRENT_BLOCK_NUMBER
+      let blockNumberEntity = await this.blockNumber()
+      blockNumberEntity.value = current.toString()
+      getConnection().manager.save(blockNumberEntity)
```
classicalliu: `await` ?

ashchan: This is by purpose. It doesn't need to wait for the writing to be finished.

Not sure if this would cause any side effect, though.

classicalliu: OK

PR #1385, opened 2020-02-20
feat: Add the component of Datetime Picker

Keith-CY:
image

### DIFF
--- a/packages/neuron-ui/src/locales/en.json
+++ b/packages/neuron-ui/src/locales/en.json
@@ -392,7 +392,7 @@
       "mon": { "full": "Monday", "short": "Mon.", "tag": "M" },
       "tue": { "full": "Tuesday", "short": "Tue.", "tag": "T" },
       "wed": { "full": "Wednesday", "short": "Wed.", "tag": "W" },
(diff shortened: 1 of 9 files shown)


packages/neuron-ui/src/locales/en.json
```diff
@@ -387,6 +387,27 @@
       "download-update": "Install Update",
       "updates-downloaded-about-to-quit-and-install": "Update downloaded. Ready to install and relaunch.",
       "quit-and-install": "Install and relaunch"
+    },
+    "datetime": {
+      "mon": { "full": "Monday", "short": "Mon.", "tag": "M" },
+      "tue": { "full": "Tuesday", "short": "Tue.", "tag": "T" },
+      "wed": { "full": "Wednesday", "short": "Wed.", "tag": "W" },
+      "thus": { "full": "Thusday", "short": "Thus.", "tag": "T" },
```
ashchan: I'm afraid `thus` and `Thusday` are typos. Was `thus` meant to be `thur`?

Keith-CY: Fixed in https://github.com/nervosnetwork/neuron/pull/1385/commits/8f7f595394c6e32c2c0c5232e433c36cdc185397

PR #1386, opened 2020-02-20
Sign and verify message

classicalliu:


packages/neuron-wallet/src/controllers/sign-and-verify.ts
```diff
@@ -0,0 +1,38 @@
+import SignAndVerify from "services/sign-and-verify"
+import { ServiceHasNoResponse } from "exceptions"
+import { ResponseCode } from "utils/const"
+
+export default class SignAndVerifyController {
```
ashchan: Would `SignMessageController` be a bit better than `SignAndVerifyController`.

PR #1395, opened 2020-02-21
feat: allow whitespaces in password

Keith-CY:


packages/neuron-ui/src/components/PasswordRequest/index.tsx
```diff
@@ -68,12 +68,9 @@ const PasswordRequest = () => {
   const onChange = useCallback(
     (e: React.SyntheticEvent<HTMLInputElement>) => {
       const { value } = e.target as HTMLInputElement
-      if (/\s/.test(value)) {
-        return
-      }
       dispatch({
         type: AppActions.UpdatePassword,
-        payload: value,
+        payload: value.trim(),
```
ashchan: If we allow space, while disallow them at the beginning or end of the password?

Keith-CY: Allow password to start and end with whitespaces in https://github.com/nervosnetwork/neuron/pull/1395/commits/f5e511c0164c67d4961a2f6b18d3794084205afb

PR #1432, opened 2020-03-03
[ᚬrc/v0.29.0] Fix estimated release time

Keith-CY:
Use the timestamp of local tip block instead of that of system to calculate the estimation time.


Keith-CY:
It's been merged into the develop branch, but now we are going to release from the rc branch.

PR #1436, opened 2020-03-03
[ᚬsUDT] feat: add the dialog of creating sUDT account

Keith-CY:
Add the create sudt account dialog
image
image
image


Keith-CY:
This component is under draft and used in storybook only.

PR #1437, opened 2020-03-03
[ᚬrc/v0.29.0] Update multi sign blake160 if transaction exists when import

classicalliu:


packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -47,6 +48,60 @@ export class TransactionPersistor {
...
       .findOne(transaction.hash)
 
+    // update multiSignBlake160
+    if (txEntity) {
+      const hasMultiSignBlake160Output: boolean = !!transaction.outputs.find(o => !!o.multiSignBlake160)
+      const hasMultiSignBlake160Input: boolean = !!transaction.inputs.find(i => !!i.multiSignBlake160)
+      if (hasMultiSignBlake160Output || hasMultiSignBlake160Input) {
+        // update multiSignBlake160Info
+        const outputsToUpdate: Output[] = hasMultiSignBlake160Output ? transaction.outputs.filter(o => !!o.multiSignBlake160) : []
+        const inputsToUpdate: Input[] = hasMultiSignBlake160Input ? transaction.inputs.filter(i => !!i.multiSignBlake160) : []
```
Keith-CY: The ternary is redundant cuz `filter` returns `[]` if `hasMultiSignBlake160Input` is false

classicalliu: Yes, but it can be faster, no need to loop once if we know the result is empty set.

Or we can remove the `find`, only use filter, may be faster.

Keith-CY: > Yes, but it can be faster, no need to loop once if we know the result is empty set.
> 
> Or we can remove the `find`, only use filter, may be faster.

I prefer to use `filter` directly and remove `find`

classicalliu: updated

PR #1460, opened 2020-03-12
[ᚬrefactor-nervos-dao] feat: add the compensation tooltip component

Keith-CY:
image
image
image
image
image
image
image
image


ashchan:
What is the `blah` text placeholder for? Is that production ready?

PR #1463, opened 2020-03-13
Add copy component

Keith-CY:
image

The overlay with text `copy` appears on mouse hover event to indicate the text is copiable.

### DIFF
--- a/packages/neuron-ui/src/widgets/CopyZone/copyZone.module.scss
+++ b/packages/neuron-ui/src/widgets/CopyZone/copyZone.module.scss
@@ -1,0 +1,25 @@
+.container {
+  position: relative;
+  display: inline-block;
+
+  &::before {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    display: none;
+    justify-content: center;
+    align-items: center;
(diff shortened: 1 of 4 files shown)


packages/neuron-ui/src/widgets/CopyZone/copyZone.module.scss
```diff
@@ -0,0 +1,25 @@
...
+    content: attr(data-prompt);
+    font-size: 0.875rem;
+    color: #666;
+    background-color: rgba(255, 255, 255, 0.7);
+    backdrop-filter: blur(1px);
+    user-select: none;
+  }
+
+  &:hover::before {
+    display: flex;
```
katat, line 23: To make the UI interaction intuitive enough that it is clickable here when hovering, maybe a cursor is helpful.
Such as: `cursor: pointer`

Keith-CY: > To make the UI interaction intuitive enough that it is clickable here when hovering, maybe a cursor is helpful.
> Such as: `cursor: pointer`

We are avoiding using `pointer` cuz it doesn't conform to the native style.

PR #1465, opened 2020-03-16
Randomize tx outputs

ashchan:


packages/neuron-wallet/src/services/tx/transaction-generator.ts
```diff
@@ -109,6 +110,8 @@ export class TransactionGenerator {
     tx.addOutput(output)
   }
 
+  tx.outputs = ArrayUtils.shuffle(tx.outputs)
```
classicalliu: Please reset `outputsData`

PR #1470, opened 2020-03-18
feat: conceal real node address if it's not localhost for privacy

Keith-CY:
URL not starts with `http://localhost` or `https://localhost` will be replaced with `http://****:port` in the status.json


Keith-CY:
> It should treat `127.0.0.1` as local too.

Updated in https://github.com/nervosnetwork/neuron/pull/1470/commits/533ad14394bb1b5fdeac4462ee7d2154045f0b6a

PR #1473, opened 2020-03-19
Refactor balance calculation

classicalliu:


packages/neuron-wallet/src/database/address/address-dao.ts
```diff
@@ -52,34 +51,48 @@ export default class AddressDao {
   // pendingBalance means balance of OutputStatus.Pending cells (sent from me, but not committed)
   // so the final balance is (liveBalance + sentBalance - pendingBalance)
   // balance is the balance of the cells those who don't hold data or type script
-  public static updateTxCountAndBalances = async (
-    addresses: string[]
-  ): Promise<Address[]> => {
+  public static async updateTxCountAndBalances(addresses: Set<string>): Promise<Address[]> {
```
ashchan: Perhaps the old interface (`Address[]`) is more generic and understandable. Proposal: using `Set` as internal collection implementation detail for performance reason, expose the method accepting array.

classicalliu: Updated

PR #1492, opened 2020-04-01
feat: add 10s buffer to display connection error for launching ckb node

Keith-CY:
ConnectionStatus.Connecting will be displayed in the first 10s since app launchs while the sync in is under error.

This status is for better UX during launching the CKB node.


Keith-CY:
> It still shows 'Sync failed, please check network' on launch with bundled node.

Updated by using `ConnectionStatus.Connecting` instead of `SyncStatus.Connecting` so that `ConnectionStatus.Offline` will be overridden by `ConnectionStatus.Connecting`, which avoiding `Connection Error` in the first 10s.

PR #1493, opened 2020-04-01
Skip rescan importing existing wallet

ashchan:


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -37,26 +37,29 @@ export default class AddressService {
...
-        address.isImporting = isImporting
-        return address
-      })
-    AddressCreatedSubject.getSubject().next(addrs)
+    const versionFilter = ((a: AddressInterface) => { return a.version === AddressService.getAddressVersion() })
+
+    // If first receiving address already exists in other wallets, treat as none importing.
+    // This assumes addresses.first is the actual first address.
+    const firstAddress = addresses.filter(versionFilter)[0]
+    const alreadyExist = AddressDao.findByAddresses([firstAddress.address]).length > 1
```
classicalliu, line 56: Why is `length > 1` but not `> 0` ?

ashchan: The new created address is also in the db store already.

classicalliu: ok

PR #1495, opened 2020-04-02
Refactor nervos dao

Keith-CY:


Keith-CY:
Hold on until this feature is accepted in the next release.

Keith-CY:
@kellyshang The fonts have been changed according to the new UI design, but this PR only covers those in the Nervos DAO.

Another PR to update fonts globally will be submitted later.

PR #1498, opened 2020-04-03
Set a timeout of 2 seconds when updating a network

ashchan:
fetching its genesis block hash and chain info

### DIFF
--- a/packages/neuron-wallet/src/services/networks.ts
+++ b/packages/neuron-wallet/src/services/networks.ts
@@ -149,21 +149,31 @@ export default class NetworksService extends Store {
 
   // Refresh a network's genesis and chain info
   private async refreshChainInfo(network: Network): Promise<Network> {
-    const ckb = new CKB(network.remote)
+    return new Promise((resolve, _) => {
+      const fail = () => {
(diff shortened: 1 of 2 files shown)


ashchan:
Hold as it has caused test failure on CI

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -149,21 +149,31 @@ export default class NetworksService extends Store {
...
+      ]).then(([genesisHash, chain]) => {
+        network.genesisHash = genesisHash
+        network.chain = chain
+        resolve(network)
+      }).catch(() => {
+        fail()
+      })
+
+      // If network.remote is reachable but doesn't respond cancel it early.
+      setTimeout(fail, 2000)
```
Keith-CY, line 176: The timeout can be implemented simply by
```
const http = require('http')
const httpAgent = new http.Agent({ keepAlive: true, timeout: 2000 })
ckb.rpc.setNode({ httpAgent })
```

ashchan: Are you 100% sure?

PR #1503, opened 2020-04-07
[ᚬrefactor-nervos-dao] fix: display -- instead of NaN

Keith-CY:


Keith-CY:
@kellyshang forced pushed, please re-pull this branch

PR #1506, opened 2020-04-08
chore: upgrade react-scripts to 3.4.1

Keith-CY:


Keith-CY:
@classicalliu react-scripts has been moved into dev deps.

PR #1512, opened 2020-04-09
[ᚬsUDT] Add AssetAccountInfo

classicalliu:


packages/neuron-wallet/src/models/asset-account-info.ts
```diff
@@ -0,0 +1,90 @@
...
+    // const mainnetAnyoneCanPayCellInfo: ScriptCellInfo = {
+    //   cellDep: new CellDep(new OutPoint('', '0'), DepType.DepGroup),
+    //   codeHash: '',
+    //   hashType: ScriptHashType.Type
+    // }
+
+    // this.sudtInfo.set(mainnetGenesisBlockHash, mainnetSudtCellInfo)
+    // this.anyoneCanPayInfo.set(mainnetGenesisBlockHash, mainnetAnyoneCanPayCellInfo)
+
+    const testnetGenesisBlockHash = '0x63547ecf6fc22d1325980c524b268b4a044d49cda3efbd584c0a8c8b9faaf9e1'
```
Keith-CY, line 41: The hard coded hash and script could be moved into `const.ts`

packages/neuron-wallet/src/models/asset-account-info.ts
```diff
@@ -0,0 +1,90 @@
...
+    const testnetAnyoneCanPayCellInfo: ScriptCellInfo = {
+      cellDep: new CellDep(new OutPoint('0x9af66408df4703763acb10871365e4a21f2c3d3bdc06b0ae634a3ad9f18a6525', '0'), DepType.DepGroup),
+      codeHash: '0x6a3982f9d018be7e7228f9e0b765f28ceff6d36e634490856d2b186acf78e79b',
+      hashType: ScriptHashType.Type
+    }
+    this.sudtInfo.set(testnetGenesisBlockHash, testnetSudtCellInfo)
+    this.anyoneCanPayInfo.set(testnetGenesisBlockHash, testnetAnyoneCanPayCellInfo)
+  }
+
+  public getSudtCellDep(): CellDep {
```
Keith-CY, line 56: Would it be better to use `getter` instead of `instance method`?

packages/neuron-wallet/src/models/asset-account-info.ts
```diff
@@ -0,0 +1,90 @@
...
+
+interface ScriptCellInfo {
+  cellDep: CellDep
+  codeHash: string
+  hashType: ScriptHashType
+}
+
+export default class AssetAccountInfo {
+  // <genesisBlockHash, sudtScriptCellInfo>
+  private sudtInfo: Map<string, ScriptCellInfo>
```
Keith-CY, line 14: Now we can use private fields of JavaScript

packages/neuron-wallet/src/models/asset-account-info.ts
classicalliu: I think it should follow its logic.

packages/neuron-wallet/src/models/asset-account-info.ts
classicalliu: updated

packages/neuron-wallet/src/models/asset-account-info.ts
classicalliu: I see, but it's not required.

PR #1516, opened 2020-04-13
[ᚬsUDT] Create asset account

classicalliu:


packages/neuron-wallet/src/controllers/asset-account.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface SendCreateAssetAccountTxParams {
+  walletID: string
+  assetAccount: AssetAccount
+  tx: Transaction
+  password: string
+}
+
+export interface UpdateAssetAccountParams {
+  id: number
+  walletID?: string
```
Keith-CY, line 30: The wallet id and token id cannot be changed.

classicalliu: updated

PR #1537, opened 2020-04-18
[ᚬsUDT] feat: assemble the sudt account list, send and receive

Keith-CY:
Assemble the s-udt account list, s-udt send, s-udt receive components into the app.

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
@@ -55,45 +55,13 @@ export default class Queue {
       }
     })
 
-    // while (!this.stopped) {
-    //   try {
-    //     this.inProcess = true
-
-    //     if (this.lockHashes.length !== 0) {
-    //       const startNumber = this.currentBlockNumber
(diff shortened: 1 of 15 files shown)


packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -55,13 +55,45 @@ export default class Queue {
...
+    //       logger.debug(`Sync:\terror:`, err)
+    //     } else {
+    //       logger.error(`Sync:\terror:`, err)
+    //     }
+    //   } finally {
+    //     await CommonUtils.sleep(this.yieldTime)
+    //     this.inProcess = false
+    //   }
+    // }
+    setTimeout(async()=>{
```
classicalliu, line 89: ?

Keith-CY: > ?

My fault, it's a workaround for the sync problem locally.

Keith-CY: It's been removed

PR #1551, opened 2020-04-22
feat: add sudt amount formatters

Keith-CY:


packages/neuron-ui/src/utils/formatters.ts
```diff
@@ -213,6 +213,58 @@ export const difficultyFormatter = (value: bigint) => {
   return `${localNumberFormatter(value)} H`
 }
 
+export const sudtAmountToValue = (amount: string = '0', decimal: string = '0') => {
```
classicalliu, line 216: Can you add some tests for this function ?

Keith-CY: > Can you add some tests for this function ?

Tests are added

PR #1557, opened 2020-04-22
feat: use new icons to indicate the status of tx

Keith-CY:


Keith-CY:
> life is like a box of chocolate

PR #1559, opened 2020-04-23
Asset account balance

classicalliu:


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -280,9 +282,11 @@ export default class WalletService {
   }
 
   // TODO: move this method and generateTx/sendTx out of this file
-  public static async updateUsedAddresses(addresses: string[]) {
+  public static async updateUsedAddresses(addresses: string[], anyoneCanPayBlake160s: string[]) {
     const addrs = await AddressService.updateTxCountAndBalances(addresses)
-    const walletIds: Set<string> = new Set(addrs.map(addr => addr.walletId))
+    const addressVersion = NetworksService.getInstance().isMainnet() ? AddressVersion.Mainnet : AddressVersion.Testnet
+    const anyoneCanPayAddrs = await AddressService.updateUsedByAnyoneCanPayByBlake160s(anyoneCanPayBlake160s, addressVersion)
```
Keith-CY, line 288: The name `updateUsedByAnyoneCanPayByBlake160s` is weird because the param is `anyoneCanPayBlake160s`

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -77,6 +77,20 @@ export default class CellsService {
     }
   }
 
+  public static async usedByAnyoneCanPayBlake160s(anyoneCanPayLockHashes: string[], blake160s: string[]): Promise<string[]> {
+    // const anyoneCanPayLockHashBuffers = anyoneCanPayLockHashes.map(h => Buffer.from(h.slice(2), 'hex'))
```
Keith-CY, line 81: remove meaningless comment

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -40,13 +40,16 @@ export default class Queue {
 
   private assetAccountInfo: AssetAccountInfo | undefined
 
-  constructor(url: string, lockHashes: string[], multiSignBlake160s: string[], startBlockNumber: bigint) {
+  private anyoneCanPayLockHashes: string[]
+
+  constructor(url: string, lockHashes: string[], anyoneCanPayLockHashes: string[], multiSignBlake160s: string[], startBlockNumber: bigint) {
```
Keith-CY, line 45: There're too many params, consider using an object instead of a param list.

packages/neuron-wallet/src/services/wallets.ts
classicalliu: `blake160s` used by anyone-can-pay

packages/neuron-wallet/src/services/cells.ts

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
classicalliu: Seems no need, any obvious benefits ?

Keith-CY: > Seems no need, any obvious benefits ?

Not necessary, just a best practice that the params in the list should be up to 3

packages/neuron-wallet/src/services/wallets.ts
Keith-CY: > `blake160s` used by anyone-can-pay

I see, the name `updateUsedByAnyoneCanPayByBlake160s` is actually `update usedByAnyoneCanPay by blake160s` instead of `update usedBy(AnyoneCanPayByBlake160s)`

PR #1561, opened 2020-04-23
feat: update amount on sending all and the generated tx is valid

Keith-CY:


Keith-CY:
Which field should I use to display the amount when it is `generate-send-all-to-anyone-can-pay-tx`? Now I use the `tx.value` as a placement.
@classicalliu

classicalliu:
> Which field should I use to display the amount when it is `generate-send-all-to-anyone-can-pay-tx`? Now I use the `tx.value` as a placement.
> @classicalliu

Do you mean sUDT amount ?
How about add to `sudtInfo.amount` ?

Keith-CY:
> > Which field should I use to display the amount when it is `generate-send-all-to-anyone-can-pay-tx`? Now I use the `tx.value` as a placement.
> > @classicalliu
> 
> Do you mean sUDT amount ?
> How about add to `sudtInfo.amount` ?

That's ok

PR #1563, opened 2020-04-24
feat: add connecting status under the balance in the overview component

Keith-CY:


Keith-CY:
This is just a patch, the new icon will be updated in the next release

PR #1564, opened 2020-04-24
Fix send all anyone can pay

classicalliu:


Keith-CY:
CI failed

classicalliu:
> CI failed

Passed now

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -526,6 +526,87 @@ export default class CellsService {
...
+
+    if (anyoneCanPayLockLiveCellEntities.length === 0) {
+      throw new CapacityNotEnough()
+    }
+
+    const inputs: Input[] = []
+    const inputOriginCells: LiveCell[] = []
+    let inputCapacities: bigint = BigInt(0)
+    let totalSize: number = baseSize
+    anyoneCanPayLockLiveCells.every(cell => {
```
Keith-CY, line 566: The returned value is fixed to `true` and not used so `Array#forEach` is more proper than `Array.every`

classicalliu: ok

PR #1568, opened 2020-04-26
refactor: update the logic of starting bundled CKB node

Keith-CY:
Try to start the bundled CKB node when
1. network is specified
2. network address is the default one
3. fail to connect the external CKB node


Keith-CY:
@classicalliu the strategy has been updated according to the new requirement, please review again.

PR #1573, opened 2020-04-28
feat: add custom exceptions

Keith-CY:


packages/neuron-ui/src/exceptions/index.ts
```diff
@@ -0,0 +1,66 @@
+import { ErrorCode } from 'utils/const'
+
+export class FieldInvalidException extends Error {
+  public code = ErrorCode.FieldInvalid
+  public i18n: {
+    fieldName: string
+  }
+
+  constructor(fieldName: string) {
+    super(`messages.codes.${ErrorCode.FieldInvalid}`)
```
classicalliu, line 10:
```suggestion
    super(`messages.codes.${this.code}`)
```

Keith-CY: `this` should not be used until `super` is called.

classicalliu: ok

PR #1575, opened 2020-04-28
feat: display password error in password dialog

Keith-CY:
1. display the password error in the password dialog
2. refactor the action creator to remove component related logic outside


Keith-CY:
@kellyshang This PR prevents the `Password Dialog` from dismissing if the password is incorrect.

PR #1577, opened 2020-04-28
Fix generate anyone-can-pay CKB tx

classicalliu:


packages/neuron-wallet/src/services/anyone-can-pay.ts
```diff
@@ -19,7 +19,7 @@ export default class AnyoneCanPayService {
     fee: string = '0',
     description?: string,
   ): Promise<Transaction> {
-    const isCKB = tokenID === 'CKBytes' || tokenID === ''
+    const isCKB = !tokenID || tokenID === 'CKBytes' || tokenID === ''
```
Keith-CY, line 22: `!tokenID` covers `tokenID === ''`

classicalliu: fixed

PR #1585, opened 2020-04-29
fix: adjust address list layout for new font familty

Keith-CY:


Keith-CY:
@kellyshang Please test the new layout of the address list

PR #1597, opened 2020-05-06
feat: remove padding of address list

Keith-CY:


Keith-CY:
Please check if the `收款地址` is wrapped under the minimal size. @kellyshang

PR #1599, opened 2020-05-06
feat: update send-all amount from anyoneCanPaySendAmount

Keith-CY:


packages/neuron-ui/src/components/SUDTSend/index.tsx
```diff
@@ -209,13 +209,24 @@ const SUDTSend = () => {
...
+    sendState.description,
+    errors,
+    globalDispatch,
+    setRemoteError,
+    accountInfo,
+  ])
 
   useEffect(() => {
-    if (sendState.sendAll && experimental?.tx?.sudtInfo?.amount && accountInfo?.decimal) {
+    const amount = experimental?.tx?.sudtInfo?.amount || experimental?.tx?.anyoneCanPaySendAmount
```
classicalliu, line 225:
```suggestion
    const amount = experimental?.tx?.anyoneCanPaySendAmount
```

PR #1623, opened 2020-05-10
feat: add custom hook for validation

Keith-CY:
Add a custom hook to validate sUDT account info


katat:
Looks good to me 👍 
Not sure the current unit test coverage for the hooks. I think it would be worthwhile to have some unit tests to cover the rules inside the validators.

Keith-CY:
> Looks good to me 👍
> Not sure the current unit test coverage for the hooks. I think it would be worthwhile to have some unit tests to cover the rules inside the validators.

For now, UI components are all tested with the storybook

PR #1631, opened 2020-05-11
Apply copy component

Keith-CY:
image
image
image
image
image


Keith-CY:
@kellyshang Please check if there are any other texts that could be copied.

PR #1633, opened 2020-05-12
feat: export more logs for debug

Keith-CY:
Include the last 32kB of run.log from the bundled CKB

### DIFF
--- a/packages/neuron-wallet/src/controllers/export-debug.ts
+++ b/packages/neuron-wallet/src/controllers/export-debug.ts
@@ -95,16 +95,16 @@ export default class ExportDebugController {
 
     return new Promise((resolve, reject) => {
       const logPath = path.resolve(ckbPath(), 'data', 'logs', 'run.log')
-      if (!fs.existsSync(logPath)) { reject(new Error("File not found")) }
+      if (!fs.existsSync(logPath)) { return reject(new Error("File not found")) }
 
       const fileStats = fs.statSync(logPath)
       const position = fileStats.size - SIZE_TO_READ
 
       fs.open(logPath, 'r', (openErr, fd) => {
-        if (openErr) { reject(openErr) }
+        if (openErr) { return reject(openErr) }
         fs.read(fd, Buffer.alloc(SIZE_TO_READ), 0, SIZE_TO_READ, position, (readErr, _, buffer) => {
-          if (readErr) { reject(readErr) }
-          resolve(buffer.toString('utf8'))
+          if (readErr) { return reject(readErr) }
+          return resolve(buffer.toString('utf8'))
         })
       })
 
--- a/packages/neuron-wallet/src/controllers/export-debug.ts
+++ b/packages/neuron-wallet/src/controllers/export-debug.ts
@@ -12,7 +12,7 @@ import SyncedBlockNumber from 'models/synced-block-number'
 
 export default class ExportDebugController {
   archive: archiver.Archiver
-  constructor() {
+  constructor () {
     this.archive = archiver('zip', {
       zlib: { level: 9 }
     })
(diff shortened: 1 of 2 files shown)


packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -88,6 +89,32 @@ export default class ExportDebugController {
     })
   }
 
+  private addBundledCKBLog = () => {
+    const name = 'bundled-ckb-log.json'
+    const SIZE_TO_READ = 32_000
+
+    return new Promise((resolve, reject) => {
+      const logPath = path.resolve(ckbPath(), 'data', 'logs', 'run.log')
+      if (!fs.existsSync(logPath)) { reject(new Error("File not found")) }
```
katat, line 98: Should it stop processing for the rest of this promise if the file doesn't exist?
For example: `return reject(new Error("File not found"))`

packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -88,6 +89,32 @@ export default class ExportDebugController {
...
+
+      const fileStats = fs.statSync(logPath)
+      const position = fileStats.size - SIZE_TO_READ
+
+      fs.open(logPath, 'r', (openErr, fd) => {
+        if (openErr) { reject(openErr) }
```
katat, line 104: Same comment as above.

packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -88,6 +89,32 @@ export default class ExportDebugController {
...
+        fs.read(fd, Buffer.alloc(SIZE_TO_READ), 0, SIZE_TO_READ, position, (readErr, _, buffer) => {
+          if (readErr) { reject(readErr) }
```
katat, line 106: Same comment as above.

packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -88,6 +89,32 @@ export default class ExportDebugController {
...
+    const SIZE_TO_READ = 32_000
+
+    return new Promise((resolve, reject) => {
+      const logPath = path.resolve(ckbPath(), 'data', 'logs', 'run.log')
+      if (!fs.existsSync(logPath)) { reject(new Error("File not found")) }
+
+      const fileStats = fs.statSync(logPath)
+      const position = fileStats.size - SIZE_TO_READ
+
+      fs.open(logPath, 'r', (openErr, fd) => {
```
katat, line 103: Does it need to call `fs.close` once operations on `fd` is finished at the end?
https://stackoverflow.com/questions/21176733/what-the-scenario-call-fs-close-is-necessary

Keith-CY: Fixed in https://github.com/nervosnetwork/neuron/pull/1633/commits/5be3afc9f76e6174e6a1f55deae1940283ca1c16

packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -88,6 +89,39 @@ export default class ExportDebugController {
...
+        if (openErr) { return reject(openErr) }
+        fs.read(fd, Buffer.alloc(SIZE_TO_READ), 0, SIZE_TO_READ, position, (readErr, _, buffer) => {
+          if (readErr) {
+            reject(readErr)
+          } else {
+            resolve(buffer.toString('utf8'))
+          }
+          fs.close(fd, closeErr => {
```
katat, line 111: Thanks for the updates.
I think `fs.close` may be better to be called within the `then` and `catch` of this promise, because it might not be able to reach this point to close the `fd` due to potential errors above within the `fs.open` callback.

PR #1643, opened 2020-05-13
feat: independent settings

Keith-CY:
This PR separates the settings into an independent window and enable language switching in general settings.

image
image
image


Keith-CY:
@kellyshang Please test the independent settings window and language switch in this branch.

packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -58,6 +59,11 @@ export default class ApiController {
   private registerHandlers() {
     const handle = this.handleChannel
 
+    // sync messages
+    ipcMain.on('get-locale', e => {
```
katat, line 63: Can we use `handle` here instead, so that it is consistent with the rest of the invoking APIs?

Keith-CY: > Can we use `handle` here instead, so that it is consistent with the rest of the invoking APIs?

We could refactor it later, I used synchronous message here to minimize the update on messages cuz `app.getLocale` is synchronous.

PR #1650, opened 2020-05-15
Performance optimization

katat:
per https://github.com/nervosnetwork/neuron-internal/issues/1091

This PR includes the following updates to improve the sync performances:
 - Update block fetch size
 - Optimize sqlite query for checking outPoints for inputs
 - Add cache for network configurations

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/tx-address-finder.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/tx-address-finder.ts
@@ -88,15 +88,20 @@ export default class TxAddressFinder {
     const inputs = this.tx.inputs!.filter(i => i.previousOutput !== null)
     const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
 
+    const outputs = await getConnection()
+      .getRepository(OutputEntity)
+      .createQueryBuilder()
+      .getMany()
+
+    const outputsMap = new Map()
+    for (const output of outputs) {
+      outputsMap.set(`${output.outPointTxHash}_${output.outPointIndex}`, output)
+    }
+
     let shouldSync = false
     for (const input of inputs) {
       const outPoint: OutPoint = input.previousOutput!
-      const output = await getConnection()
-        .getRepository(OutputEntity)
-        .findOne({
-          outPointTxHash: outPoint.txHash,
-          outPointIndex: outPoint.index,
-        })
+      const output = outputsMap.get(`${outPoint.txHash}_${outPoint.index}`)
       if (output && this.anyoneCanPayLockHashes.has(output.lockHash)) {
         shouldSync = true
         // anyoneCanPayBlake160s.push(output.lockArgs)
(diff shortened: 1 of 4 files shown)


Keith-CY:
Cached network id and list are available by subscribing [NetworkSubject](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/models/subjects/networks.ts)

katat:
Thanks @Keith-CY for the suggestion.
I will take a look and see how to refactor it use the subscription model.

katat:
@Keith-CY 
I just took a look at the NetworkSubject. If I understand correctly, it requires changing the `block pipeline` code to adapt to the rxjs event pattern programming. However, the current `block pipeline` section is not coded in an event pattern. Feel free to let me know I got it right.

Meanwhile, I also spot there is already a network cache stored in local storage encapsulated in the  `localCache.ts`. Do you think it makes senses to use the caches `networks` and `currentNetworkID` instead?

Keith-CY:
> @Keith-CY
> I just took a look at the NetworkSubject. If I understand correctly, it requires changing the `block pipeline` code to adapt to the rxjs event pattern programming. However, the current `block pipeline` section is not coded in an event pattern. Feel free to let me know I got it right.
> 
> Meanwhile, I also spot there is already a network cache stored in local storage encapsulated in the `localCache.ts`. Do you think it makes senses to use the caches `networks` and `currentNetworkID` from `localCache.ts` instead?

It's not mandatory to convert the `block pipeline` into `rxjs-like` style. Just subscribe to the `network subject` where `cached network` is used like this:

```js
// with cached network
const doSomthingWithNetwork = () => {
  const { currentNetwork } = networkService
  doSomething({ network: currentNetwork })
}

// with network subject
const doSomethingWithNetwork = () => {
  NetworkSubject.pipe(take(1)).subscribe(currentNetwork => {
    doSomething({ network: currentNetwork })
  })
}
```

The `network service` and `network subject` both work in the main process while the `local cache` works in the renderer process. It's possible but not good to take the cached network from the renderer process.

katat:
@Keith-CY 
Thanks for the examples. I think it makes sense not to use localCache.ts within the backend domain as you said it is not meant to be used by the backend code.

I think I can go with the `NetworkSubject` approach for now, although the programming pattern seems a bit mismatch within the `pipeline`.

packages/neuron-wallet/src/block-sync-renderer/sync/tx-address-finder.ts
```diff
@@ -88,15 +88,20 @@ export default class TxAddressFinder {
     const inputs = this.tx.inputs!.filter(i => i.previousOutput !== null)
     const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
 
+    const outputs = await getConnection()
```
classicalliu, line 91: If wallet has about 1000 transaction outputs, execute this sql and push to the map will cost about `30 ~ 60ms`, so every 1000 blocks, it will cost about `30 ~ 60s`. It will be much slower than the original version.

PR #1651, opened 2020-05-15
Merge master of v0.30.0 back to develop

kellyshang:


Keith-CY:
Usually, we merge master back to dev locally and push it to the remote one.

I'll do it since there's a conflict in the `addresses.module.scss`.

kellyshang:
> Usually, we merge master back to dev locally and push it to the remote one.
> 
> I'll do it since there's a conflict in the `addresses.module.scss`.

Got you, thanks.

PR #1657, opened 2020-05-19
Refactor create asset account flow

classicalliu:


packages/neuron-wallet/src/services/asset-account-service.ts
```diff
@@ -271,22 +271,27 @@ export default class AssetAccountService {
...
+      throw new Error(`Asset account already exists!`)
     }
 
-    // 3. update address for usedByAnyoneCanPay
+    // 2. send tx
+    const txHash = await new TransactionSender().sendTx(walletID, tx, password)
+
+    // 3. save asset account
+    const entity = AssetAccountEntity.fromModel(assetAccount)
+    await connection.manager.save([entity.sudtTokenInfo, entity])
```
katat, line 292: Just curious. What would happen if the transacting operation succeeded on a node, but somehow the chain is doesn't accept the new transaction due various reasons such as uncle block or chain has been forked in the same period of transacting?

classicalliu: Transaction will failed, and it won't listed in asset accounts.

PR #1661, opened 2020-05-20
Merge sUDT into rc/v0.31.0

Keith-CY:


packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -172,7 +206,11 @@ export default class Queue {
             }
           }
           await TransactionPersistor.saveFetchTx(tx)
-          await WalletService.updateUsedAddresses(addresses)
+          const anyoneCanPayBlake160s = anyoneCanPayInfos.map(info => info.blake160)
+          await WalletService.updateUsedAddresses(addresses, anyoneCanPayBlake160s)
```
katat, line 210: I would be interested in knowing more about how the AddressDao works. Particularly how ACP related addresses are differentiated from the other addresses.

Maybe @classicalliu can clarify a bit on this one.

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -61,7 +61,7 @@ const trackingStatus = async () => {
     const blake160s = await FailedTransaction.updateFailedTxs(failedTxs.map(tx => tx.hash))
     const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
     const usedAddresses = blake160s.map(blake160 => AddressGenerator.toShortByBlake160(blake160, prefix))
-    await WalletService.updateUsedAddresses(usedAddresses)
+    await WalletService.updateUsedAddresses(usedAddresses, blake160s)
```
katat, line 64: Are `block160s` the ACP related lock hashes? If so, it may be better to include the ACP in terms of variable naming. For example: `anyoneCanPayBlake160s`

packages/neuron-wallet/src/services/asset-account-service.ts
```diff
@@ -0,0 +1,323 @@
...
+    const entity = AssetAccountEntity.fromModel(assetAccount)
+    const savedEntity = await connection.manager.save([entity.sudtTokenInfo, entity])
+
+    // 2. send tx
+    // if failed, remove saved entity
+    let txHash: string | undefined
+    try {
+      txHash = await new TransactionSender().sendTx(walletID, tx, password)
+    } catch (err) {
+      await connection.manager.remove(savedEntity)
```
katat, line 285: I think it would be valuable to keep this as an failed record instead of deleting them when the transaction is failed.

Having these failed transaction in DB would be helpful for the users to review the whole transaction history and also helpful when it comes to technical debugging.

@classicalliu @Keith-CY @kellyshang Let me know what you think :)

packages/neuron-wallet/src/controllers/anyone-can-pay.ts
```diff
@@ -0,0 +1,93 @@
...
+    return {
+      status: ResponseCode.Success,
+      result: tx,
+    }
+  }
+
+  public async generateSendAllTx(params: GenerateAnyoneCanPayAllTxParams): Promise<Controller.Response<Transaction>> {
+    return this.generateTx({
+      ...params,
+      amount: 'all'
```
katat, line 59: I couldn't find the relevant logics to handle this `all` amount inside the AnyoneCanPayService. I probably misunderstood something here.

@classicalliu

packages/neuron-wallet/src/controllers/anyone-can-pay.ts
```diff
@@ -0,0 +1,93 @@
...
+  tx: Transaction
+  password: string
+}
+
+export default class AnyoneCanPayController {
+  public async generateTx(params: GenerateAnyoneCanPayTxParams): Promise<Controller.Response<Transaction>> {
+    const tx = await AnyoneCanPayService.generateAnyoneCanPayTx(
+      params.walletID,
+      params.address,
+      params.amount,
```
katat, line 39: I couldn't find the logics to handle this `amount` variable in the `AnyoneCanPayService`: https://github.com/nervosnetwork/neuron/blob/c7e3067fd0864d82724139de73751764ce56c337/packages/neuron-wallet/src/services/anyone-can-pay.ts#L19

packages/neuron-wallet/src/database/address/address-dao.ts
```diff
@@ -113,25 +141,41 @@ export default class AddressDao {
         && value.version === version
         && value.addressType == AddressType.Receiving
         && value.txCount === 0
+        && !value.usedByAnyoneCanPay
```
katat, line 144: Isn't `value.txCount === 0` already covering `!value.usedByAnyoneCanPay`?

@classicalliu

katat: My understanding is when an address is used to send an ACP transaction, then the `txCount` will be incremented for the corresponding address.

packages/neuron-wallet/src/models/asset-account-info.ts
```diff
@@ -0,0 +1,85 @@
...
+  private static MAINNET_GENESIS_BLOCK_HASH: string = '0x92b197aa1fba0f63633922c61c92375c9c074a93e85963554f5499fe1450d0e5'
+
+  public get infos(): {[name: string]: ScriptCellInfo} {
+    return {
+      sudt: this.sudtInfo,
+      anyoneCanPay: this.anyoneCanPayInfo
+    }
+  }
+
+  constructor(genesisBlockHash: string = NetworksService.getInstance().getCurrent().genesisHash) {
```
katat, line 26: It might be better to encapsulate these cell deps in a JSON file, which would be easier for maintenance and more flexible for testings.

packages/neuron-wallet/src/services/tx/transaction-generator.ts
```diff
@@ -477,6 +479,266 @@ public static generateSendingAllTx = async (
 
     return tx
   }
+
+  // sUDT
+  public static async generateCreateAnyoneCanPayTx(
+    tokenID: string,
+    lockHashes: string[],
+    blake160: string,
```
katat, line 487: These variable names (...blake160...) would be better to be more explicit. In this case, it would be more readable to indicate whether it is involving the sender or recipient.

packages/neuron-wallet/src/controllers/anyone-can-pay.ts
classicalliu: Here: https://github.com/nervosnetwork/neuron/blob/sUDT/packages/neuron-wallet/src/services/cells.ts#L416

packages/neuron-wallet/src/database/address/address-dao.ts
classicalliu: `txCount` only covers transactions which uses default lock.

katat:
@Keith-CY There is a code conflict blocking auto-merging.

Keith-CY:
> @Keith-CY There is a code conflict blocking auto-merging.

conflicts fixed

PR #1667, opened 2020-05-21
feat: use the Copy Component on more texts

Keith-CY:
1. balance
2. capacity

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAORecord/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAORecord/index.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import CompensationProgressBar from 'components/CompensationProgressBar'
 import Button from 'widgets/Button'
+import CopyZone from 'widgets/CopyZone'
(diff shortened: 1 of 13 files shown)


packages/neuron-ui/src/components/NervosDAORecord/index.tsx
```diff
@@ -257,9 +260,7 @@ export const DAORecord = ({
         </span>
       </div>
 
-      <div className={styles.amount}>
-        <span>{`${shannonToCKBFormatter(capacity)} CKB`}</span>
-      </div>
+      <CopyZone className={styles.amount} content={amount.replace(/,/g, '')}>{`${amount} CKB`}</CopyZone>
```
katat, line 263: We may need a number formatter so that these replace regex can be re-used in different places.

Keith-CY: Yep, the method [shannon to ckb](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/utils/formatters.ts#L102) has the third param to specify the delimiter. I'll use update them in the future.

PR #1669, opened 2020-05-21
feat: enable exporing transaction history

Keith-CY:
image


packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -30,10 +33,21 @@ const History = () => {
...
   const history = useHistory()
   const { search } = useLocation()
+  const [isExporting, setIsExporting] = useState(false)
   const isMainnet = isMainnetUtil(networks, networkID)
 
   const { keywords, onKeywordsChange } = useSearch(search, id, dispatch)
   const onSearch = useCallback(() => history.push(`${RoutePath.History}?keywords=${keywords}`), [history, keywords])
+  const onExport = useCallback(() => {
+    setIsExporting(true)
+    const timer = setTimeout(() => {
```
katat, line 43: Just curious, why does it need to set a timeout here?

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+const exportHistory = ({
+  walletID,
+  dbPath,
+  lockHashList = [],
+  filePath
+}: ExportHistoryParms): Promise<number> => {
+  return new Promise((resolve, reject) => {
+    if (!dbPath) {
+      return reject(new Error(`Database is not required`))
```
katat, line 22: "Database is not required" => "Database is required"?

packages/neuron-ui/src/components/History/index.tsx
Keith-CY: The flag `isExporting` is not meant to indicate the completion of exporting the history cuz there's a dialog for that. This flag is used to prevent from double-click on the button.

packages/neuron-wallet/src/utils/export-history.ts
Keith-CY: > "Database is not required" => "Database is required"?

Sure, it's a typo.

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -0,0 +1,149 @@
...
+    let total: number | undefined
+    let inserted = 0
+
+    const writeStream = fs.createWriteStream(filePath)
+    writeStream.write(
+      `${['time', 'block-number', 'tx-hash', 'tx-type', 'amount', 'description']
+        .map(label => i18n.t(`export-transactions.column.${label}`))}\n`
+    )
+
+    const db = new sqlite3.Database(dbPath, sqlite3.OPEN_READONLY)
```
katat, line 46: Do we have to sqlite3 directly here? Would it be better to use typeorm since we use typeorm in most of the cases?

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -0,0 +1,149 @@
...
+import shannonToCKB from 'utils/shannonToCKB'
+
+
+interface ExportHistoryParms {
+  walletID: string
+  dbPath: string
+  lockHashList?: string[]
+  filePath: string
+}
+const exportHistory = ({
```
katat, line 14: Since it involves quite a number of logics in different parts, I would suggest having some unit tests or integration tests to cover them.

packages/neuron-wallet/src/utils/export-history.ts
Keith-CY: > Do we have to sqlite3 directly here? Would it be better to use typeorm since we use typeorm in most of the cases?

`Typeorm`'s `stream` doesn't work with `sqlite3` so I've to use `sqlite3#each` in case of bunches of data.

packages/neuron-wallet/src/utils/export-history.ts
Keith-CY: Sure, but this feature is urgent and the test cases of `TransactionService.getAll` are absent, so I've decided to add them a few later.

PR #1677, opened 2020-05-25
Fix layout of independent window

Keith-CY:
image
1. Set the title of Sign-And-Verify the same as that of Settings
2. Remove taskbar of independent windows
3. Disable resize of independent windows


Keith-CY:
image
@kellyshang The network settings is refined.

PR #1685, opened 2020-05-28
feat: cancel request handler if urls are not matched

Keith-CY:


packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -27,34 +27,45 @@ const SYNC_INTERVAL_TIME = 4000
...
-            payload: {
-              tipBlockNumber: `${BigInt(header.number)}`,
-              tipBlockHash: header.hash,
-              tipBlockTimestamp: +header.timestamp,
-              chain: chainInfo.chain,
-              difficulty: BigInt(chainInfo.difficulty),
-              epoch: chainInfo.epoch,
-            },
-          })
+          if (isCurrentUrl(chainURL)) {
```
katat, line 43: I am curious to know bit more about the scenarios where `chainURL` is not equal to `current url`. 
Can you please clarify a bit when the `isCurrentUrl(chainURL) === false`?

packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -27,34 +27,45 @@ const SYNC_INTERVAL_TIME = 4000
...
+            dispatch({
+              type: AppActions.UpdateChainInfo,
+              payload: {
+                tipBlockNumber: `${BigInt(header.number)}`,
+                tipBlockHash: header.hash,
+                tipBlockTimestamp: +header.timestamp,
```
katat, line 49: I am wondering why it is using the operation `+` here.

packages/neuron-ui/src/containers/Main/hooks.ts
Keith-CY: Suppose that the request is sent to `url a` but the response is held, while the network config is set to  `url b`. Then the response from `url a` should be discarded.

packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -27,34 +27,45 @@ const SYNC_INTERVAL_TIME = 4000
...
+              payload: ErrorCode.NodeDisconnected,
+            })
+          }
         })
         .catch(() => {
-          dispatch({
-            type: NeuronWalletActions.UpdateConnectionStatus,
-            payload: Date.now() > CONNECTING_DEADLINE ? ConnectionStatus.Offline : ConnectionStatus.Connecting,
-          })
+          if (isCurrentUrl(chainURL)) {
```
katat, line 63: Looks like this silences the other errors? If the other parts can't receive these errors, will there be any potential issues?

packages/neuron-ui/src/containers/Main/hooks.ts
Keith-CY: The type of `header.timestamp` is string while that of `state.tipBlockTimestamp` is number.

packages/neuron-ui/src/containers/Main/hooks.ts
Keith-CY: > Looks like this silences the other errors? If the other parts can't receive these errors, will there be any potential issues?

The response from an inactive URL should be ignored.

packages/neuron-ui/src/containers/Main/hooks.ts
katat: Ok, I see now. So this `+` operator is used to implicitly convert string to a number.
Maybe it is more readable by using `parseInt(header.timestamp)`.

Keith-CY: > Ok, I see now. So this `+` operator is used to implicitly convert string to a number.
> Maybe it is more readable by using `parseInt(header.timestamp)`.

`+` is better than `parseInt` in most cases cuz more types are correctly handled by the `+` operator, e.g. `null` is converted into `0` by `+null` while `NaN` by `parseInt(null)`

katat: Thanks for the clarification. I see it is intended for convenience. 

I think the point to use `parseInt` is that it can precisely return the value in context of number type. In other words, `NaN` returned by `parseInt` for a `null` value makes more sense in my view. 

So `parseInt(value) === NaN ? 0 : parseInt(value)` in here seems more readable and less potential security issues that would be introduced by the use of `+` operator, which basically coalesce certain values in an opinioned way.

Keith-CY: > Thanks for the clarification. I see it is intended for convenience.
> 
> I think the point to use `parseInt` is that it can precisely return the value in context of number type. In other words, `NaN` returned by `parseInt` for a `null` value makes more sense in my view.
> 
> So `parseInt(value) === NaN ? 0 : parseInt(value)` in here seems more readable and less potential security issues that would be introduced by the use of `+` operator, which basically coalesce certain values in an opinioned way.

In most cases I've met and read the `+` is preferable to the `parseInt`. BTW, the `parseInt(value) === NaN` is always `false`.

katat: I think the potential issue could be at the edge cases.

Try `isNaN(parseInt(null))`

PR #1687, opened 2020-05-31
Update testnet acp & sudt script info

classicalliu:


Keith-CY:
Let's add the script info in the `env.ts` that we can debug on the local dev chain without overriding the info for mainnet/testnet.

PR #1690, opened 2020-06-01
feat: detect vcredist

Keith-CY:
allow user to install/skip/exit when the vcredist is missing

### DIFF
--- a/packages/neuron-wallet/src/services/node.ts
+++ b/packages/neuron-wallet/src/services/node.ts
@@ -1,9 +1,14 @@
-import { remote } from 'electron'
+import { exec } from 'child_process'
+import { promisify } from 'util'
+import path from 'path'
+import https from 'https'
+import http from 'http'
+import { remote, dialog, shell } from 'electron'
 import CKB from '@nervosnetwork/ckb-sdk-core'
 import { interval, BehaviorSubject, merge } from 'rxjs'
 import { distinctUntilChanged, sampleTime, flatMap, delay, retry, debounceTime } from 'rxjs/operators'
-import https from 'https'
-import http from 'http'
+import env from 'env'
+import i18n from 'locales/i18n'
 import { ShouldBeTypeOf } from 'exceptions'
 import { ConnectionStatusSubject } from 'models/subjects/node'
 import { CurrentNetworkIDSubject } from 'models/subjects/networks'
@@ -38,7 +43,7 @@ class NodeService {
 
   public ckb: CKB = new CKB('')
 
-  constructor() {
+  constructor () {
     this.start()
     this.syncConnectionStatus()
     CurrentNetworkIDSubject.subscribe(async ({ currentNetworkID }) => {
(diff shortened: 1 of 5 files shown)


packages/neuron-wallet/src/services/node.ts
```diff
@@ -138,13 +143,66 @@ class NodeService {
...
+      title: i18n.t(`${I18N_PATH}.title`),
+      message: i18n.t(`${I18N_PATH}.message`),
+      detail: i18n.t(`${I18N_PATH}.detail`),
+      cancelId: 0,
+      noLink: true,
+    }).then(({ response }) => {
+      switch (response) {
+        case 1: {
+          // open browser and shutdown
+          const VC_REDIST_URL = `https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads`
```
katat, line 166: Maybe we should put this URL in `env`?

packages/neuron-wallet/src/services/node.ts
```diff
@@ -138,13 +143,66 @@ class NodeService {
...
+
+  private detectDependency = async () => {
+    if (process.platform !== 'win32') {
+      return true
+    }
+    const execPromise = promisify(exec)
+    const arches = process.arch === 'x64' ? ['x86', 'x64'] : ['x64']
+    const queries = arches.map(arch =>
+      `REG QUERY ` +
+      [`HKEY_LOCAL_MACHINE`, `SOFTWARE`, `Microsoft`, `VisualStudio`, `14.0`, `VC`, `Runtimes`, arch].join(path.sep))
```
katat, line 188: Not sure if I got it right.
What if the link `https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads` provides the download for version `15.0`? Will the dialog keep prompting because it checks against version `14`?

packages/neuron-wallet/src/services/node.ts
Keith-CY: > Maybe we should put this URL in `env`?

It seems that the link has nothing to do with the environment so why should it be in the `env`?

packages/neuron-wallet/src/services/node.ts
Keith-CY: It does have a potential problem and the hardcoded query is not ideal.

For now, it's just a workaround cuz the version specified in the documentation is 2019, which shares the same files with 2015/2017 and they are all located in the `VisualStudio/14/...`.

We should update the query along with the version specified in the documentation and perhaps we should make it automatic in the CI.

Speaking of the question

> What happens if the link https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads provides the download for version 15.0? Will the dialog keep prompting because it checks against version 14?

For now, we should still check the `14`, and users should pick the `vc_redist 2015-2019` as the documentation says.

katat: We may also need to consider the cases for the older versions of windows, such as Win7, which is not supported officially anymore. I am wondering if the latest `vc_redist` is supported on such legacy windows versions. If they are not supported, meaning users can only install older versions of `vc_redist`, which would prompt the dialog every time the Neuron starts?

Do we know the lowest version of `vc_redist` that CKB node requires? If so, maybe we can loop and check the existence of registres from the lowest to highest version. This way may provide a bit more compatibility over combinations of different windows and `vc_redist` versions.

packages/neuron-wallet/src/services/node.ts
katat: My initial thought is that it is a URL variable, which is normally not hardcoded and should be referenced from a place responsible for configurations. It is cleaner to update this type of variable in a configuration file, which is easier to plugin to a CI task when necessary.

But I think you are right, the incentive of relocating the value of this variable is not that obvious.

packages/neuron-wallet/src/services/node.ts
Keith-CY: The consideration does make sense.
The vc_redist 2015-2019 works well on Win7 thanks to the awesome backward-compatibility of Windows.
And the officially specified by CKB is the 2019 one

> (minimum required) Or you can just select two individual components: "MSVC v142 - VS 2019 C++ x64/x86 build tools (vXX.XX)" and "Windows 10 SDK (10.0.X.0)".

katat: Sounds good.
Let's keep our eyes on the newer versions of `vs_redist` in the future, because the newer version could bring in higher number than `14`, which would unnecessarily trigger the dialog. At that stage, we need to refactor the code to make it more adaptable if we decided to leave it for now.

PR #1692, opened 2020-06-02
Enable sudt on testnet

Keith-CY:
This branch enables the sUDT when it's the TESTNET.

### DIFF
--- a/packages/neuron-ui/src/utils/hooks.ts
+++ b/packages/neuron-ui/src/utils/hooks.ts
@@ -1,6 +1,7 @@
-import { useState, useMemo, useCallback, useEffect } from 'react'
+import { useState, useMemo, useCallback, useEffect, useRef } from 'react'
 import { useHistory } from 'react-router-dom'
 import { TFunction, i18n as i18nType } from 'i18next'
+import CKB from '@nervosnetwork/ckb-sdk-core'
 import { openContextMenu, requestPassword, setCurrentNetowrk, deleteNetwork } from 'services/remote'
 import { SetLocale as SetLocaleSubject } from 'services/subjects'
 import {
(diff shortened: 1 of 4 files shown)


Keith-CY:
This PR is pending until the https://github.com/nervosnetwork/neuron/pull/1691 is merged. Please test the functionality on this branch for now @kellyshang.

kellyshang:
> This PR is pending until #1691 is merged. Please test the functionality on this branch for now @kellyshang.

OK.

Keith-CY:
> > This PR is pending until #1691 is merged. Please test the functionality on this branch for now @kellyshang.
> 
> OK.

I've added a commit that navigates the page from sUDT to Overview when the network is checked from TESTNET to NON-TESTNET, please pull again.

packages/neuron-ui/src/utils/hooks.ts
```diff
@@ -393,3 +394,46 @@ export const useOnHandleNetwork = ({ history }: { history: ReturnType<typeof use
...
+  const timerRef = useRef<NodeJS.Timeout | undefined>()
+  const BUFFER_TIME = 200
+
+  useEffect(() => {
+    if (timerRef.current) {
+      clearTimeout(timerRef.current)
+    }
+
+    if (url) {
+      timerRef.current = setTimeout(() => {
```
katat, line 408: Could you please clarify a bit on why we need `setTimeout` here? Is it because it needs to wait for the connection to the CKB node?
I am concerned the use of `setTimeout` would introduce some stability issues at a certain stage. Are there other ways to achieve the same purpose without using the `setTimeout`?

Keith-CY: > Could you please clarify a bit on why we need `setTimeout` here? Is it because it needs to wait for the connection to the CKB node?
> I am concerned the use of `setTimeout` would introduce some stability issues at a certain stage. Are there other ways to achieve the same purpose without using the `setTimeout`?

It works as a `debounce` method to reduce unexpected HTTP requests.

PR #1693, opened 2020-06-02
Handle invalid sUDT data

katat:
- Added `BufferUtils#parseAmountFromSUDTData(leHex)` 
- Restructured and added unit tests for `BufferUtils` 
- Uses `BufferUtils#parseAmountFromSUDTData(leHex)` to parse amounts for sUDT data

### DIFF
--- a/packages/neuron-wallet/src/utils/buffer.ts
+++ b/packages/neuron-wallet/src/utils/buffer.ts
@@ -25,7 +25,7 @@ export default class BufferUtils {
 
   public static parseAmountFromSUDTData(leHex: string): bigint {
     try {
-      return this.readBigUInt128LE(leHex)
+      return this.readBigUInt128LE(leHex.slice(0, 34))
     } catch (error) {
       return BigInt(0)
     }
--- a/packages/neuron-wallet/src/utils/buffer.ts
+++ b/packages/neuron-wallet/src/utils/buffer.ts
@@ -22,4 +22,12 @@ export default class BufferUtils {
     const buf = Buffer.from(leHex.slice(2), 'hex')
     return (buf.readBigUInt64LE(8) << BigInt(64)) + buf.readBigUInt64LE(0)
   }
+
+  public static parseAmountFromSUDTData(leHex: string): bigint {
+    try {
+      return this.readBigUInt128LE(leHex)
+    } catch (error) {
+      return BigInt(0)
+    }
+  }
 }
--- a/packages/neuron-wallet/src/services/asset-account-service.ts
+++ b/packages/neuron-wallet/src/services/asset-account-service.ts
@@ -63,7 +63,7 @@ export default class AssetAccountService {
       } else {
         const sumOfAmount = (output.dataArray as string)
           .split(',')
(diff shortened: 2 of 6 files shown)


packages/neuron-wallet/src/utils/buffer.ts
```diff
@@ -22,4 +22,12 @@ export default class BufferUtils {
     const buf = Buffer.from(leHex.slice(2), 'hex')
     return (buf.readBigUInt64LE(8) << BigInt(64)) + buf.readBigUInt64LE(0)
   }
+
+  public static parseAmountFromSUDTData(leHex: string): bigint {
+    try {
+      return this.readBigUInt128LE(leHex)
```
classicalliu, line 28:
```suggestion
      return this.readBigUInt128LE(leHex.slice(0, 34))
```

sUDT only read first 16 bytes of cell data

packages/neuron-wallet/src/utils/buffer.ts
```diff
@@ -22,4 +22,12 @@ export default class BufferUtils {
     const buf = Buffer.from(leHex.slice(2), 'hex')
     return (buf.readBigUInt64LE(8) << BigInt(64)) + buf.readBigUInt64LE(0)
   }
+
+  public static parseAmountFromSUDTData(leHex: string): bigint {
```
Keith-CY, line 26: It seems that hex string whose length is not 34 is invalid in the method and 0 will be returned. But according to the [RFC Draft](https://talk.nervos.org/t/rfc-simple-udt-draft-spec/4333)

>Simple UDT Rule 1: a SUDT cell must store SUDT amount in the first 16 bytes of cell data segment, the amount should be stored as little endian, 128-bit unsigned integer format. In the case of composable scripts, the SUDT amount must still be located at the initial 16 bytes in the data segment which corresponds to the composed SUDT script

The first 16 bytes will be treated as the amount field, namely the data whose length is greater than 34 should be regarded as valid.

packages/neuron-wallet/src/utils/buffer.ts
katat: Thanks @classicalliu for reviewing.

I think we want to follow the strict sUDT protocol standard for the data field here. If it does not strictly equal to 16 bytes, the `readBigUInt128LE ` will throw an error. Then `parseAmountFromSUDTData ` will catch and return `BigInt(0)` for this invalid sUDT data, so that the downstream won't crash or sum up the amount.

packages/neuron-wallet/src/utils/buffer.ts
katat: That is interesting to know. Great reference from the RFC. Any thoughts on this one @ququzone? Do you think we should treat it strictly 16 bytes or we can just parse the first 16 bytes and ignore the rest of the hex string?

katat: @Keith-CY @classicalliu 
Updated the code to support parsing for the first 16 bytes in the data field even when the field is greater than 16 bytes.

PR #1703, opened 2020-06-08
Fix mainnet acp resync

ququzone:


packages/neuron-wallet/src/models/synced-block-number.ts
```diff
@@ -87,7 +87,21 @@ export default class SyncedBlockNumber {
     if (!this.#liveCellBlockNumberEntity) {
       this.#liveCellBlockNumberEntity = new SyncInfoEntity()
       this.#liveCellBlockNumberEntity.name = SyncInfoEntity.CURRENT_LIVE_CELL_BLOCK_NUMBER
-      this.#liveCellBlockNumberEntity.value = assetAccountInfo.sudtDeployHeight.toString()
+      const sudtDeployHeight = assetAccountInfo.sudtDeployHeight.toString()
+      if (sudtDeployHeight != "-1") {
```
katat, line 91: use strict `!==`

packages/neuron-wallet/src/models/asset-account-info.ts
```diff
@@ -37,7 +37,7 @@ export default class AssetAccountInfo {
         codeHash: '0x6a3982f9d018be7e7228f9e0b765f28ceff6d36e634490856d2b186acf78e79b',
         hashType: ScriptHashType.Type
       }
-      this.sudtDeployHeight = BigInt(14922)
+      this.sudtDeployHeight = BigInt(-1)
```
katat, line 40: Maybe setting this to null is more meaningful.

katat:
@ququzone Please also reorganize the udt/acp settings to the configuration file `env`.

packages/neuron-wallet/package.json
```diff
@@ -38,6 +38,7 @@
     "archiver": "3.1.1",
     "bn.js": "4.11.8",
     "chalk": "3.0.0",
+    "dotenv": "^8.2.0",
```
Keith-CY, line 41:
```suggestion
    "dotenv": "8.2.0",
```

Keith-CY: Isn't the `lock` file updated?

packages/neuron-wallet/.env
```diff
@@ -0,0 +1,25 @@
+## mainnet
```
Keith-CY, line 1: Uppercases are preferable.

packages/neuron-wallet/.env
```diff
@@ -0,0 +1,25 @@
...
+MAINNET_SUDT_DEP_TXHASH=0x0000000000000000000000000000000000000000000000000000000000000000
+MAINNET_SUDT_DEP_INDEX=0
+MAINNET_SUDT_DEP_TYPE=code
+MAINNET_SUDT_SCRIPT_CODEHASH=0x0000000000000000000000000000000000000000000000000000000000000000
+MAINNET_SUDT_SCRIPT_HASHTYPE=data
+MAINNET_SUDT_HEIGHT=-1
+MAINNET_ACP_DEP_TXHASH=0x0000000000000000000000000000000000000000000000000000000000000000
+MAINNET_ACP_DEP_INDEX=0
+MAINNET_ACP_DEP_TYPE=code
+MAINNET_ACP_SCRIPT_CODEHASH=0x0000000000000000000000000000000000000000000000000000000000000000
```
katat, line 11: Do we have to set these variables to a dummy values if they are not available yet on mainnet?
Can we just comment them out for now?

katat: I think it's better to leave them `NULL` value if these values are not available yet. Setting them to a dummy value like `0x0000000000000000000000000000000000000000000000000000000000000000 ` or `-1` could be misleading.

ququzone: `null` will cause type mismatch error

katat: Then we can change the type to be nullable.

katat: After a detailed look at the codes coupled with these variables, it does need some works to refactor these values to `nulls`. 

I will create an issue to remind the refactoring of these variables in a later stage.

Keith-CY:
Please cherry-pick to the [rc/v0.31.0](https://github.com/nervosnetwork/neuron/tree/rc/v0.31.0)

PR #1711, opened 2020-06-09
Indexer refactor

katat:
- Refactored the indexing process using the lumos indexer with a caching mechanism.
- Made it capable of querying live cells on-demand without caching the live cells on-chain and deprecated the live cells table for storage.
- Fixed async error in tests originated from the leveldb related operations by mocking up XHR requests.
- Fixed a bug in the asset account feature in which the newly created UDT account might not be shown up.
- Enable asset account feature for dev chain as well.


packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -1,251 +1,208 @@
...
-  public start = async () => {
-    await this.expandToTip(await this.rpcService.getTipBlockNumber())
-
-    this.tipNumberListener = this.tipNumberSubject.subscribe(async num => {
-      if (num) {
-        await this.expandToTip(num)
+  public async start() {
+    const indexedDataPath = path.resolve(
+      env.fileBasePath,
+      './indexer_data',
```
Keith-CY, line 57: Don't use magic string

packages/neuron-wallet/tests/setup.ts
```diff
@@ -0,0 +1,26 @@
...
+  readyState: null,
+  responseText: null,
+}
+
+Object.defineProperty(mockedXMLHttpRequest, 'readyState', {
+  get: jest.fn(() => 4),
+  set: jest.fn()
+});
+Object.defineProperty(mockedXMLHttpRequest, 'responseText', {
+  get: jest.fn(() => JSON.stringify([])),
```
Keith-CY, line 15: What does `JSON.stringify([])` mean?

packages/neuron-wallet/tests/block-sync-render/queue.test.ts
```diff
@@ -0,0 +1,244 @@
...
+
+const stubbedRPCServiceConstructor = jest.fn().mockImplementation(
+  () => ({
+    getChain: stubbedGetChainFn,
+    getTransaction: stubbedGetTransactionFn,
+    genesisBlockHash: stubbedGenesisBlockHashFn
+  })
+)
+
+const resetMocks = () => {
```
Keith-CY, line 49: `jest.resetAllMocks()` should help.

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
katat: Could you please clarify a bit on what you mean by `magic string` here?

packages/neuron-wallet/tests/setup.ts
katat: It converts a JSON to a string. https://www.w3schools.com/js/js_json_stringify.asp

Keith-CY: I mean, why is a stringified `[]` returned

packages/neuron-wallet/tests/block-sync-render/queue.test.ts
katat: Thanks for the suggestion.
However, the `jest.resetAllMocks` somehow is not working well with the mocks from `jest.doMock`, which we use to mock import dependencies.

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
Keith-CY: A plain string without any indicators. In this case, the `indexer_data` seems to be a dir keeping databases.

packages/neuron-wallet/tests/setup.ts
katat: No specific reasons. It is just a mocked return for all the XMLHttpRequests in the tests. Later we may want to change this value if it is necessary in some tests.

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
katat: It is the folder path for holding the data built from lumos indexer? Any suggestions to better name this one from your view?

packages/neuron-wallet/src/services/live-cell-service.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+      } else {
+        return item
+      }
+    }
+
+    return null
+  }
+
+  public async getByLockScriptAndTypeScript(lock: Script | null, type: Script | null) {
```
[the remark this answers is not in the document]
katat: I think this is a good suggestion.
@ququzone Can you handle this one? Thanks

packages/neuron-wallet/src/services/live-cell-service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    return cells
+  }
+
+  public async getByLockScriptsAndTypeScript(locks: Script[], type: Script | null) {
+    const result = []
+
+    for (const lock of locks) {
+      const cells = await this.getByLockScriptAndTypeScript(lock, type)
+      for (const cell of cells) {
```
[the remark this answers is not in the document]
katat: Good to know this one. Thanks
@ququzone Do you agree to make an update to this one?

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
Keith-CY: Does `data built from lumos indexer` mean the `rocksdb` files? The `indexer_data` could be named as `LUMOS_INDEXER_DB_FOLDER` if it is.

katat:
@Keith-CY @ququzone 
Please see the latest commits according to the review comments.

packages/neuron-wallet/src/block-sync-renderer/index.ts
```diff
@@ -132,3 +122,15 @@ export const killBlockSyncTask = () => {
     backgroundWindow.close()
   }
 }
+
+export const queryIndexer = async (query: any): Promise<LumosCell[]> => {
```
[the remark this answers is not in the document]
katat: Yep, it should be. I have just updated this one.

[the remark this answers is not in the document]
katat: This `async` should have been removed.

packages/neuron-wallet/src/block-sync-renderer/index.ts
```diff
@@ -132,3 +122,15 @@ export const killBlockSyncTask = () => {
...
+  return new Promise(resolve => {
```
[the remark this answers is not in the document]
katat: Really great spot 👍 

I have pushed the update to prevent potential racing situations.

katat:
After I merged the develop branch, now the test throws
```
tests/utils/export-history/index.test.ts
  Test exporting history › when several transactions exported › should export table with records

    TypeError: Cannot read property 'get' of undefined



      at Object.<anonymous>.exports.get (src/database/leveldb/transaction-description.ts:260:20)
      at Object.exportHistory [as default] (src/utils/export-history.ts:1197:62)
      at Object.<anonymous> (tests/utils/export-history/index.test.ts:144:26)

  Test exporting history › when several transactions exported › should export table with records

    expect.assertions(2)

    Expected two assertions to be called but received zero assertion calls.

      140 |
      141 |     it('should export table with records', async () => {
    > 142 |       expect.assertions(2)
          |              ^
      143 |       const expectedTotalCount = 3
      144 |       const totalCount = await exportHistory(stubProvider)
      145 |       const actual = fs.readFileSync(stubProvider.filePath, 'utf8')

      at Object.<anonymous> (tests/utils/export-history/index.test.ts:142:14)
```

Looks like it is something related to the leveldb. I guess the tests for exporting history is are deep integration tests that somehow trigger this error.

Keith-CY:
> @Keith-CY
> After I merged the develop branch, now the test throws
> 
> ```
> tests/utils/export-history/index.test.ts
>   Test exporting history › when several transactions exported › should export table with records
> 
>     TypeError: Cannot read property 'get' of undefined
> 
> 
> 
>       at Object.<anonymous>.exports.get (src/database/leveldb/transaction-description.ts:260:20)
>       at Object.exportHistory [as default] (src/utils/export-history.ts:1197:62)
>       at Object.<anonymous> (tests/utils/export-history/index.test.ts:144:26)
> 
>   Test exporting history › when several transactions exported › should export table with records
> 
>     expect.assertions(2)
> 
>     Expected two assertions to be called but received zero assertion calls.
> 
>       140 |
>       141 |     it('should export table with records', async () => {
>     > 142 |       expect.assertions(2)
>           |              ^
>       143 |       const expectedTotalCount = 3
>       144 |       const totalCount = await exportHistory(stubProvider)
>       145 |       const actual = fs.readFileSync(stubProvider.filePath, 'utf8')
> 
>       at Object.<anonymous> (tests/utils/export-history/index.test.ts:142:14)
> ```
> 
> Looks like it is something related to the leveldb. I guess the tests for exporting history are deep integration tests that somehow trigger this error.

Actually it's not deeply integrated, this method(transaction-description#get) is directly used in the module: https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/utils/export-history.ts#L60

But I do think so that the error comes from the leveldb.

Keith-CY:
> > @Keith-CY
> > After I merged the develop branch, now the test throws
> > ```
> > tests/utils/export-history/index.test.ts
> >   Test exporting history › when several transactions exported › should export table with records
> > 
> >     TypeError: Cannot read property 'get' of undefined
> > 
> > 
> > 
> >       at Object.<anonymous>.exports.get (src/database/leveldb/transaction-description.ts:260:20)
> >       at Object.exportHistory [as default] (src/utils/export-history.ts:1197:62)
> >       at Object.<anonymous> (tests/utils/export-history/index.test.ts:144:26)
> > 
> >   Test exporting history › when several transactions exported › should export table with records
> > 
> >     expect.assertions(2)
> > 
> >     Expected two assertions to be called but received zero assertion calls.
> > 
> >       140 |
> >       141 |     it('should export table with records', async () => {
> >     > 142 |       expect.assertions(2)
> >           |              ^
> >       143 |       const expectedTotalCount = 3
> >       144 |       const totalCount = await exportHistory(stubProvider)
> >       145 |       const actual = fs.readFileSync(stubProvider.filePath, 'utf8')
> > 
> >       at Object.<anonymous> (tests/utils/export-history/index.test.ts:142:14)
> > ```
> > 
> > 
> > Looks like it is something related to the leveldb. I guess the tests for exporting history are deep integration tests that somehow trigger this error.
> 
> Actually it's not deeply integrated, this method(transaction-description#get) is directly used in the module: [/packages/neuron-wallet/src/utils/export-history.ts@`develop`#L60](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/utils/export-history.ts?rgh-link-date=2020-06-29T07%3A41%3A44Z#L60)
> 
> But I do think so that the error comes from the leveldb.

I think it's from the dep order, maybe we should introduce IOC in the project.

katat:
Then it should be fixed by introducing necessary mocks around the integration tests.

Keith-CY:
> Then it should be fixed by introducing necessary mocks around the integration tests.

I'm not sure, I'd like to regard it as a bug..

katat:
Now it should be fixed by https://github.com/nervosnetwork/neuron/pull/1711/commits/c1698603b09cb94d35c0c51f1b88ede16f7b88d8

Keith-CY:
> > Then it should be fixed by introducing necessary mocks around the integration tests.
> 
> I'm not sure, I'd like to treat it as a bug..

I think the `transaction-description#get` is expected to return an empty string as its default action.

katat:
Yep, we will deprecate `leveldb` in near future anyway.

Keith-CY:
But will we leave the potential problem of dep order there?

katat:
I think that the failed test is not caused by `dep order`. Instead, it is caused by the global mock for `leveldb`, which previously returns undefined by default. That is why we saw that error.

katat:
@Keith-CY @ququzone 
Finally, the CI builds are all passed without errors. 
I have pushed the updates according to the review comments, please take a look.

Keith-CY:
Please rebase or squash before a merge.

PR #1712, opened 2020-06-09
feat: update style of balance in overview and send

Keith-CY:
Make them consistent with each other
image
image

### DIFF
--- a/packages/neuron-ui/src/components/Send/index.tsx
+++ b/packages/neuron-ui/src/components/Send/index.tsx
@@ -1,6 +1,6 @@
 import React, { useMemo, useState, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
-import { Label, Text, List } from 'office-ui-fabric-react'
+import { List } from 'office-ui-fabric-react'
 import { ckbCore } from 'services/chain'
 import TransactionFeePanel from 'components/TransactionFeePanel'
 import TextField from 'widgets/TextField'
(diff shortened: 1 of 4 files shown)


packages/neuron-ui/src/components/Send/index.tsx
```diff
@@ -149,36 +149,15 @@ const Send = () => {
...
-  } else if (
-    [SyncStatus.Syncing, SyncStatus.SyncPending].includes(syncStatus) ||
-    ConnectionStatus.Connecting === connectionStatus
-  ) {
-    balancePrompt = <span className={styles.balancePrompt}>{t('sync.syncing-balance')}</span>
-  }
-
   return (
-    <form style={{ padding: '39px 0 0 0' }} onSubmit={disabled ? undefined : onSubmit} data-wallet-id={walletID}>
+    <form onSubmit={disabled ? undefined : onSubmit} data-wallet-id={walletID}>
```
katat, line 153: It might be better to encapsulate the `disabled` check within `onSubmit` callback function for mainly two reasons:
1. This section of onSubmit binding would be more readable. Normally I would try to avoid putting too much logic in the JSX.
2. Later on, it would be easier to handle the extended logics along with the current `disabled` within the callback function.

Keith-CY: Sure, I'll do it later

PR #1719, opened 2020-06-10
refactor: check the form status in the onSubmit handler

Keith-CY:


packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -180,9 +180,12 @@ const useOnSubmit = (items: Readonly<State.Output[]>, dispatch: StateDispatch) =
   useCallback(
     (e: React.FormEvent) => {
       const {
-        dataset: { walletId },
+        dataset: { walletId, status },
```
katat, line 183: Looks like the react hook doesn't provide a way to derive this status, using react props, without the need of passing from the upstream 🤔

Keith-CY: Yes, parameters should be passed explicitly.

PR #1720, opened 2020-06-10
feat: share token infos across accounts

Keith-CY:
1. add an api to fetch token info list
2. fetch the list on creating an account

### DIFF
--- a/packages/neuron-wallet/tests/services/asset-account-service.test.ts
+++ b/packages/neuron-wallet/tests/services/asset-account-service.test.ts
@@ -485,35 +485,25 @@ describe('AssetAccountService', () => {
 
   describe('Test Token Info List', () => {
     beforeEach(async () => {
-      const assetAccounts = [
-        AssetAccount.fromObject({
+      const tokens = [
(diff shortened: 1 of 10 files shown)


packages/neuron-wallet/tests/services/asset-account-service.test.ts
```diff
@@ -482,4 +482,43 @@ describe('AssetAccountService', () => {
       expect(result.length).toEqual(2)
     })
   })
+
+  describe('Test Token Info List', () => {
+    beforeEach(async () => {
+      const assetAccounts = [
```
katat, line 488: Since we only test against the `SudtTokenInfo` entity, do we have to persist these `Asset Accounts`? I think we can just persist the `SudtTokenInfo` entities for this test case?

Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1720/commits/c11d27d991f5183bdb99d6742b102997a46eeb3c

PR #1723, opened 2020-06-11
feat: refine exporting history

Keith-CY:


Keith-CY:
I'm adding more test cases. Please review the logic, please.

packages/neuron-ui/src/utils/formatters.ts
```diff
@@ -244,16 +243,17 @@ export const sudtValueToAmount = (value: string = '0', decimal: string = '0', sh
...
   const unsignedValue = value.replace(/^-?0*/, '')
+  const dec = +decimal
   if (dec === 0) {
     return +unsignedValue ? `${sign}${unsignedValue}` : '0'
   }
-  let unsignedCKB = ''
+  let unsignedSUDTValue = ''
   if (unsignedValue.length <= dec) {
-    unsignedCKB = `0.${unsignedValue.padStart(dec, '0')}`.replace(/\.?0+$/, '')
+    unsignedSUDTValue = `0.${unsignedValue.padStart(dec, '0')}`.replace(/\.?0+$/, '')
```
katat, line 252: If this regex pattern is used in more than one place, I would suggest creating a function to encapsulate it.

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -1,110 +1,269 @@
...
 import sqlite3 from 'sqlite3'
 import { get as getDescription } from 'database/leveldb/transaction-description'
 import i18n from 'locales/i18n'
+import AssetAccountInfo from 'models/asset-account-info'
 import shannonToCKB from 'utils/shannonToCKB'
+import sudtValueToAmount from 'utils/sudt-value-to-amount'
+import BufferUtils from 'utils/buffer'
+import { ChainType } from 'models/network'
 
+const formatDatetime = (datetime: Date) => {
```
katat, line 12: We may want to consider using library such as `moment.js` to parse the date-time format. `moment.js` also supports localization.

packages/neuron-ui/src/utils/formatters.ts
Keith-CY: OK, I'll do it later because the formatters of CKB value and sUDT value are going to be merged into a general one when the Send Components of CKB and sUDT are merged.

We separate them deliberately when sUDT is experimental.

packages/neuron-wallet/src/utils/export-history.ts
Keith-CY: For the sake of security, we prefer not to use a third-party package.

katat: Then this would mean we need to reinvent a lot of wheels 🤔

Keith-CY: > Then this would mean we need to reinvent a lot of wheels 🤔

Yes, we have to do that. There're real-world attacks from famous third-party packages.

Ref: https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident

katat: We also need to have a check and see if `npm audit` can help here. https://blog.npmjs.org/post/173719309445/npm-audit-identify-and-fix-insecure#:~:text=npm%20audit%20is%20a%20new,and%20recommendations%20for%20further%20troubleshooting.

Keith-CY: > We also need to have a check and see if `npm audit` can help here. [blog.npmjs.org/post/173719309445/npm-audit-identify-and-fix-insecure#:~:text=npm%20audit%20is%20a%20new,and%20recommendations%20for%20further%20troubleshooting](https://blog.npmjs.org/post/173719309445/npm-audit-identify-and-fix-insecure#:~:text=npm%20audit%20is%20a%20new,and%20recommendations%20for%20further%20troubleshooting).

`npm audit` or `yarn upgrade-interactive` are not smart enough.

We have compared `Renovate`, `Snyk`, and `dependabot` and picked the last one as our dep bot who will push fixes timely.

E.g. https://github.com/nervosnetwork/neuron/pull/1700 and

Keith-CY:
The TypeORM doesn't support a shared in-memory sqlite database which blocks more tests. So I prefer to do the tests manually for now and try to find a way better than using a temporary on-disk database.

Keith-CY:
@kellyshang Please test on this branch before merging.

packages/neuron-wallet/src/env.ts
```diff
@@ -19,6 +17,11 @@ const app = electronApp ?? (remote?.app) ?? {
 const isTestMode = process.env.NODE_ENV === 'test'
 const isDevMode = !app.isPackaged && !isTestMode
 
+if (isDevMode) {
+  dotenv.config({ path: path.resolve(__dirname, '..', '.env.development.local') })
```
katat, line 21: Maybe `.env.local` is expressive enough?

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -1,110 +1,269 @@
...
+    SELECT
+      tokenID,
+      symbol,
+      decimal
+    FROM
+      sudt_token_info
+    `
+  )
+
+  return new Promise((resolve, reject) => {
```
katat, line 181: I would suggest encapsulating these logics under a return operator into a function as it is quite long. Encapsulating in a function makes this code more readable.

packages/neuron-wallet/src/env.ts
Keith-CY: I prefer to keep the keyword which is more clear.

`.env.local` and `.env.development.local` are both listed in the `.gitignore` so it might be considered by someone.

katat:
> The TypeORM doesn't support a shared in-memory sqlite database which blocks more tests. So I prefer to do the tests manually for now and try to find a way better than using a temporary on-disk database.

Hmm weird. The tests with TypeORM on sqlite in-memory database seem working fine on my side. Maybe it is caused by a dev environment issue 🤔

Keith-CY:
> > The TypeORM doesn't support a shared in-memory sqlite database which blocks more tests. So I prefer to do the tests manually for now and try to find a way better than using a temporary on-disk database.
> 
> Hmm weird. The tests with TypeORM on sqlite in-memory database seem working fine on my side. Maybe it is caused by a dev environment issue 🤔

TypeORM works with a pure in-memory database. I'm using a shared in-memory database, which uses `:memory:?cache=shared`

katat:
Could you please clarify a bit more on why it needs to be configured with `?cache=shared`? I think by default our test setup doesn't include this parameter, right?

packages/neuron-wallet/src/utils/export-history.ts
Keith-CY: I'll commit a refactor in this PR.

Keith-CY: The refactor is committed: https://github.com/nervosnetwork/neuron/pull/1723/commits/1479d87e17bb18dbd8f3264a1e2bd4ea10a17353

Keith-CY:
> Could you please clarify a bit more on why it needs to be configured with `?cache=shared`? I think by default our test setup doesn't include this parameter, right?

According to the doc:

>Every :memory: database is distinct from every other. So, opening two database connections each with the filename ":memory:" will create two independent in-memory databases.

And I'm using two connections in tests(one to set data before each case and the other one to load data for verification).

Keith-CY:
Finally, I use a temporary file as a database in tests in this PR and I'll keep an eye on finding a better solution. @katat @kellyshang

katat:
> And I'm using two connections in tests(one to set data before each case and the other one to load data for verification).

I think you don't have to create more than one connection in a test suit, in which a connection can be reused. Here is a setup example that might be helpful for you: https://github.com/nervosnetwork/neuron/blob/8b646c4c999cedaacefa50692d88712baef46e4b/packages/neuron-wallet/tests/services/asset-account-service.test.ts#L59

Keep in mind that the DB test setup needs to take care of the db connection `init`, `reset data`, and finally `close`.

packages/neuron-wallet/src/database/chain/ormconfig.ts
```diff
@@ -33,17 +33,18 @@ import { RemoveAssetAccountWalletID1589273902050 } from './migrations/1589273902
 
 export const CONNECTION_NOT_FOUND_NAME = 'ConnectionNotFoundError'
 
-const dbPath = (name: string): string => {
+export const dbPath = (name: string): string => {
```
katat, line 36: I guess these changes are mainly meant for the convenience of your test suit in this PR. 

Let's try to resolve the issue in the context of the test suit and **avoid** these changes in this file.

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -1,110 +1,255 @@
...
+
+  const serializedlockHashList = [...lockHashList, ...anyoneCanPayLockHashList].map(l => `'${l}'`).join(`,`)
+  const { sudt } = new AssetAccountInfo().infos
+
+  const writeStream = fs.createWriteStream(filePath)
+  writeStream.write(
+    `${headers.map(label => i18n.t(`export-transactions.column.${label}`))}\n`
+  )
+
+  const db = new sqlite3.Database(dbPath, sqlite3.OPEN_READONLY)
```
katat, line 96: Uses TypeORM

katat:
I think there are two major issues need to be resolved in this PR:
1. Avoid using node-sqlite3 directly as we are already using TypeORM. Sqlite related operations should be encapsulated by TypeORM.
2. The tests should be all relying on the 'memory' sqlite, which is much faster and doesn't need to mess with the file system.

packages/neuron-wallet/src/database/chain/ormconfig.ts
Keith-CY: > I guess these changes are mainly meant for the convenience of your test suit in this PR.
> 
> Let's try to resolve the issue in the context of the test suit and **avoid** these changes in this file.

The fact is that the old one is misleading. I expect the `dbPath` to return the real dbPath but it returns `cell-0x1234.sqlite` instead of `:memory:` in tests. Because of its value is discarded in the test: https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/database/chain/ormconfig.ts#L43

Keith-CY: And that why I've to use 
```
const connection = getConnection()
const dbPath = connection.options.database as string`
```
in the code.

Keith-CY:
Now the export history module is using `AddressesService` and `TransactionService` to load data @katat 
And please test it again @kellyshang

packages/neuron-wallet/tests/utils/export-history/index.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      tx.outputs.forEach((o, i) => o.daoData = (fixture.transaction.outputs[i] as any).daoData)
+      await TransactionPersistor.saveFetchTx(tx)
+    }
+  })
+
+  afterAll(() => {
+    return getConnection().close()
+  })
+
+  it.skip('Test export to csv', async () => {
```
katat, line 27: tests?

PR #1733, opened 2020-06-19
Add tests of get all by addresses

Keith-CY:


packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -66,23 +66,38 @@ export class TransactionsService {
...
+              SELECT output.transactionHash FROM output WHERE output.lockHash in (:...lockHashes)
+              UNION
+              SELECT input.transactionHash FROM input WHERE input.lockHash in (:...lockHashes)
+            )
+          `,
+          { lockHashes }
+        )
+          .orderBy('tx.timestamp', 'DESC')
+          .getRawMany().then(txs => txs.map(tx => tx.txHash))
+        console.log({
```
katat, line 91: I guess we won't need this log to be merged.

packages/neuron-wallet/tests/services/tx/transaction-persistor.test.ts
```diff
@@ -1,118 +1,11 @@
...
-import CellDep, { DepType } from "../../../src/models/chain/cell-dep"
-import OutPoint from "../../../src/models/chain/out-point"
-import Input from "../../../src/models/chain/input"
-import Script, { ScriptHashType } from "../../../src/models/chain/script"
-import Output from "../../../src/models/chain/output"
 import { TransactionPersistor } from "../../../src/services/tx"
 import initConnection from "../../../src/database/chain/ormconfig"
 import TransactionEntity from "../../../src/database/chain/entities/transaction"
 import { getConnection } from "typeorm"
+import { loadTransactions } from '../../setupAndTeardown'
```
katat, line 6: Nice to see these test fixtures are located in one place.

Just one suggestion. Would it be more concise to name this file as `fixtures`, `setups` or `utils`, instead of `setupAndTeardown` since it doesn't seem to include the functions for teardowns?

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
```diff
@@ -0,0 +1,243 @@
...
+    return closeConnection()
+  })
+
+  describe('Test filter search type', () => {
+    it('Should return empty', () => {
+      const actual = TransactionService.filterSearchType('')
+      expect(actual).toBe(SearchType.Empty)
+    })
+
+    it('Should return address', () => {
```
katat, line 27: Most of these tests are having both `condition (when)` and `then` located inside a `it()`.
For better readability and reusability of the preconditions for the different tests, how about we organize the tests in a bit more expressive way, such as 
```
describe('when', () => {
  beforeEach(() => {
    //precondition logics
  })
  it('then', () => {
    //assert logics
  })
})
```

packages/neuron-wallet/src/services/tx/transaction-service.ts
Keith-CY: 😢 The eslint didn't hold it. I'll fix it.

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
```diff
@@ -0,0 +1,243 @@
+import os from 'os'
```
katat, line 1: How about we have a file naming standard to differentiate between the unit tests and integration tests?
Take integration tests for example, we name the test file as `*.intg.test.ts`. If the file contains unit tests, we name as `*.test.ts`.

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
```diff
@@ -0,0 +1,243 @@
...
+    it('Should return empty', async () => {
+      const lockHash = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee'
+      const lockHashList = new Set([lockHash])
+      const status = new Set([TransactionStatus.Success])
+      const actual = await TransactionService.getCountByLockHashesAndStatus(lockHashList, status)
+      expect(actual).toEqual(new Map())
+    })
+  })
+
+  describe('Test update transaction description', () => {
```
katat, line 80: If it is mainly for testing a function, how about we name the scope as `describe('#updateDescription')`, since the function name is expressive enough and it is easier to locate the tests for a specific function?

packages/neuron-wallet/tests/services/tx/transaction-persistor.test.ts
Keith-CY: The method `closeConnection` in the `setupAndTeardown` clears the database so I'd like to keep the name of `setupAndTeardown`. But in this case, the transactions can be loaded directly from the fixture. I'll rename it.

packages/neuron-wallet/src/services/tx/transaction-service.ts
Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1733/commits/1c3156d84c97dae96e17ab871b27a5bc15e3794c

packages/neuron-wallet/tests/services/tx/transaction-persistor.test.ts
Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1733/commits/b9b17e784713afb0d72ab7c7e19c24b97a61fc3b

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1733/commits/00d73419d36f406d3eed6753679f1de1b07055db

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1733/commits/00d73419d36f406d3eed6753679f1de1b07055db

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
Keith-CY: Let's do this systematically.

packages/neuron-wallet/tests/setupAndTeardown/index.ts
```diff
@@ -1,7 +1,6 @@
 import initDB from "../../src/database/chain/ormconfig"
 import { getConnection } from 'typeorm'
-import transactions from './transactions'
-import accounts from './accounts'
+// import accounts from './accounts.fixture'
```
katat: Remove this comment if it is dead code.

packages/neuron-wallet/tests/setupAndTeardown/index.ts
```diff
@@ -1,6 +1,5 @@
 import initDB from "../../src/database/chain/ormconfig"
 import { getConnection } from 'typeorm'
-// import accounts from './accounts.fixture'
```
katat: Remove this dead code?

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
katat: Maybe I miss something. The `when` and `then` are still within a `it()` in this new commit.

Keith-CY: Do you mean the arrangement like

```
describe("blah blah", () => {
  describe("When blah blah", () => {
    const TX_HASH = '...'

    beforeEach(() => {
      setup()
    })

    it("Should blah blah", () => {
    }
  }
}
```

packages/neuron-wallet/tests/setupAndTeardown/index.ts
Keith-CY: I think they've been removed in the following commits.

packages/neuron-wallet/tests/services/tx/transaction-service.test.ts
katat: Yes

Keith-CY: Could you show me some real-world examples? Because most of the libraries I've read follow the same style of the assertion message that I think `it("Should blah blah when/if", () => {}` is more general.

katat: This is the spec style of writing tests. I think this article makes a good introduction to the benefits of adopting the `spec` style in a comparison with the traditional `test` style: https://www.bignerdranch.com/blog/why-do-javascript-test-frameworks-use-describe-and-beforeeach/

This is a real-world example (I might be biased since this is a project I participated): https://github.com/hubiinetwork/nahmii-sdk/blob/master/lib/wallet/wallet.spec.js

Keith-CY: I'm still confused. According to the article and examples, should the code above be changed into the following?

```js
describe("blah blah", () => {
  let txHash

  describe("When tx hash is valid", () => {
    beforeEach(() => {
      txHash = '0x...'
    }
    it("Should blah blah", () => {})
  })

  describe("When tx hash is invalid", () => {
    beforeEach(() => {
      txHash = 'invalid tx hash'
    }
    it("Should blah blah", () => {})
  })
})
```

katat: Yes, I think the example you outlined is now much easier to comprehend and to reuse the setups in a spec group. Feel free to let me know if there are still bits unclear.

Keith-CY: It's clear but not convincing enough. A lot of libraries I've read, e.g. `lodash`, `ramda`, etc. don't have the redundant nesting, even the `jest` is also using the `should blah blah when blah blah` style assertion message. I prefer the general one.

Keith-CY: And I'm not sure how to express 2 booleans with the `describe(when)` style, should it look like the following?
```js
describe(`when x is true, y is true`, () => {
  beforeEach(() => {
    x = true
    y = true
  })
})

describe(`when x is true, y is false`, () => {
  beforeEach(() => {
    x = true
    y = false
  })
})

describe(`when x is false, y is true`, () => {
  beforeEach(() => {
    x = false
    y = true
  })
})

describe(`when x is false, y is false`, () => {
  beforeEach(() => {
    x = false
    y = false
  })
})
```

Keith-CY: 🤔 Well, I think it’s not a serious problem, I’ll update the style tomorrow. But we need more time to update other tests to keep the style consistent.

katat: There are two alternatives to do the same tests in your example.

`jest` or other popular testing frameworks, such as `mocha`, supports nested test suits. Nested test suits make it straightforward to extend the existing tests and for the new tests to reuse the setups in the parent test suit. For example:
```
describe(`when x is true`, () => {
  beforeEach(() => {
    x = true
  })
  describe('when y is true', () => {
    beforeEach(() => {
      y = true
    })
    it()
  })
  describe('when y is false', () => {
    beforeEach(() => {
      y = false
    })
    it()
  })
})
```

The other way is to create dynamic tests using a `for` loop. For example: 
```
[
  [true, true, expected],
  [true, false, expected],
  [false, true, expected],
  [false, false, expected]
].forEach(([_y, _x, expected]) => {
  describe(`when y = ${y} and x = ${x}`, () => {
    let x
    let y
    beforeEach(() => {
      x = _x
      y = _y
    })
    it('should', () => {})
  })
})
```

Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1733/commits/37d656e18627748973fee2915c2c1bbf4ab74790

PR #1737, opened 2020-06-25
feat: refactor validators in ui

Keith-CY:


packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -100,6 +103,8 @@ const updateTransactionWith = (generator: typeof generateTx | typeof generateSen
         setErrorMessage(err.message)
         return undefined
       })
+  } catch {
```
katat, line 106: Are there any potential side effects if this error is silenced?

Keith-CY: It has the same functionality before the update. The `catch` here is similar to
```js
const onSubmit = () => {
  if (disabled) {
    return
  }
}
```

PR #1738, opened 2020-06-26
Add test of export history

Keith-CY:


packages/neuron-wallet/tests/utils/export-history/index.test.ts
```diff
@@ -1,29 +1,147 @@
...
+    return saveTransactions(transactions)
+  })
+
+  afterEach(() => {
+    try {
+      fs.unlinkSync(stubProvider.filePath)
+    } catch {
+      // ignore
     }
+    stubProvider.walletID = WALLET_ID
```
katat, line 36: The alternative way to do this is initializing the stubbed provider in the `beforeEach`, so you don't need to reset the data in `afterEach`, for example:
```
let stubProvider

  beforeAll(() => {
    stubProvider = {
      walletID: WALLET_ID,
      chainType: MAINNET_CHAIN_TYPE,
      filePath: FILE_PATH,
    }
  })
```

packages/neuron-wallet/tests/utils/export-history/index.test.ts
```diff
@@ -1,29 +1,147 @@
...
+    })
   })
 
-  it.skip('Test export to csv', async () => {
+  describe('When file exists', () => {
+    const originalExistsSync = fs.existsSync
+    const originalUnlinkSync = fs.unlinkSync
+
+    afterAll(() => {
+      fs.existsSync = originalExistsSync
```
katat, line 77: The cleaner way to mock this is via the `jest.doMock`, which can be used to mock the `fs` dependency when loading `exportHistory`. So that we don't need to restore the functions by ourselves. 

https://jestjs.io/docs/en/jest-object#jestdomockmodulename-factory-options

packages/neuron-wallet/tests/utils/export-history/index.test.ts
```diff
@@ -1,29 +1,147 @@
...
+      it('Should export table header without sudt column', async () => {
+        expect.assertions(2)
+        const totalCount = await exportHistory(stubProvider)
+        const actual = fs.readFileSync(stubProvider.filePath, 'utf8')
+        expect(totalCount).toBe(0)
+        expect(actual).toBe('Time,Block Number,Transaction Hash,Transaction Type,CKB Amount,Description\n')
+      })
+    })
+
+    describe('When it\'s Mainnet', () => {
```
katat, line 109: I guess the wording here is `when it's testnet`

packages/neuron-wallet/tests/utils/export-history/index.test.ts
Keith-CY: I changed the code into
```js
describe('When file exists', () => {
    let mockFs: any
    beforeEach(() => {
      mockFs = jest.doMock('fs', () => ({
        __esModule: true,
        default: {
          existsSync: jest.fn().mockReturnValue(true),
          unlinkSync: jest.fn().mockReturnValue(undefined),
        },
        existsSync: jest.fn().mockReturnValue(true),
        unlinkSync: jest.fn().mockReturnValue(undefined),
      }))
    })

    it('Should remove file first', async () => {
      expect.assertions(2)
      await exportHistory(stubProvider)
      expect(mockFs.existsSync).toHaveBeenCalledWith(stubProvider.filePath)
      expect(mockFs.unlinkSync).toHaveBeenCalledWith(stubProvider.filePath)
    })
})
```
But the tests failed and `mockFs.existsSync` is `undefined`

Is there anything missing?

packages/neuron-wallet/tests/utils/export-history/index.test.ts
Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1738/commits/a4d8f07350da45d47b237333eeab458e26822346

packages/neuron-wallet/tests/utils/export-history/index.test.ts
Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1738/commits/3ee84eca552ac3b7e467f3823b8c5a8b76766ac8

packages/neuron-wallet/tests/utils/export-history/index.test.ts
Keith-CY: And I've tried
```js
describe('When file exists', () => {
    jest.doMock('fs', () => {
      return {
        existsSync: jest.fn().mockReturnValue(true),
        unlinkSync: jest.fn().mockReturnValue(undefined)
      }
    })

    it('Should remove file first', async () => {
      expect.assertions(2)
      await exportHistory(stubProvider)
      expect(fs.existsSync).toHaveBeenCalledWith(stubProvider.filePath)
      expect(fs.unlinkSync).toHaveBeenCalledWith(stubProvider.filePath)
    })
```
Doesn't work either, and the following tests are broken.

katat: I think it is because the `exportHistory` has to be loaded after the `jest.doMock`.
Here is an example: https://github.com/nervosnetwork/neuron/blob/72fdf037fb5898c81222e24efde9633dfe958c9c/packages/neuron-wallet/tests/block-sync-render/queue.intg.test.ts#L180

Keith-CY: Do I need to import it again each time I define the mock? It seems quite weird... I'll have a try.

katat: You should just need to import once. Normally you defer the import until all the necessary mocks via `jest.doMock` has been set up.

Keith-CY:
```js
  describe('When file exists', () => {
    beforeAll(() => {
      jest.doMock('fs', () => ({
        existsSync: jest.fn().mockReturnValue(true),
        unlinkSync: jest.fn().mockReturnValue(undefined)
      }))
    })

    it('Should remove file first', async () => {
      
      expect.assertions(2)
      const fn = require('../../../src/utils/export-history').default
      const mockFs = require('fs')
      await fn(stubProvider)
      expect(mockFs.existsSync).toHaveBeenCalledWith(stubProvider.filePath)
      expect(mockFs.unlinkSync).toHaveBeenCalledWith(stubProvider.filePath)
    })
  })
```
Still not work, could you do me a favor to fix the test with `doMock`?

And one more question, if I want to define a series of returned value of the mocked function for each case, should I define it at the top of test block with `mockReturnValueOne(A).mockReturnValue(B)...` and then import the module to test instead of defining the mock in each test case independently?

katat: Sure, no worries. I will take a look at this later today and try to come up with some examples of doing the mocks for this case.

katat: Actually no needs to use `jest.doMock` here for fs module. `jest.spyOn` is sufficient. Please see PR https://github.com/nervosnetwork/neuron/pull/1741

Keith-CY: `spyOn` does the same work here and can be restored with `jest.restoreAllMocks` manually, but what's the benefit of using `spyOn` instead of `fn`?

IMO, `spyOn` is used to assert the behavior of calls, and expected to return the real result instead of a mock result. And the docs also suggests that use `jest.spyOn(object, methodName).mockImplementation(() => customImplementation)` or `object[methodName] = jest.fn(() => customImplementation)` to skip the real invocation.

katat: > spyOn does the same work here and can be restored with jest.restoreAllMocks manually, but what's the benefit of using spyOn instead of fn?

With `spyOn`, you don't need to manually keep the original implementations at somewhere such as
```
      fs.existsSync = originalExistsSync
      fs.unlinkSync = originalUnlinkSync
```

> And the docs also suggests that use jest.spyOn(object, methodName).mockImplementation(() => customImplementation) or object[methodName] = jest.fn(() => customImplementation) to skip the real invocation.

What the docs don't mention is the return of `spyOn` is actually the return of `jest.fn()`, which you can use for both assertions of expected calls and mock return values.

Keith-CY: It's indeed a sugar of `fn` but the semantic is to call the real method which is emphasized in the docs, I think we should follow the convention.

BTW, I'd like to know the benefit of using `spyOn` instead of `fn` in this case. From my perspective, the only difference is the clean after test case where `spyOn` uses `restore` and `fn` uses assignment.

Keith-CY: I've checked some articles, it's ok to use `spyOn` here.

katat: What I see in your test, it is intended to mock values for the `fs` functions instead of calling through the real `fs` functions, correct? 

AFAIK, if we don't use `spyOn` here, then it has to do the setup you did, which could be quite cumbersome when it comes to maintenance IMHO because managing the original functions at somewhere like this could be a coding overhead for other developers. With these mocks all managed by `jest`, it is much easier to comprehend as we all have the common practice in terms of how to do the mocks.
```
fs.existsSync = originalExistsSync
fs.unlinkSync = originalUnlinkSync
```

PR #1743, opened 2020-06-29
Add notification of rebuild cache

Keith-CY:
image


packages/neuron-ui/src/locales/en.json
```diff
@@ -335,6 +336,7 @@
       "full-addr-info": "Additional software or dApp interface might be required to operate the assets on this address. Please double check its validity.",
       "decimal-range": "Decimal should be an interger of {{range}}(inclusive)",
       "experimental-message": "This is an experimental feature, it could change at any time. Please use with caution.",
+      "rebuild-sync": "To achieve better user experience, Neuron has adopted a brand new index mechanism, which demands a rebuilding of cache database (estimated 30 ~ 60min).\nSorry for the inconvenience.",
```
katat, line 339: Wording suggestion:
"a brand new index mechanism" => "a brand new indexing mechanism"

katat: "To achieve better user experience" => "For better user experience"

katat: "which demands" => "which requires"

Keith-CY: This kind of update should be confirmed in the PRD issue..

packages/neuron-ui/src/utils/is.ts
```diff
@@ -9,3 +9,13 @@ export const isMainnet = (networks: Readonly<State.Network[]>, networkID: string
 export const isSuccessResponse = (res: Pick<ControllerResponse, 'status'>): res is SuccessFromController => {
   return res.status === ResponseCode.SUCCESS
 }
+
+export const isReadyByVersion = (targetVersion: number, lastVersion: number | null) => {
+  if (lastVersion === null || Number.isNaN(+lastVersion)) {
```
katat, line 14: is `+lastVersion` always a number?

Keith-CY: Yes, it consists of a major version and a minor version.

packages/neuron-ui/src/locales/en.json
katat: That is why we need PR review, right? 😛 
They are just some suggestions. The original version might be better.

packages/neuron-ui/src/utils/is.ts
katat: If so, I am concerned about the `Number.isNaN(+lastVersion)` is always `true`, which makes this `if` check meaningless.

packages/neuron-ui/src/locales/en.json
Keith-CY: > That is why we need PR review, right? 😛
> They are just some suggestions. The original version might be better.

😣 Actually I don't make a judgement on the expression, I just want to make a consensus among us. Otherwise, the content of notification should fail in tests.

packages/neuron-ui/src/utils/is.ts
Keith-CY: Sure, it's not necessary, I'll remove it.

Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1743/commits/2e9d8493772231751a4ac1c2e810865374036335

PR #1747, opened 2020-06-29
Install clang

Keith-CY:


azure-pipelines.yml
```diff
@@ -77,17 +77,21 @@ stages:
...
     - script: |
-        yarn bootstrap
-        set CI=false
-        yarn build
-      name: Build
-    - script: |
-        set CI=true
-        yarn test
-      name: Test
+        powershell -Command "$ProgressPreference = 'SilentlyContinue'; iwr -outf %TEMP%\LLVM-8.0.0-win64.exe https://rust-lang-ci2.s3.amazonaws.com/rust-ci-mirror/LLVM-8.0.0-win64.exe"
```
katat, line 81: I am wondering if it makes sense to add this new script above the original script of `install lerna`, and keep the rest of the scripts as it is without removing them adding the `pwsh` script.

Keith-CY: I'll restore them if this script works.

Keith-CY:
Seems good, I will pick this to the target branch.

katat:
🎆  🍾

PR #1756, opened 2020-07-01
caching yarn bootstrap to speed up the CI tasks

katat:
* Uses azure pipeline cache to speed up the CI builds

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,9 +39,6 @@ stages:
       inputs:
         key: yarn bootstrap | $(Agent.OS) | yarn.lock
         path: $(YARN_CACHE_FOLDER)
-        restoreKeys: |
-          yarn bootstrap | $(Agent.OS)
-          yarn bootstrap
       displayName: Cache Yarn packages
     - script: |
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -12278,7 +12278,7 @@ jest-watcher@^24.3.0, jest-watcher@^24.9.0:
     jest-util "^24.9.0"
     string-length "^2.0.0"
 
-jest-when@^2.7.2:
+jest-when@2.7.2:
```
Keith-CY, line 12281: 🤔 Seems that the `yarn.lock` validation in the CI doesn't work. The CI should fail if the lock file is not matched with the package.json.

Need further debug

azure-pipelines.yml
```diff
@@ -156,6 +229,8 @@ stages:
       inputs:
         versionSpec: 12.x
       displayName: 'Install Node.js'
+    - script: |
```
Keith-CY, line 232: Does `macos`(the VM in this job) need this?

katat:
Close this since the latest version of lumos provides prebuilt binaries for electron 7

PR #1757, opened 2020-07-02
fix: add missing error codes and messages

Keith-CY:
image


packages/neuron-wallet/src/exceptions/address.ts
```diff
@@ -1,18 +1,25 @@
 import i18n from 'locales/i18n'
 
 export class InvalidAddress extends Error {
+  public code = 102
+  public i18n = {
```
katat, line 5: Does this variable name `i18n` conflict with the import `i18n`?

Keith-CY: It's the convention used in UI, I'll change the imported name.
But I'm not sure if it's reasonable to include i18n in the custom error class.

Keith-CY: Maybe [this](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/api.ts#L389) is a better aspect to handle the i18n

katat: Actually I am still not quite sure how this property `i18n` will be used in other places. If it has not yet been referenced in the code, maybe we can remove this property for now and see if there is a better alternative once the code started referencing this variable.

Keith-CY: Updated in https://github.com/nervosnetwork/neuron/pull/1757/commits/dec1417f15d3e8d47a4364025ce3f7c61861e915

packages/neuron-wallet/src/exceptions/address.ts
```diff
@@ -1,27 +1,34 @@
-import i18n from 'locales/i18n'
+import { t } from 'locales/i18n'
```
katat, line 1: I would suggest make this variable bit more expressive. Maybe `import {t as translate} from 'locals/i18n'` is a good alternative?

Keith-CY: > I would suggest make this variable bit more expressive. Maybe `import {t as translate} from 'locals/i18n'` is a good alternative?

The `t` is the standard name of this function, and its type is `TFunction` so I think it's clear enough for i18n users.

Keith-CY: 🤔 seems that I don't need to export it from `locals/i18n`, it can be imported from `i18next` directly.

packages/neuron-wallet/src/exceptions/address.ts
Keith-CY: The `i18n` property is not used in UI for now, but if some exception happens and UI doesn't handle it, the extra info will be used by i18n in UI automatically.

katat: I think I kind of got it now.

packages/neuron-wallet/src/exceptions/address.ts
katat: If `i18next` can be used in all the places where `locals/i18n` is used, then I would suggest we should only use `i18next`. Using both `i18n` and `i18next` might be confusing at some point.

Keith-CY: I'm not sure why the `i18n` is exported from `locals/i18n`.

AFAIK the `locals/i18n` is just used to set up configuration and will take effect by imported in the entry. And then we can use `i18n` exported from `i18next` in other places. So I think `import i18n from 'locals/i18n'` can be replaced by `import i18n from 'i18next'` globally.

I'll check it later.

katat: Sounds good

PR #1761, opened 2020-07-03
fix: fix typo in i18n[skip ci]

Keith-CY:


Keith-CY:
@kellyshang The label of `rebuild index` is fixed.

kellyshang:
> @kellyshang The label of `rebuild index` is fixed.

Got it.

PR #1762, opened 2020-07-03
Rc/v0.32.0-beta.1

kellyshang:
Beta version, scheduled to be released on July 4th.


CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
...
+* Detect vcredist on Windows when user launches Neuron.
+* Enable the button of exporting transaction history.
+* Support two modes for clearing caches: Refresh cache / Fully rebuild index.
+* UI tune for pop-up windows.
+* Share token info across different wallets.
+* Add a link of `Deposit rules` on `Nervos DAO` page.
+
+### Syncing
+
+It has adopt a brand new indexing mechanism, which can significantly speed up the cache indexing process. The caches will be rebuilt for the first time, but it should be much faster than the previous versions (up to 3-4x on macOS, Windows and Linux).
```
Keith-CY, line 20:
```suggestion
It has adopted a brand new indexing mechanism, which can significantly speed up the cache indexing process. The caches will be rebuilt for the first time, but it should be much faster than the previous versions (up to 3-4x on macOS, Windows and Linux).
```

PR #1765, opened 2020-07-03
ci: bundle win-ver on windows vm

Keith-CY:
set the vmImage for bundling Neuron-win-ver `vs2017-win2016`


Keith-CY:
The attachments now in the release draft are uploaded by this PR and the one for windows is verified on Windows Server 2019.

PR #1768, opened 2020-07-06
fix: ignore null error on writable stream closing

Keith-CY:


packages/neuron-wallet/jest.config.js
```diff
@@ -18,5 +18,10 @@ module.exports = {
     "json",
     "node"
   ],
-  setupFiles: ['<rootDir>/tests/setup.ts']
+  setupFiles: ['<rootDir>/tests/setup.ts'],
```
katat, line 21: We currently have the `tests/setup.ts` to setup the global mocks. It would be better to put all the global mocks in the same place.

packages/neuron-wallet/tests/controllers/export-debug.test.ts
```diff
@@ -0,0 +1,134 @@
...
+jest.mock('fs', () => {
+  return {
+    createWriteStream: () => null
+  }
+})
+
+jest.mock('../../src/utils/logger', () => ({
+  error: console.error,
+  transports: {
+    file: {
```
katat, line 24: Please see my comments above.

packages/neuron-wallet/jest.config.js
Keith-CY: The `queue.test.ts` fails all the time and `queue.intg.test.ts` fails sometimes when the `electron` is mocked globally.
All most all tests fail when `fs` is mocked globally.
Should I fix them all?

katat: Ok, I see. Sounds like if `electron` is mocked globally, then there could be conflicts with the mocks in some of the tests.

If this is the case, then I would suggest the mocks for `electron` shouldn't be controlled within the individual tests instead of globally.

Keith-CY:
@kellyshang The error on unzipping the debug info should be fixed with this PR, please have a test.

kellyshang:
> @kellyshang The error on unzipping the debug info should be fixed with this PR, please have a test.

Got it.

PR #1769, opened 2020-07-06
refactor: toggle sudt nav item based on cached chain type

Keith-CY:
Avoid extra network requests and is more sensitive to toggle the nav item

By this PR the sUDT nav item will be updated on switching network even they are offline.


Keith-CY:
With this fix, the `sUDT` nav item should update even it's offline, please have a test @kellyshang

PR #1775, opened 2020-07-09
Backend fixes for 0.32-beta.1

katat:
* Upgrade to lumos indexer 0.4.8
* Refactors according to the API changes between lumos indexer 0.3.0 and 0.4.8
* Prevent unnecessary index rebuild after imported a wallet
* Fix the incorrect balance issue caused by the silenced errors when saving transactions
* Update yarn bootstrap script to deprecate building for lumos indexer
* Reload sync block render after deleted a wallet

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "packages/*"
   ],
   "scripts": {
-    "bootstrap": "lerna bootstrap && lerna link",
+    "bootstrap": "LUMOS_NODE_RUNTIME=electron LUMOS_NODE_RUNTIME_VERSION=7.1.4 lerna bootstrap && lerna link",
     "start:ui": "cd packages/neuron-ui && yarn run start",
     "start:wallet": "cd packages/neuron-wallet && yarn run start:dev",
     "start": "concurrently \"cross-env BROWSER=none yarn run start:ui\" \"wait-on http://localhost:3000 && yarn run start:wallet\"",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "packages/*"
   ],
   "scripts": {
-    "bootstrap": "LUMOS_NODE_RUNTIME=electron LUMOS_NODE_RUNTIME_VERSION=7.1.4 lerna bootstrap && lerna link",
+    "bootstrap": "cross-env LUMOS_NODE_RUNTIME=electron LUMOS_NODE_RUNTIME_VERSION=7.1.4 lerna bootstrap && lerna link",
+    "prebootstrap": "npm i cross-env@5.2.1 --no-save",
     "start:ui": "cd packages/neuron-ui && yarn run start",
     "start:wallet": "cd packages/neuron-wallet && yarn run start:dev",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -22,7 +22,8 @@
     "packages/*"
   ],
   "scripts": {
-    "bootstrap": "lerna bootstrap && lerna link",
+    "bootstrap": "cross-env LUMOS_NODE_RUNTIME=electron LUMOS_NODE_RUNTIME_VERSION=7.1.4 lerna bootstrap && lerna link",
+    "prebootstrap": "npm i cross-env@5.2.1 --no-save",
```
Keith-CY, line 26: `npx` can do this work without another script.

packages/neuron-wallet/tests/block-sync-render/indexer-cache-service.intg.test.ts
```diff
@@ -1,7 +1,7 @@
 import { when } from 'jest-when'
 import { getConnection } from 'typeorm'
 import { initConnection } from '../../src/database/chain/ormconfig'
-import IndexerCacheService from '../../src/block-sync-renderer/sync/indexer-cache-service';
+// import IndexerCacheService from '../../src/block-sync-renderer/sync/indexer-cache-service';
```
Keith-CY, line 4: Useless comment

package.json
katat: I think it is a nice approach to go with `npx` in this case.

I had a try with `npx -c 'cross-env...` in the script `bootstrap`, but it doesn't seem to work. The child script `rebuild:nativemodules`  of `postinstall` throws an error at the end.

Please feel free to push the code or suggest if you know how to achieve this with `npx`.

Keith-CY: Don't use `-c` with `npx`, I got an error with that kind of usage.
And it reminds me that the `env` should be limited in the neuron-wallet, better in the `rebuild:neodeps` script as the following:
```sh
"rebuild:neondeps": "npx cross-env LUMOS_NODE_RUNTIME=electron LUMOS_NODE_RUNTIME_VERSION=7.1.4 electron-build-env neon build -p ../../node_modules/@ckb-lumos/indexer --release"
```

katat: Thanks for the example.

As we are using lumos indexer. 0.4.8, which comes with a prebuilt binary for electron 7, we don't need to build it anymore. The reason we need to have these env variables `LUMOS_NODE_RUNTIME=electron LUMOS_NODE_RUNTIME_VERSION=7.1.4` is to tell the lumos npm installation script to download the electron version of binary instead of the default node version. They need to be passed on to the `lerna bootstrap`, which in turn passes on to the lumos indexer npm installation script. 

To make these env variables compatible with different OS, it uses `cross-env`. In a clean setup, `cross-env` executable won't be available when running `yarn bootstrap`, thus we need to figure out a way to make sure `cross-env` is installed. I had another try with the `npx` without `-c` so that we don't need `prebootstrap` script, and it seems working now.

PR #1795, opened 2020-07-31
upgrade elliptic to 6.5.3

katat:
```
$ yarn why elliptic
yarn why v1.22.4
[1/4] 🤔  Why do we have the module "elliptic"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "elliptic@6.5.2"
info Has been hoisted to "elliptic"
info Reasons this module exists
   - "workspace-aggregator-f458ce74-0352-4ac0-9021-0c53e697e0b2" depends on it
   - Hoisted from "_project_#neuron-wallet#@nervosnetwork#ckb-sdk-utils#elliptic"
   - Hoisted from "_project_#webpack#node-libs-browser#crypto-browserify#browserify-sign#elliptic"
   - Hoisted from "_project_#webpack#node-libs-browser#crypto-browserify#create-ecdh#elliptic"
info Disk size without dependencies: "172KB"
info Disk size with unique dependencies: "548KB"
info Disk size with transitive dependencies: "548KB"
info Number of shared dependencies: 7
=> Found "neuron-wallet#elliptic@6.5.1"
info This module exists because "_project_#neuron-wallet" depends on it.
info Disk size without dependencies: "172KB"
info Disk size with unique dependencies: "548KB"
info Disk size with transitive dependencies: "548KB"
info Number of shared dependencies: 7
```

@Keith-CY The same dependency on `ckb-sdk-utils` may also need to be updated to the version higher than `6.5.2` due to the dependabot alert.

Also I am curious why the `yarn.lock` has the webpack dependency, which we don't really use for the building process at the moment?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2645,37 +2645,42 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
-"@nervosnetwork/ckb-sdk-core@0.32.0":
-  version "0.32.0"
-  resolved "https://registry.yarnpkg.com/@nervosnetwork/ckb-sdk-core/-/ckb-sdk-core-0.32.0.tgz#6609b14e7daa5e61bb55ea361581dd634c65b2d8"
-  integrity sha512-h7sAT2XosXoq6SCLdmu5cxYpJU/7+9LyMcxGLDc81ubpbq1uuayaT8XBCBW7129ACoMxeE96R1Ac4kgcnRNYqQ==
+"@nervosnetwork/ckb-sdk-core@0.34.0":
+  version "0.34.0"
+  resolved "https://registry.yarnpkg.com/@nervosnetwork/ckb-sdk-core/-/ckb-sdk-core-0.34.0.tgz#d68f5ce98cb419457327dd6168a40f3190186cb5"
+  integrity sha512-+ZDAZ1R/ILHyQsZRbG5qHir62gNZdfZCkGoU8vgTrKGl7hPHwP7L0+8bgyO1tVxg5oO/UGTRur+Egs4XeiwFOg==
   dependencies:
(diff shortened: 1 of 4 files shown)


yuche:
Webpack is a dependency on `react-scripts`, and we use `react-scripts` (via `react-app-rewired`) to compile the render process.

Keith-CY:
> @Keith-CY The same dependency on ckb-sdk-utils may also need to be updated to the version higher than 6.5.2 due to the dependabot alert.

Please upgrade the SDK to 0.34 which uses elliptic@6.5.3
https://github.com/nervosnetwork/ckb-sdk-js/blob/develop/packages/ckb-sdk-utils/package.json#L36

katat:
Thanks @Keith-CY . Upgraded SDK to 0.34.

yarn.lock
```diff
@@ -2645,38 +2645,43 @@
...
   dependencies:
-    "@nervosnetwork/ckb-types" "0.32.0"
+    "@nervosnetwork/ckb-types" "0.34.0"
     blake2b-wasm "2.1.0"
-    elliptic "6.5.2"
-    jsbi "3.1.2"
+    elliptic "6.5.3"
+    jsbi "3.1.3"
 
 "@nervosnetwork/ckb-types@0.32.0":
```
Keith-CY, line 2675: Outdated dep.

katat:
Screen Shot 2020-07-31 at 6 00 33 PM
There seems to be a compatibility issue with the latest SDK. 

Since the dependabot automatically created a [PR](https://github.com/nervosnetwork/neuron/pull/1796) for the vulnerability fix. I close this PR now, and we will come back to fix the SDK issue in other PR.

PR #1800, opened 2020-08-05
Upgrade to Electron 9 

yuche:
This PR has updated the Electron version to `v9.1.2`.

The biggest change besides upgrade is moving the sync task from the render thread to the main thread. This is because rendering threads cannot introduce Native Modules in Electron 9 that do not meet the `Context-Aware` condition or are not built with `N-API`.


packages/neuron-wallet/package.json
```diff
@@ -55,6 +55,8 @@
     "sha3": "2.0.7",
     "sqlite3": "4.1.1",
     "subleveldown": "^4.1.4",
+    "worker-threads.js": "1.6.5",
```
Keith-CY, line 58: Is it necessary to introduce a third-party package?

yuche: By using this library, the code for master/child process communication is much clearer than the built-in pub/sub pattern, and the same codebase can be changed by changing a line of code to start `worker thread` or `child_process`.

Keith-CY: > By using this library, the code for master/child process communication is much clearer than the built-in pub/sub pattern, and the same codebase can be changed by changing a line of code to start `worker thread` or `child_process`.

Can we implement it in this project? We are going to reduce deps since they are out of control.

yuche: Sure, we can. I'll remove this dependency and implement it myself.
BTW, How to define a library that is "out of control", is there a guideline?

Keith-CY: > Sure, we can. I'll remove this dependency and implement it myself.
> BTW, How to define a library that is "out of control", is there a guideline?

There're no explicit rules to decide if a dep is acceptable, but the strategy made before is conservative(and some relative rules can be found in wiki/package-dependency-upgrade-strategy in the internal repo).

- The best choice is not to use a third-party package.
- The second one is to use a package having reputations.
- The third one is to use a package that is generally used and implement it ourselves once we can.

The 2nd. and 3rd. are not ideal though since famous packages or widely used packages are also fragile, such as `event-stream` which was actually attacked and had an incredible effect.

yuche: Thank you for your thorough explanation. 👍

packages/neuron-wallet/tests/test-utils.ts
```diff
@@ -1,4 +1,25 @@
+import { EventEmitter } from 'events'
+
 export const flushPromises = () => {
   jest.runAllImmediates()
   return new Promise(setImmediate)
 };
+
+
+export class SyncTask extends EventEmitter {
```
katat, line 9: Maybe we can use `jest.doMock`, so we don't need to manually create this mock class?

packages/neuron-wallet/tests/utils/worker.test.ts
```diff
@@ -0,0 +1,47 @@
+import path from 'path'
+import { expose, spawn, terminate } from '../../src/utils/worker'
+import { fork } from 'child_process'
+
+describe('workers', () => {
+  const noop = (..._: any) => undefined
+  const delay = (ms: number) => new Promise(resolve => setTimeout(resolve, ms))
+
+  it('expose should send object keys as channels to the master process', async () => {
```
katat, line 9: I would suggest using the pattern similar to the other tests in the sync queue module:
```js
describe('#function', () => {
  describe('condition', () => {
    it('success case')
    it('failed case')
  })
})
```

packages/neuron-wallet/tests/test-utils.ts
yuche: Actually, we don't need this mock anymore, as the main process won't import `SyncTask` once the entire sync task has been moved to a child thread.

packages/neuron-wallet/tests/utils/worker.test.ts
yuche: Fixed.

packages/neuron-wallet/package.json
```diff
@@ -18,8 +18,8 @@
   "license": "MIT",
   "scripts": {
     "start": "electron .",
-    "start:dev": "yarn run build && electron .",
-    "build": "ttsc && ncp ./src/block-sync-renderer/index.html ./dist/block-sync-renderer/index.html",
+    "start:dev": "ttsc && electron .",
```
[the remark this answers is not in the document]
yuche: > This modification is unnecessary.

Because we no longer need `. /src/block-sync-renderer/index.html` file, we still need to remove the copy html command. But we do can leave the "start:dev" command unchanged.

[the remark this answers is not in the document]
Keith-CY: Yes, I mean we can keep `yarn build` in `start:dev`

packages/neuron-wallet/src/database/chain/ormconfig.ts
```diff
@@ -75,6 +75,7 @@ const connectOptions = async (genesisBlockHash: string): Promise<SqliteConnectio
       RemoveLiveCell1592781363749,
       AddIndexerTxHashCache1592727615004,
     ],
+    logger: 'simple-console',
```
katat, line 78: I am wondering what is this new config property `logger` for? Is it a new logging format in typeorm?

packages/neuron-wallet/src/database/address/address-dao.ts
```diff
@@ -331,6 +331,13 @@ class AddressStore {
   }
 
   static changed() {
-    AddressDbChangedSubject.getSubject().next("Updated")
+    if (process.send) {
+      process.send({
```
katat, line 335: Could you please explain a bit how this `process.send` works and how it relates to the existing rxjs subjects?

packages/neuron-wallet/src/database/chain/ormconfig.ts
yuche: Default options `advance-logger` parses and highlights sql query, which affects performance.
see: https://github.com/nervosnetwork/neuron/pull/1800/commits/e21e27b13d7bca8ad083bc912479486986306eff
and profile:
https://www.dropbox.com/s/m5ky04e7dl2ts5u/Heap-20200811T000034.heapsnapshot?dl=0
typeorm logger docs:
https://orkhan.gitbook.io/typeorm/docs/logging#changing-default-logger

packages/neuron-wallet/src/database/address/address-dao.ts
yuche: We have subscribed to messages from sub threads in the main process.
https://github.com/nervosnetwork/neuron/blob/eddb4e4b78053ba81288b9bd9416f99b52a4434f/packages/neuron-wallet/src/block-sync- renderer/index.ts#L95-L113

In the original implementation, we can get the Rx Subjects instance of the main process through `remote.require`, but since the `remote module` is no longer available in the child process, we can only communicate through the child process's `process.send`.

packages/neuron-wallet/src/database/chain/ormconfig.ts
katat: Got it. Is there a performance benchmark between these two log mode?

packages/neuron-wallet/src/database/address/address-dao.ts
katat: I would suggest wrapping this global variable `process` into a class such as `MasterProcess`. This makes the variable more expressive and easier to maintain whenever needed, such as when you want to stub the master process in the automated tests or modify the behavior of the function `send`.

packages/neuron-wallet/src/block-sync-renderer/index.ts
```diff
@@ -73,74 +78,73 @@ export const createBlockSyncTask = async (clearIndexerFolder = false) => {
...
-        )
+  syncTask = await spawn<SyncTask>(
+    fork(path.join(__dirname, 'task.js'), [], {
+      env: {
+        fileBasePath: env.fileBasePath
       }
+    })
+  )
+
+  subscribe(syncTask, msg => {
```
katat, line 95: Renaming this local function name to `subscribeToWorkerProcess` would be more expressive.

packages/neuron-wallet/src/block-sync-renderer/task.ts
```diff
@@ -1,37 +1,45 @@
...
+    syncQueue?.stop()
+    syncQueue = null
+  },
+  async start(url: string, genesisHash: string, addressesMetas: AddressInterface[]) {
+    if (syncQueue) {
+      await syncQueue.stopAndWait()
+    }
+
+    // reset `fileBasePath` from master process
+    env.fileBasePath = process.env['fileBasePath'] ?? env.fileBasePath
```
katat, line 28: Where does this `fileBasePath` value come from originally?

packages/neuron-wallet/src/database/address/address-dao.ts
yuche: > I would suggest wrapping this global variable `process` into a class such as `MasterProcess`. This makes the variable more expressive and easier to maintain whenever needed, such as when you want to stub the master process in the automated tests or modify the behavior of the function `send`.

Good idea.

packages/neuron-wallet/src/block-sync-renderer/task.ts
yuche: From master process:
https://github.com/nervosnetwork/neuron/blob/eddb4e4b78053ba81288b9bd9416f99b52a4434f/packages/neuron-wallet/src/block-sync-renderer/index.ts#L87-L93

packages/neuron-wallet/src/database/chain/ormconfig.ts
yuche: > Got it. Is there a performance benchmark between these two log mode?

I just did a simple benchmark on my own, using a query that Neuron frequently warns about with `query-is-slow`:
https://github.com/yuche/typeorm-logger-bench

Benchmark shows that `advance-logger` is 20x-30x slower than `simple-logger`.
#### macos default Terminal:
simpleLogger x 38,057 ops/sec ±0.59% (94 runs sampled)
advanceLogger x 1,198 ops/sec ±1.14% (95 runs sampled)
#### iTerm 2
simpleLogger x 34,657 ops/sec ±6.23% (78 runs sampled)
advanceLogger x 1,033 ops/sec ±1.47% (84 runs sampled)

packages/neuron-wallet/src/block-sync-renderer/task.ts
katat: Will this be changed when it switches to different networks?

packages/neuron-wallet/src/database/chain/ormconfig.ts
katat: Interesting to know 👍

packages/neuron-wallet/src/block-sync-renderer/task.ts
yuche: It won't changed.
`env.fileBasePath` should stay intact when Electron starts, but since there is no `electron.app` or `remote.app` module in the child process, the child process's `env.fileBasePath` is actually a fake value for passing the test.
https://github.com/nervosnetwork/neuron/blob/5a7f78d5ec4d674b035140fe6944e8d4eeab8120/packages/neuron-wallet/src/env.ts#L6-L16

katat (approved):

Keith-CY (approved):

PR #1801, opened 2020-08-05
Optimize renderer process build system

yuche:
This PR does two main things.
1. Compiling TypeScript files with `ts-loader` allows us to use more native TypeScript features, such as `const enum`, and we can update the TypeScript version at will without waiting for Babel Parser support.
2. Set the JavaScript compilation target to Electron 9, which is more accurate than the original target of ES2019.

`code spliting` has not been implemented yet, and needs to be investigated to see how much of a performance benefit this will give us.

As investigate go further, the final implementation may look very different from the draft.

Note that this draft is not yet complete, the current code may not be available and there is still some code that needs to be committed.


yuche:
Since ts-loader and react-app-wired don't work well together, and code spiliting doesn't significantly speed up application opening, this PR is closed.

PR #1802, opened 2020-08-05
Import electron modules in the render process

yuche:
In earlier versions of Electron, we could import the Electron module directly in the renderer process by setting `electron` to `externals` in the Bundler. However, in newer versions of Electron, where `nodeIntegration` is set to `false` by default for security reasons, we can't do that and have to set the global object to `preload.js` in the renderer process.

In this PR, we're modularizing the global object via Webpack's Alias feature, so we can also import the Electron module directly in the renderer process without setting `nodeIntegration` to `true`. This makes our code clearer and type-safe, and the Electron module imported in this way is always accessible in the renderer process. 

Unfortunately some developers will run the renderer process in a browser, so the code to determine if the module is accessible remains:

https://github.com/nervosnetwork/neuron/blob/0dd8dace9a18573eb36bc435c4949755d042b3f4/packages/neuron-ui/src/services/remote/index.ts#L24-L32

### Question

Usually the renderer process are run in the browser because some developers `cd neuron-ui` and then run `yarn start`. 
Can we remind the developer only once that he should run `yarn start` in the root directory when the renderer process code is detected running in the browser, rather than warning in every(to be precise, some functions have warnings and some don't) function？

That way we can get rid of all the code that determines if a Electron module exists or not.


Keith-CY:
We introduced electron modules for convenience in the early stage, now we can move related functions into the main process.

And for this case

> Can we remind the developer only once that he should run yarn start in the root directory when the renderer process code is detected running in the browser, rather than warning in every(to be precise, some functions have warnings and some don't) function？

I see that you are injecting electron modules into `window.electron`, then we can use Proxy to trap properties of `window.electron` and show a warning once `window.electron.xxx`, which is `undefined`, is used.

Keith-CY:
> > We introduced electron modules for convenience in the early stage, now we can move related functions into the main process. And some of them can be replaced by browser APIs.
> 
> I'm not sure if I'm misinterpreting what that means. IMHO, I don't see the benefit of moving Electron modules which are belong to renderer process to the main process, but rather adding complexity to IPC communications. As for using some of the browser APIs as replacement is a good idea, but I think this is something we should leave for the future when we need to move from a C/S architecture to a B/S architecture.
> 
> > I see that you are injecting electron modules into window.electron, then we can use Proxy to trap properties of window.electron and show a warning once window.electron.xxx, which is undefined, is used.
> 
> That's something we can do now.

According to the [declaration](https://github.com/nervosnetwork/neuron/blob/refactor/import-electron-module-in-render-process/packages/neuron-ui/src/types/global/index.d.ts), the following modules are injected

- clipboard: can be replaced by `window.navigator.clipboard`
- remote: can be imported by `require('electron').remote`, actually it's just used to get the `require` module, but the `require` module is injected directly.
- shell: can be replaced by [open in window](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/api.ts#L142)
- require
- nativeImage: can be imported by `require('electron').nativeImage`, actually we can replace it by canvas
- desktopCapture: can be imported by `require('electron').desktopCapture`, actually we can do it in the main process since qrcode scanning may be called outside the window
- ipcRenderer: can be imported by `require('electron').ipcRenderer`

Keith-CY:
When I said we can move some electron module into the main process, I mean that they are declared to be used in the main process

For example, the `MenuItem` is introduced in https://github.com/nervosnetwork/neuron/pull/1802/files#diff-4e970f0c9abcd8f2e2b4c8f56356442aR8, while the document says that it is called in the main process. When it's used in the renderer process directly, an IPC is connected and a remote object is created, which is bug-prone. We have treated similar problems due to remote objects.

packages/neuron-ui/.eslintrc.js
```diff
@@ -26,9 +26,7 @@ module.exports = {
       "exports": "always-multiline",
       "functions": "ignore"
     }],
-    "import/no-extraneous-dependencies": [2, {
-      "devDependencies": true
-    }],
+    "import/no-extraneous-dependencies": "off",
```
Keith-CY, line 29: It's strongly recommended to turn this on

packages/neuron-ui/.eslintrc.js
```diff
@@ -26,9 +26,7 @@ module.exports = {
...
     "no-unused-vars": "off",
```
Keith-CY, line 30: 🤔 and I don't why this is off, it should be on too

packages/neuron-ui/config-overrides.js
```diff
@@ -1,4 +1,7 @@
-module.exports = function override(config, env) {
-  //do stuff with the webpack config...
-  return config;
-}
\ No newline at end of file
+const path = require('path')
+
+module.exports = function override(config) {
+  const webpackCofnig = { ...config }
```
Keith-CY, line 4: It seems meanless here.

packages/neuron-ui/.eslintrc.js
yuche: If we turn this on, we can't introduce electron in the render process because the render process has no dependency on electron.

packages/neuron-ui/config-overrides.js
yuche: Can't get through ESLint without doing this.

packages/neuron-ui/src/services/remote/index.ts
```diff
@@ -1,3 +1,13 @@
+import {
+  ipcRenderer,
+  remote,
+  desktopCapturer,
+  shell,
+  OpenDialogOptions,
+  MenuItemConstructorOptions,
```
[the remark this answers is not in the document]
yuche: `import  type { A, B } from 'example'` will can cause ESLint errors.  But `import  type { A } from 'example'` will do fine.
I think maybe that's a `@typescript-eslint/parser` bug,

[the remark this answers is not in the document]
yuche: image
If importing multiple types under the ESLint configuration of the rendering process raises an error, I'm guessing it's an ESLint bug (which would explain why the current code import types all have only one type).

packages/neuron-ui/.eslintrc.js
yuche: Are this and [that](https://github.com/nervosnetwork/neuron/pull/1802#discussion_r474162544) the same thing?

packages/neuron-ui/src/services/remote/index.ts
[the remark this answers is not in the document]
Keith-CY: I think that's an error from `prettier`(in most cases prettier is so stupid and always outdated) since I can import 2 types in `neuron-wallet`. I also tried to remove `prettier` from `neuron-ui` and the alert was dismissed.

For now, let's just leave it.

packages/neuron-ui/config-overrides.js
Keith-CY: I see, don't change the argument.

packages/neuron-ui/.eslintrc.js
Keith-CY: How about turning it off in specific modules.

yuche:
> > > We introduced electron modules for convenience in the early stage, now we can move related functions into the main process. And some of them can be replaced by browser APIs.
> > 
> > 
> > I'm not sure if I'm misinterpreting what that means. IMHO, I don't see the benefit of moving Electron modules which are belong to renderer process to the main process, but rather adding complexity to IPC communications. As for using some of the browser APIs as replacement is a good idea, but I think this is something we should leave for the future when we need to move from a C/S architecture to a B/S architecture.
> > > I see that you are injecting electron modules into window.electron, then we can use Proxy to trap properties of window.electron and show a warning once window.electron.xxx, which is undefined, is used.
> > 
> > 
> > That's something we can do now.
> 
> According to the [declaration](https://github.com/nervosnetwork/neuron/blob/refactor/import-electron-module-in-render-process/packages/neuron-ui/src/types/global/index.d.ts), the following modules are injected
> 
> * clipboard: can be replaced by `window.navigator.clipboard`
> * remote: can be imported by `require('electron').remote`, actually it's just used to get the `require` module, but the `require` module is injected directly.
> * shell: can be replaced by [open in window](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/api.ts#L142)
> * require
> * nativeImage: can be imported by `require('electron').nativeImage`, actually we can replace it by canvas
> * desktopCapture: can be imported by `require('electron').desktopCapture`, actually we can do it in the main process since qrcode scanning may be called outside the window
> * ipcRenderer: can be imported by `require('electron').ipcRenderer`

Please see the [PR description](https://github.com/nervosnetwork/neuron/pull/1802#issue-463089688): We can't import any electron module using `require()` after `nodeIntegration` is set to `false`.  The current implementation (using Webpack aliasing) is mainly for code consistency and type-safe. The main purpose of the [declaration](https://github.com/nervosnetwork/neuron/blob/refactor/import-electron-module-in-render-process/packages/neuron-ui/src/types/global/index.d.ts) is to pass type checking, the real injected modules are in [preload.ts](https://github.com/nervosnetwork/neuron/blob/2c92e6266b736cdb998774415028435cbcf0aea3/packages/neuron-wallet/src/controllers/app/preload.ts) and [electron-modules.ts](https://github.com/nervosnetwork/neuron/blob/2c92e6266b736cdb998774415028435cbcf0aea3/packages/neuron-ui/src/electron-modules.tsx).






> When I say we can move some electron modules into the main process, I mean that they are expected to be used in the main process, not those declared available in the renderer process.
> 
> For example, the `Menu` is introduced in https://github.com/nervosnetwork/neuron/blob/refactor/import-electron-module-in-render-process/packages/neuron-ui/src/services/remote/index.ts#L86, while the document says that it is called in the main process. When it's used in the renderer process directly, an IPC is connected and a remote object is created, which is bug-prone. We have treated similar problems due to remote objects.
> 
> PS: I noticed this due to `import { MenuItem } from 'electron'`(https://github.com/nervosnetwork/neuron/blob/refactor/import-electron-module-in-render-process/packages/neuron-ui/src/services/remote/index.ts#L8), which misled me to regard them as an imported module, but they are imported as type in fact and should be imported as `import type { MenuItem } from 'electron'`

Please see #1803, Looking at the two PRs together, a module that should be running in the main process (e.g. `Menu` you mentioned) has been refactored into the main process by using IPC, and a module that could be used in the renderer process remains in the renderer process.

Keith-CY (approved):

katat (approved):

PR #1804, opened 2020-08-07
Hot fix to TLS support on indexer

katat:
* Upgrade Lumos indexer to support node connection via TLS

@kellyshang Based on the current branching practice we are using, should I target this PR to another branch such as `hotfix/0.32.1` and leave it for you to handle the rest, such as the version bumping and updating release notes? Please feel free to suggest.

### DIFF
--- a/packages/neuron-wallet/package.json
+++ b/packages/neuron-wallet/package.json
@@ -34,8 +34,8 @@
     ]
   },
   "dependencies": {
-    "@ckb-lumos/base": "0.5.1",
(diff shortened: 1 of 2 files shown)


kellyshang:
> * Upgrade Lumos indexer to support node connection via TLS
> 
> @kellyshang Based on the current branching practice we are using, should I target this PR to another branch such as `hotfix/0.32.1` and leave it for you to handle the rest, such as the version bumping and updating release notes? Please feel free to suggest.

It's OK for me.

PR #1811, opened 2020-08-18
Build and test in same step in Windows CI

yuche:
Windows CI seems to purge files that have been created after each step, resulting in compiled files that cannot be found in the test.

This PR resolves the testing issue with #1800.

failed case:
https://dev.azure.com/nervosnetwork/neuron/_build/results?buildId=13509&view=logs&j=1dd1a9fa-5bd6-5a58-be75-4682264ffaf4&t=0fd07730-01dd-5646-2168-32456ff1e2aa&l=861

success case(in my own repo):
https://dev.azure.com/nervjs/Nervjs/_build/results?buildId=65&view=logs&j=1dd1a9fa-5bd6-5a58-be75-4682264ffaf4&t=0fd07730-01dd-5646-2168-32456ff1e2aa&l=232

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -83,10 +83,10 @@ stages:
         yarn bootstrap
         set CI=false
         yarn build
-      name: Build
+      name: Bootstrap
     - script: |
         set CI=true
-        yarn test
+        yarn test:ci
       name: Test
 
 - stage: release
--- a/package.json
+++ b/package.json
@@ -29,8 +29,10 @@
     "clean": "lerna run --stream clean",
     "prebuild": "yarn clean",
     "build": "lerna run --stream build",
+    "build:main": "lerna run build --scope=neuron-wallet",
     "release": "yarn build && ./scripts/copy-ui-files.sh && ./scripts/release.sh",
     "test": "lerna run --parallel test",
+    "test:ci": "yarn build:main && yarn test",
     "lint": "lerna run --stream lint",
     "postinstall": "lerna run rebuild:nativemodules",
     "db:chain": "node ./node_modules/.bin/typeorm"


azure-pipelines.yml
```diff
@@ -83,10 +83,10 @@ stages:
         yarn bootstrap
         set CI=false
         yarn build
-      name: Build
+      name: Bootstrap
     - script: |
         set CI=true
-        yarn test
+        yarn test:ci
```
Keith-CY, line 89: Thanks to this commit that I've found something interesting.

We're testing `*.ts` then why do we build before the tests?

azure-pipelines.yml
```diff
@@ -83,10 +83,10 @@ stages:
         yarn bootstrap
         set CI=false
         yarn build
```
Keith-CY, line 85: Thanks to this commit that I've found something interesting.

We're testing `*.ts` then why do we build before the tests?

yuche: Now we have some test cases executing .js files via child_process.fork:
https://github.com/nervosnetwork/neuron/blob/7d1b00b58fda76e282cbc92206b1550caa36eb29/packages/neuron-wallet/tests/utils/worker.test.ts#L39

Keith-CY: > Now we have some test cases executing .js files via child_process.fork:
> https://github.com/nervosnetwork/neuron/blob/7d1b00b58fda76e282cbc92206b1550caa36eb29/packages/neuron-wallet/tests/utils/worker.test.ts#L39

I was curious about adding the build step before. 🤔

katat: @yuche 
Could you please explain a bit more about why we need to change the build script to accommodate this test suit?

katat:
Looks like this test suit is for integration tests on `utils/worker`. If so, use the name `utils/worker.intg.test.ts`. We use the file extension `intg` to differentiate from unit tests.

azure-pipelines.yml
katat: Because it is a mock-up fixture, it is only for the test purpose. What if we directly save it as a `js` file for the fork function to refer to?

yuche: > Because it is a mock-up fixture, it is only for the test purpose. What if we directly save it as a `js` file for the fork function to refer to?

We are using a pure javascript file for `child_process.fork`,  But that file needs to import the compiled file:

https://github.com/nervosnetwork/neuron/blob/7d1b00b58fda76e282cbc92206b1550caa36eb29/packages/neuron-wallet/tests/utils/fixtures/worker.js#L1-L6

To summarize, we need to test `child_process.fork`, but `child_process.fork` can only use .js files, and .js files can only import .js source code, so we have to compile the source code before testing.

katat: Right. I mean can we just save that specific worker fixture file as a pure javascript file instead of a typescript, so no need to involve the compilation?

yuche: > > Now we have some test cases executing .js files via child_process.fork:
> 
> I was curious about adding the build step before. 🤔




> Right. I mean can we just save that specific worker fixture file as a pure javascript file instead of a typescript, so no need to involve the compilation?

I'm sorry I didn't get that, but the worker fixture file is now a pure javascript file.

katat: Ok, I assumed it was a ts file. 

How about using this technique. Does it work in our context? https://github.com/TypeStrong/ts-node/issues/619#issuecomment-511366883

yuche: > Ok, I assumed it was a ts file.
> 
> How about using this technique. Does it work in our context? [TypeStrong/ts-node#619 (comment)](https://github.com/TypeStrong/ts-node/issues/619#issuecomment-511366883)

I haven't tested it yet, but this should work. But then we need to introduce an additional dependency: `ts-node`.

Perhaps I wasn't  describe it clear enough, but the current implementation actually runs on macOS and Linux, as well as Windows on the local machine, but the Windows instance of the Azure pipeline cleans up the files already created at each step, so the test fails.

Before this PR, the original procedure was to compile the source files before testing, but testing and compiling were in separate steps, except that this PR puts compiling and testing in one step in Windows CI.

katat: Makes sense to me

katat (approved):

azure-pipelines.yml
Keith-CY: If I got it right, this commit is to fix a test failure introduced by a PR that is not merged yet? Then why not merge this fix into that PR?

Keith-CY: And any documents indicate that `but the Windows instance of the Azure pipeline cleans up the files already created at each step, so the test fails.`? It is kinda weird since Mac and Linux don't do that.

yuche: > And any documents indicate that `but the Windows instance of the Azure pipeline cleans up the files already created at each step, so the test fails.`? It is kinda weird since Mac and Linux don't do that.

No, but this PR does fix the problem, and I've tested it in my own repo. Please check out the two links in the PR description.

Keith-CY: > > And any documents indicate that `but the Windows instance of the Azure pipeline cleans up the files already created at each step, so the test fails.`? It is kinda weird since Mac and Linux don't do that.
> 
> No, but this PR does fix the problem, and I've tested it in my own repo. Please check out the two links in the PR description.

Don't mind that, I was just curious about that.

Keith-CY: But [this](https://github.com/nervosnetwork/neuron/pull/1811#discussion_r472677882) is my question, why not merge this into the PR introduced the test failure.

yuche: > But [this](https://github.com/nervosnetwork/neuron/pull/1811#discussion_r472677882) is my question, why not merge this into the PR introduced the test failure.

I don't have access to the Neuron repository's branch protection rules, but I'm guessing we're supposed to be protecting the develop branch. since #1800 merge to develop branch, so we need to merge this PR into the develop branch for the CI to work.

Keith-CY: I mean merge this commit into https://github.com/nervosnetwork/neuron/tree/refactor/upgrade-to-electron-9

Keith-CY (approved):

azure-pipelines.yml
Keith-CY: It seems that this branch including extra commits to https://github.com/nervosnetwork/neuron/tree/refactor/upgrade-to-electron-9, I set the pr base back to develop

PR #1815, opened 2020-08-26
Fix icon package issue in Electron 9

yuche:
For some how, in Electron 9 packaged mode, BrowserWindow's `icon` can only accept a instance of `NativeImage`.

ref:
https://github.com/electron/electron/issues/24221

### DIFF
--- a/packages/neuron-wallet/src/controllers/app/index.ts
+++ b/packages/neuron-wallet/src/controllers/app/index.ts
@@ -1,6 +1,6 @@
 import path from 'path'
 import { t } from 'i18next'
-import { app as electronApp, remote, BrowserWindow } from 'electron'
+import { app as electronApp, remote, BrowserWindow, nativeImage } from 'electron'
 import windowStateKeeper from 'electron-window-state'
 
 import env from 'env'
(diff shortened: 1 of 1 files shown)


packages/neuron-wallet/src/controllers/app/index.ts
```diff
@@ -95,7 +95,13 @@ export default class AppController {
       minHeight: 600,
       show: false,
       backgroundColor: '#e9ecef',
-      icon: path.join(__dirname, '../../../assets/icons/icon.png'),
+      icon: nativeImage.createFromPath(
+        app.isPackaged
+          ? path.join(__dirname, '../../neuron-ui/icon.png')
```
Keith-CY, line 100: Why not use `assets/icons/icon.png` in packaged mode?

Keith-CY, line 100: Why not use `assets/icons/icon.png` in packaged mode?

yuche: `neuron-ui/icon.png` is the path [previously used](https://github.com/nervosnetwork/neuron/pull/1800/commits/539f7fe4f140a6b3dd41fe0928687ac43f5860eb), since `neuron-ui` is copied to dist folder during packaging, and `assets/icons/icon.png` was not packaged into asar.

Keith-CY: I see, it's kinda smelling but ok for now.

Keith-CY (approved):

katat (approved):

PR #1816, opened 2020-08-26
Refactor address store

katat:
This PR refactored the address storage to be based on a sqlite table storing the public keys for wallets rather than a JSON file.

* Instead of storing the balances for each address in the JSON file, they are calculated on the fly when needed.
* Similar to balance calculation, the transaction count of each address is also calculated on demand.
* Most of the queries based on the lock hashes relating to all the addresses under a wallet are refactored to be based on `wallet id` instead. This fixes the `too many arguments in SQL` when there are too many lock hashes to query with.


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -149,128 +160,196 @@ export default class AddressService {
...
-    const addressEntity = AddressDao.nextUnusedAddress(walletId,  AddressService.getAddressVersion())
-    if (!addressEntity) {
+  private static async maxAddressIndex(walletId: string, addressType: AddressType): Promise<number | undefined> {
+    const result = await getConnection()
+      .getRepository(HdPublicKeyInfo)
+      .createQueryBuilder()
+      .select('addressIndex')
+      .where({walletId, addressType})
+      .orderBy('addressIndex', 'DESC')
+      .getRawOne()
```
Keith-CY, line 198: `count` is enough

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -149,128 +160,196 @@ export default class AddressService {
...
+      .filter(addr => addr.addressType === AddressType.Change)
+
+    return [unusedReceivingAddresses, unusedChangeAddresses]
   }
 
-  public static nextUnusedChangeAddress = (walletId: string): AddressInterface | undefined => {
-    const addressEntity = AddressDao.nextUnusedChangeAddress(walletId,  AddressService.getAddressVersion())
-    if (!addressEntity) {
-      return undefined
+  public static async getAllUnusedAddressesByWalletId (walletId: string) {
```
Keith-CY, line 218: Hard to tell the difference between `getUnusedAddresses` and `getAllUnusedAddressesByWalletId` by their names.

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -149,128 +160,196 @@ export default class AddressService {
...
+    const [ unusedReceivingAddresses ] = await this.getUnusedAddresses(walletId)
+    return unusedReceivingAddresses
   }
 
-  public static allLockHashes(): string[] {
-    const addresses = AddressService.allAddresses().map(address => address.address)
-    return AddressParser.batchToLockHash(addresses)
+
+  public static allAddresses = async (): Promise<AddressInterface[]> => {
+    const publicKeyInfos = await AddressDao.allAddresses()
```
Keith-CY, line 269: The name is `allAddresses` but only those of `AddressDao` are returned?

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -149,128 +160,196 @@ export default class AddressService {
...
+      addressType: publicKeyInfo.addressType,
+      blake160: publicKeyInfo.publicKeyInBlake160,
+      description: publicKeyInfo.description,
+    }))
   }
 
-  public static allLockHashesByWalletId(walletId: string): string[] {
-    const addresses = AddressService.allAddressesByWalletId(walletId).map(addr => addr.address)
-    return AddressParser.batchToLockHash(addresses)
+  public static async allAddressesByWalletId (walletId: string): Promise<AddressInterface[]> {
```
Keith-CY, line 281: Hard to tell the difference between `allAddresses` and `allAddressesByWalletId` by their names.

packages/neuron-wallet/src/utils/export-history.ts
```diff
@@ -43,7 +43,9 @@ const exportHistory = async ({
     `${headers.map(label => t(`export-transactions.column.${label}`))}\n`
   )
 
-  const addresses = AddressService.allAddressesByWalletId(walletID).map(addr => addr.address)
+  const allAddresses = await AddressService.allAddressesByWalletId(walletID)
```
Keith-CY, line 46: `allAddresses` and `addresses` are confusing

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -186,40 +214,46 @@ export default class CellsService {
...
       .orderBy('tx.timestamp', 'ASC')
-      .skip(skip)
-      .take(pageSize)
       .getMany()
 
-    const cells: Cell[] = outputs.map(o => {
+    const matchedOutputs = allMutiSignOutputs.filter(o => {
+      if (o.multiSignBlake160) {
+        return multiSignHashes.has(o.multiSignBlake160)
+      }
```
Keith-CY, line 240: Expect to return `false` at the end

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -186,40 +214,46 @@ export default class CellsService {
...
+    const cells: Cell[] = matchedOutputs.map(o => {
       const cell = o.toModel()
       cell.setCustomizedAssetInfo({
         lock: CustomizedLock.SingleMultiSign,
         type: '',
         data: ''
       })
       return cell
-    })
+    }).slice(skip, pageNo * pageSize);
```
Keith-CY, line 253: `slice` can be prepended to `map`

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -58,14 +55,14 @@ export default class TransactionSender {
     await TransactionPersistor.saveSentTx(tx, txHash)
 
     // update addresses txCount and balance
-    const blake160s = TransactionsService.blake160sOfTx(tx)
-    const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
-    const usedAddresses = blake160s.map(blake160 => AddressGenerator.toShortByBlake160(blake160, prefix))
-    await WalletService.updateUsedAddresses(usedAddresses, blake160s)
+    // const blake160s = TransactionsService.blake160sOfTx(tx)
+    // const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
+    // const usedAddresses = blake160s.map(blake160 => AddressGenerator.toShortByBlake160(blake160, prefix))
```
Keith-CY, line 60: Meaningless comments

packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -96,7 +96,6 @@ export const useOnCurrentWalletChange = ({
   dispatch: StateDispatch
 }) => {
   useEffect(() => {
-    console.info(`switching to ${walletID}`)
     initAppState()(dispatch, history)
   }, [walletID, dispatch, history])
```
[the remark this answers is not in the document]
katat: I thought this logging was added by me, but actually not. I will keep this info log.
Just curious, why removing this line of logging would introduce a warning from that hook dependency?

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -57,10 +54,10 @@ const trackingStatus = async () => {
...
 
   if (failedTxs.length) {
-    const blake160s = await FailedTransaction.updateFailedTxs(failedTxs.map(tx => tx.hash))
-    const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
-    const usedAddresses = blake160s.map(blake160 => AddressGenerator.toShortByBlake160(blake160, prefix))
-    await WalletService.updateUsedAddresses(usedAddresses, blake160s)
+    await FailedTransaction.updateFailedTxs(failedTxs.map(tx => tx.hash))
+    // const prefix = NetworksService.getInstance().isMainnet() ? AddressPrefix.Mainnet : AddressPrefix.Testnet
+    // const usedAddresses = blake160s.map(blake160 => AddressGenerator.toShortByBlake160(blake160, prefix))
+    // await WalletService.updateUsedAddresses(usedAddresses, blake160s)
```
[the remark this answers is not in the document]
katat: They are indeed dead code. I will remove them.

packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -132,6 +131,7 @@ export const useSubscription = ({
           if (!isAllowedToFetchList) {
             break
           }
+          updateAddressListAndBalance(walletID)(dispatch)
```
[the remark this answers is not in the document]
katat: The refreshments of the balances are used to listen to the `address-update`. After refactored the address store, the calculations are done on the fly so it should listen to the updates of transactions instead.

The `address-update` event would be triggered only when there are changes to a wallet's address list.

packages/neuron-wallet/src/database/chain/entities/hd-public-key-info.ts
```diff
@@ -0,0 +1,70 @@
...
+export default class HdPublicKeyInfo {
+
+  @PrimaryGeneratedColumn()
+  id!: number
+
+  @Column({
+    type: "varchar",
+  })
+  @Index()
+  walletId!: string
```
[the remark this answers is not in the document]
katat: This table will contain **many** `public keys` relating to **one** `walletId`. It is `one to many` relationships. Defining the `walletId` as a primary key won't allow this `one to many` relationships in our use case.

packages/neuron-wallet/src/database/chain/entities/hd-public-key-info.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  @Column({
+    type: 'varchar',
+  })
+  path!: string
```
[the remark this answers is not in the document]
katat: Indeed. If we want it inferred from these two columns, perhaps we need another column, or somewhere else outside the table, to denote its parent path.

packages/neuron-wallet/src/database/chain/entities/hd-public-key-info.ts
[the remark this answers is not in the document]
Keith-CY: I see

packages/neuron-wallet/src/database/chain/entities/hd-public-key-info.ts
```diff
@@ -0,0 +1,70 @@
...
+    publicKeyInfo.addressType = model.addressType
+    publicKeyInfo.addressIndex = model.addressIndex
+    publicKeyInfo.publicKeyInBlake160 = model.publicKeyInBlake160
+    publicKeyInfo.description = model.description
+
+    return publicKeyInfo
+  }
+
+  public static fromObject(...args: any[]): HdPublicKeyInfo {
+    const model = HdPublicKeyInfoModel.fromObject.apply(HdPublicKeyInfoModel, args)
```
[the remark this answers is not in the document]
katat: It is just to proxy anything from the arguments to the other method. Feel free to suggest if any other better ways to way to do this.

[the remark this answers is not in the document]
Keith-CY: `HdPublicKeyInfoModel.fromObject(...args)` could be enough.

packages/neuron-wallet/src/services/addresses.ts
katat: I want it to ensure the max index is indeed the max one stored in the table, just in case there are corruptions in the table data.

packages/neuron-wallet/src/services/addresses.ts
katat: The name of `AddressDao` is quite misleading. It encapsulating the data operations for addresses no matter it is related to DAO or not. I will merge this `AddressDao` class into `AddressService`.

packages/neuron-wallet/src/database/chain/entities/hd-public-key-info.ts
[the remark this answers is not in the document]
katat: Then it can't get around the checks by TS.
Screen Shot 2020-08-27 at 12 13 28 PM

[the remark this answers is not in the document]
Keith-CY: Declare the `args` as `Parameter<typeof HdPublicKeyInfoModel.fromObject>`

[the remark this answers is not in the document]
katat: It works. Thanks a lot for this TS technique.

packages/neuron-wallet/src/utils/export-history.ts
katat: Please see the latest commits

Keith-CY (approved):

packages/neuron-wallet/src/database/chain/entities/hd-public-key-info.ts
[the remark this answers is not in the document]
katat: Removed this and derive the path from the `HdPublicKeyInfoModel` instead.

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -149,128 +173,194 @@ export default class AddressService {
...
+
+  public static getAddressesByAllWallets = async (): Promise<AddressInterface[]> => {
+    const publicKeyInfos = await getConnection()
+      .getRepository(HdPublicKeyInfo)
+      .createQueryBuilder()
+      .getMany()
+
+    return publicKeyInfos
+      .map(publicKeyInfo => publicKeyInfo.toModel())
+      .map(model => (AddressMeta.fromHdPublicKeyInfoModel(model)))
```
yuche, line 289: Is there an extra loop?  We only need to call the `map` function once to get the result.

packages/neuron-wallet/src/services/asset-account-service.ts
```diff
@@ -324,8 +336,8 @@ export default class AssetAccountService {
     await connection.manager.save([entity.sudtTokenInfo, entity])
 
     // 4. update address for usedByAnyoneCanPay
-    const addressVersion = NetworksService.getInstance().isMainnet() ? AddressVersion.Mainnet : AddressVersion.Testnet
-    AddressService.updateUsedByAnyoneCanPay(walletID, assetAccount.blake160, addressVersion, true)
+    // const addressVersion = NetworksService.getInstance().isMainnet() ? AddressVersion.Mainnet : AddressVersion.Testnet
+    // AddressService.updateUsedByAnyoneCanPay(walletID, assetAccount.blake160, addressVersion, true)
```
yuche, line 341: These comments can be removed.

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -149,128 +173,194 @@ export default class AddressService {
...
+      .getRepository(HdPublicKeyInfo)
+      .createQueryBuilder()
+      .where({walletId})
+      .getMany()
+
+    return publicKeyInfos.sort((lhs, rhs) => {
+        return lhs.addressType - rhs.addressType || lhs.addressIndex - rhs.addressIndex
+      })
+      .map(publicKeyInfo => publicKeyInfo.toModel())
+      .map(model => (AddressMeta.fromHdPublicKeyInfoModel(model)))
```
yuche, line 303: extra loop.

packages/neuron-wallet/src/database/address/meta.ts
```diff
@@ -62,19 +63,39 @@ export default class AddressMeta implements Address {
...
+      obj.publicKeyInBlake160,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      obj.description,
+      undefined,
+      undefined
```
yuche, line 95: Are the last two parameters unnecessary?

packages/neuron-wallet/src/models/keys/hd-public-key-info.ts
```diff
@@ -0,0 +1,49 @@
...
+export default class HdPublicKeyInfoModel {
+  public walletId: string
+  public address: string
+  public addressType: AddressType
+  public addressIndex: number
+  public publicKeyInBlake160: string
+  public description?: string
+
+  public get path(): string {
+    return `${AccountExtendedPublicKey.ckbAccountPath}/${this.addressType}/${this.addressIndex}`
```
yuche, line 13: This function may be more expressively by implementing with `Address.pathFor()`.

katat: Reall good spot here 👍

katat:
@yuche 
Just pushed an update according to the comments.

yuche (approved):

Keith-CY:
Please rebase or squash before merging

PR #1817, opened 2020-08-30
Support public key management for hardware wallet

katat:
- Support creating single address hardware wallet
- Support the public key management for both software and hardware wallets
- Encapsulates some of the address related functions in `WalletService`

### DIFF
--- a/packages/neuron-wallet/src/exceptions/wallet.ts
+++ b/packages/neuron-wallet/src/exceptions/wallet.ts
@@ -65,6 +65,12 @@ export class InvalidKeystore extends Error {
   }
 }
 
+export class WalletFunctionNotSupported extends Error {
+  constructor(func: string) {
+    super(t('messages.not-supported-function', {func}))
+  }
+}
+
 export default {
   WalletNotFound,
   CurrentWalletNotSet,
--- a/packages/neuron-wallet/src/services/anyone-can-pay.ts
+++ b/packages/neuron-wallet/src/services/anyone-can-pay.ts
@@ -1,4 +1,3 @@
-import AddressService from "./addresses"
 import AssetAccountInfo from "models/asset-account-info"
 import AddressParser from "models/address-parser"
 import { TransactionGenerator } from "./tx"
@@ -11,6 +10,7 @@ import { TargetOutputNotFoundError } from "exceptions"
 import { AcpSendSameAccountError } from "exceptions"
 import Script from "models/chain/script"
 import LiveCellService from "./live-cell-service"
+import WalletService from "./wallets"
 
 export default class AnyoneCanPayService {
   public static async generateAnyoneCanPayTx(
@@ -68,7 +68,8 @@ export default class AnyoneCanPayService {
       outPoint: targetOutputLiveCell.outPoint(),
     })
 
-    const changeBlake160: string = (await AddressService.getNextUnusedChangeAddressByWalletId(walletID))!.blake160
+    const wallet = WalletService.getInstance().get(walletID)
+    const changeBlake160: string = (await wallet.getNextChangeAddressByWalletId())!.blake160
 
     const tx = isCKB ? await TransactionGenerator.generateAnyoneCanPayToCKBTx(
       walletID,
--- a/packages/neuron-wallet/src/services/wallets.ts
+++ b/packages/neuron-wallet/src/services/wallets.ts
@@ -1,5 +1,5 @@
 import { v4 as uuid } from 'uuid'
-import { WalletNotFound, IsRequired, UsedName } from 'exceptions'
+import { WalletNotFound, IsRequired, UsedName, WalletFunctionNotSupported } from 'exceptions'
 import Store from 'models/store'
 import Keystore from 'models/keys/keystore'
 import WalletDeletedSubject from 'models/subjects/wallet-deleted-subject'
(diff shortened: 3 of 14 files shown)


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -1,42 +1,45 @@
...
-export interface Wallet {
-  id: string
+export interface DeviceProperties {
   name: string
-
-  loadKeystore: () => Keystore
-  accountExtendedPublicKey: () => AccountExtendedPublicKey
+  publicKey: string
+  addressIndex: number
+  addressType: AddressType
```
yuche, line 24: IMHO, `addressIndex` and `addressType` can also be added as properties of `WalletProperties`, since in software wallets we need to calculate the path as well：
 https://github.com/nervosnetwork/neuron/blob/9857dc9d5e0d3a9bf605d1825a77a8d8f59edb24/packages/neuron-wallet/src/services/transaction-sender.ts#L86-L91

It might be better for `DeviceProperties` to be named `DeviceInfo`, there are some fields that need to be added to implement some functionality, here is the` DeviceInfo` data structure for my branch:
```typescript
export enum Manufacturer {
  Ledger = 'Ledger'
}

export interface DeviceInfo {
  descriptor: string
  publicKey: string
  vendorId: string
  manufacturer: Manufacturer
  product: string
  connectType: 'bluetooth' | 'usb'
  // The following information may not be available to us in some wallets
  isConnected?: boolean
  appVersion?: string
  firewareVersion?: string
}
```
But we can just ignore that for now and deal with it when I merge the code.

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -45,17 +48,49 @@ export class FileKeystoreWallet implements Wallet {
       throw new IsRequired('Name')
     }
 
+    if (!extendedKey && !device) {
```
yuche, line 51: Is this tester supposed to be `!extendedKey || !device`?

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -45,17 +48,49 @@ export class FileKeystoreWallet implements Wallet {
...
+
+  public fromJSON = () => {
+    throw new Error('not implemented')
   }
 
-  static fromJSON = (json: WalletProperties) => {
-    return new FileKeystoreWallet(json)
+  public isHardware = (): boolean => {
+    throw new Error('not implemented')
+  }
```
yuche, line 74: TypeScript's `abstract class` can omit the implementation of functions like interface:
```typescript
public abstract isHardware (): boolean
```
Should be enough.

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -88,6 +149,77 @@ export class FileKeystoreWallet implements Wallet {
...
+      addressIndex
+    )
+
+    if (address) {
+      return [address]
+    }
+  }
+
+  public getNextAddressByWalletId = async (): Promise<AddressInterface | undefined> => {
+    return AddressService.getFirstAddressByWalletId(this.id)
```
yuche, line 208: I find it a bit odd that the function is named `getNextAddressByWalletId`, since it doesn't take a `walletId` argument. Maybe `getNextAddress` is enough?

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -130,6 +262,15 @@ export default class WalletService {
     })
   }
 
+  private fromJSON(json: WalletProperties) {
+    if (json.device) {
+      return HardwareWallet.fromJSON(json)
+    }
+    else {
```
yuche, line 269: This `else` is not necessary.

packages/neuron-wallet/src/exceptions/wallet.ts
```diff
@@ -65,6 +65,12 @@ export class InvalidKeystore extends Error {
   }
 }
 
+export class WalletFunctionNotSupported extends Error {
```
Keith-CY, line 68: Error code is preferable.

packages/neuron-wallet/src/services/anyone-can-pay.ts
```diff
@@ -11,6 +10,7 @@ import { TargetOutputNotFoundError } from "exceptions"
 import { AcpSendSameAccountError } from "exceptions"
 import Script from "models/chain/script"
 import LiveCellService from "./live-cell-service"
+import WalletService from "./wallets"
```
Keith-CY, line 13: The module name (`Wallet`) is different from the file name(`wallets`)

packages/neuron-wallet/src/services/wallets.ts
katat: It requires either `extendedKey` or `device`. If both are missing, it throws an error.

packages/neuron-wallet/src/services/wallets.ts
katat: > IMHO, addressIndex and addressType can also be added as properties of WalletProperties, since in software wallets we need to calculate the path as well

These two properties are for single address hardware wallet for now. For software wallet, the addresses are derived from the extended public key `extendedKey` field and the two properties are recorded in the hd_public_key_info table for each address generation.

Because the initial version of hardware wallet support is single address based, assuming they are not derived from an extended public key, I store the `addressIndex` and `addressType` in the `DeviceInfo` as a part of wallet JSON store.

packages/neuron-wallet/src/services/wallets.ts
katat: Ah right. The result of copy and paste. `getNextAddress` is a better naming.

PR #1819, opened 2020-08-31
Enable sudt mainnet

Keith-CY:


packages/neuron-ui/src/utils/validators/sudtAddress.ts
```diff
@@ -18,10 +18,17 @@ export const validateSUDTAddress = ({
...
     const parsed = ckbCore.utils.parseAddress(address, 'hex')
-    // only anyone can pay address is validated
-    if (!parsed.startsWith(LONG_TYPE_PREFIX)) {
+
+    /**
+     * Only anyone can pay address(type id ver/data ver) is under validation
+     * - data version for mainnet
+     * - type version for testnet
+     */
+    const addrType = isMainnet ? LONG_DATA_PREFIX : LONG_TYPE_PREFIX
```
katat, line 27: The network type doesn't necessarily determine the address type for ACP. I think we can remove this check.

Keith-CY: Can neuron-wallet pick the right code hash by the address type?

katat: It should be able to.

Keith-CY: Then [this method](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/api.ts#L390-L392) should be refactored to provides those two kinds of code hash that UI can impose validation on the addresses.

Keith-CY: And [this part](https://github.com/nervosnetwork/neuron/blob/8954b7b59c4d7ed67a7d669727f246a0e1f3604a/packages/neuron-wallet/src/services/tx/transaction-generator.ts#L696) of generating ACP tx should be refactored two

katat: Ok, let's keep this check for now. We will come back to review these reactors are necessary if this check is removed later.

katat (approved):

yuche (approved):

PR #1823, opened 2020-09-02
Fix `InvalidCodeHash` error when sending tx

katat:
* Fix incorrect inputs gathering, which could resulting in `InvalidCodeHash` error if a gathered input cell comes with an unexpected hash type in the lock script.
* Count transactions based on public key hash

### DIFF
--- a/packages/neuron-wallet/src/services/cells.ts
+++ b/packages/neuron-wallet/src/services/cells.ts
@@ -16,7 +16,7 @@ import BufferUtils from 'utils/buffer'
 import LiveCell from 'models/chain/live-cell'
 import Output from 'models/chain/output'
 import SystemScriptInfo from 'models/system-script-info'
-import Script from 'models/chain/script'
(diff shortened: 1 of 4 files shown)


packages/neuron-wallet/src/services/cells.ts
```diff
@@ -295,7 +295,10 @@ export default class CellsService {
       input: Input,
       witness: WitnessArgs,
     },
-    lockCodeHashes: string[] = [SystemScriptInfo.SECP_CODE_HASH]
+    lockClass: {
```
Keith-CY, line 298: What's the meaning of `lockClass`?

katat: It means the combination of code hash and hash type under a lock. This is meant to filter the cells with these two data points. I tried to avoid calling the `lockType`, the wording of which would be too confusing with the other concepts on CKB.

yuche (approved):

Keith-CY (approved):

PR #1838, opened 2020-09-07
Deprecating leveldb

katat:
- Refactor to use a sqlite table for storing the tx descriptions
- Migrate data from leveldb to the sqlite table
- leveldb dependency will be deprecated in the next minor version (0.34)

### DIFF
--- a/packages/neuron-wallet/src/database/chain/ormconfig.ts
+++ b/packages/neuron-wallet/src/database/chain/ormconfig.ts
@@ -13,6 +13,7 @@ import SyncInfo from './entities/sync-info'
 import AssetAccount from './entities/asset-account'
 import SudtTokenInfo from './entities/sudt-token-info'
 import IndexerTxHashCache from './entities/indexer-tx-hash-cache'
+import TxDescription from './entities/tx-description'
 
 import { InitMigration1566959757554 } from './migrations/1566959757554-InitMigration'
 import { AddTypeAndHasData1567144517514 } from './migrations/1567144517514-AddTypeAndHasData'
@@ -35,6 +36,7 @@ import { RemoveAssetAccountWalletID1589273902050 } from './migrations/1589273902
 import { RemoveLiveCell1592781363749 } from './migrations/1592781363749-RemoveLiveCell'
 import { AddIndexerTxHashCache1592727615004 } from './migrations/1592727615004-AddIndexerTxHashCache'
 import { HDPublicKeyInfo1598087517643 } from './migrations/1598087517643-HDPublicKeyInfo'
+import { TxDescription1599441769473 } from './migrations/1599441769473-TxDescription'
 
 export const CONNECTION_NOT_FOUND_NAME = 'ConnectionNotFoundError'
 
@@ -57,6 +59,7 @@ const connectOptions = async (genesisBlockHash: string): Promise<SqliteConnectio
     entities: [
       HdPublicKeyInfo,
       Transaction,
+      TxDescription,
       Input,
       Output,
       SyncInfo,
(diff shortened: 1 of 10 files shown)


packages/neuron-wallet/src/database/chain/ormconfig.ts
```diff
@@ -57,6 +59,7 @@ const connectOptions = async (genesisBlockHash: string): Promise<SqliteConnectio
     entities: [
       HdPublicKeyInfo,
       Transaction,
+      TxDescription,
```
Keith-CY, line 62: I'd like to know why not use glob patterns here?

packages/neuron-wallet/src/database/leveldb/index.ts
```diff
@@ -13,7 +13,9 @@ const leveldb = (dbname: string): LevelUp => {
 
   const dbpath = path.join(dir, dbname)
   return levelup(leveldown(dbpath), (err: Error | undefined) => {
-    logger.error(`Database:\tfail to open leveldb ${dbname}:`, err?.toString())
+    if (err) {
+      logger.error(`Database:\tfail to open leveldb ${dbname}:`, err?.toString())
```
Keith-CY, line 17: Optional chaining on `err` is useless.

packages/neuron-wallet/src/services/tx/transaction-description.ts
```diff
@@ -0,0 +1,36 @@
...
+    await getConnection().manager.save(entity)
+    return
+  }
+
+  const txDesc = new TxDescription()
+  txDesc.walletId = walletID
+  txDesc.txHash = txHash
+  txDesc.description = description
+
+  await getConnection().manager.save(txDesc)
```
Keith-CY, line 35: Set `txHash` as primary and use `save` can upsert it.

Keith-CY (approved):

packages/neuron-wallet/src/database/chain/ormconfig.ts
katat: Do you mean the naming pattern?

packages/neuron-wallet/src/database/leveldb/index.ts
katat: Right, this will be removed soon.

packages/neuron-wallet/src/services/tx/transaction-description.ts
katat: Sounds like a nice feature to try out. But using txhash alone is insufficient to uniquely locate an record in this case. It has to use the composite key txhash and walletid to properly locate an record.

packages/neuron-wallet/src/database/chain/ormconfig.ts
Keith-CY: > Do you mean the naming pattern?

Yep

katat: Right. So perhaps TransactionDescription would be more concise.

Keith-CY: Actually, glob patterns or naming patterns I mentioned here are things like https://github.com/typeorm/typeorm/blob/master/README.md#L304

Keith-CY:
```JSON
{
  "entities": [
    "src/entity/**/*.ts"
  ],
}
```

katat: Looks convenient to register entities in this way. We can do this in another PR dedicated for this.

Keith-CY (approved):

yuche (approved):

PR #1850, opened 2020-09-14
fix: reset description on entering customized asset list and ignore d…

Keith-CY:
…escription on unlocking assets

Since only the generated tx was reset on mounting the customized asset list, and description was passed on sending unlock transaction, an unexpected description was appended to unlock transaction.
This commit remove the description field on sending unlock transaction and reset the entire send state on mounting the customized asset list to fix the bug.

### DIFF
--- a/packages/neuron-ui/src/components/PasswordRequest/index.tsx
+++ b/packages/neuron-ui/src/components/PasswordRequest/index.tsx
@@ -97,7 +97,7 @@ const PasswordRequest = () => {
             if (isSending) {
               break
             }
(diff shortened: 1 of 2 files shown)


katat (approved):

katat:
I would suggest merging this into the `rc/0.33` branch.

Keith-CY:
> I would suggest merging this into the `rc/0.33` branch.

PR base is changed to rc/v0.33.0

yuche (approved):

PR #1853, opened 2020-09-16
Additional flags for bundled node

katat:
Added the two additional flags for the bundled node:
- isBundledNode
- startedBundledNode

These flags would help the front end refine the display of connection statuses, particularly for the bundled node.

### DIFF
--- a/packages/neuron-ui/src/components/SUDTSend/index.tsx
+++ b/packages/neuron-ui/src/components/SUDTSend/index.tsx
@@ -370,7 +370,7 @@ const SUDTSend = () => {
               />
             </div>
             <div className={styles.remoteError}>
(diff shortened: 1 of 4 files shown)


Keith-CY:
`startedBundledNode` is not reset on network switching.

katat:
> startedBundledNode is not reset on network switching.

Updated, please have a check.

Keith-CY:
https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/api.ts#L152
`connecting` should be returned in this case.

Keith-CY (approved):

yuche (approved):

PR #1857, opened 2020-09-20
Fix lumos query for multisign lock

katat:
- Fix incorrect `args` for the multi-sign lock 
- Use `argsLen` lumos query filter to locate the cells by multi-sign lock and cache their corresponding transaction hashes for indexing.

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
@@ -3,7 +3,7 @@ import { queue } from 'async'
 import AddressMeta from "database/address/meta"
 import IndexerTxHashCache from 'database/chain/entities/indexer-tx-hash-cache'
 import RpcService from 'services/rpc-service'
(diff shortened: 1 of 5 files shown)


packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
```diff
@@ -158,7 +181,7 @@ export default class IndexerCacheService {
 
     for (const txWithStatus of txsWithStatus) {
       const {transaction, txStatus} = txWithStatus
-      const mappings = mappingsByTxHash.get(transaction.hash)
+      const mappings = mappingsByTxHash.get(transaction.hash!)!
```
Keith-CY, line 184: IMO too many assertions is prone to bug.

katat: There are two assertions here: 
1. make sure the `transaction.hash` is not empty.
2. make sure there is always a value for the hash in a `Map` in this case.

If the assertion failed, then it should throw errors to prevent the potential hidden bugs. Let me know if you got any other ways around this.

yuche (approved):

PR #1861, opened 2020-09-23
Optimize query performance for tx hash upsert

katat:
Concurrent SQLite queries are slower than the ones running in sequence. This PR make the concurrent queries run in order.


Keith-CY:
Better to have a benchmark since the diff shows mere difference.

Keith-CY (approved):

katat:
> Better to have a benchmark since the diff shows mere difference.

The major difference is the speed of individual SQL query can be done. When making queries in parallel, take number 400 for example, the average execution time for each individual is around 100ms. Although the total execution can be done within 500ms, but it leads typeorm to complain a lot of slow query due to parallel executions. 

On the other hand, it takes around 1ms for each sql execution in sequence and it reduces the chance for the typeorm to complain slow execution.

yuche (approved):

PR #1865, opened 2020-09-24
feat: show network type in the navbar

Keith-CY:
image


packages/neuron-ui/src/stories/NetworkStatus.stories.tsx
```diff
@@ -65,11 +88,15 @@ Object.entries(states).forEach(([title, props]) => {
 
 stories.add('With knobs', () => {
   const props = {
-    networkName: text('Network name', 'network name'),
+    network: {
+      name: text('Network name', 'network name'),
+      remote: text('Remote', 'http://localhost:3000'),
+      type: select('Type', [0, 1], 0) as any,
```
katat, line 94: Does this mean it only support two network types: `mainnet` and `testnet`. Will it be able to recognize `devnet`?

Keith-CY: > Does this mean it only support two network types: `mainnet` and `testnet`. Will it be able to recognize `devnet`?

The [type](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/models/network.ts#L1-L4) field is used to denote if it's a preset. [A preset cannot be deleted](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/NetworkSetting/index.tsx#L92)

The [chain](https://github.com/nervosnetwork/neuron/pull/1865/files/322b14cf246d2f22b483ae4c857d8c911b68bbf5#diff-76efc76b6579d3a42b09639324787eb7R96) field is a mark of `ckb`, `ckb_test`, `ckb_dev` and custom string, used in the `NetworkTypeLabel` as follows
https://github.com/nervosnetwork/neuron/pull/1865/files/322b14cf246d2f22b483ae4c857d8c911b68bbf5#diff-444ca03252393f6960dd5798d19ff48dR50

katat (approved):

yuche (approved):

PR #1868, opened 2020-09-25
Estimation on synchronization

katat:
- Deprecated the `synced-block-number-updated` by `sync-estimate-updated`. 
- Updated the front end code to account for this change, so the sync status still works in this PR.
- `sync-estimate-updated` event has the following schema.
- For now, it uses a custom call to fetch sync state from node instead of using the built-in API of SDK +0.35. We will need to go through a refactor to get the code base adapted to typescript 4, which is required by SDK version greater than 0.35

```javascript
{
  nodeUrl: 'http://localhost:8224', // current node url
  timestamp: 1601044101705, // the time when delivered this estimation event
  indexerTipNumber: 369397, // current block tip on indexer
  cacheTipNumber: 367434, // current cached block tip 
  bestKnownBlockNumber: 493640, // best known block number from the network
  indexRate: 0.5230482154434903, // indexing rate [blocks/ms]
  cacheRate: undefined, // caching rate -- just a place holder for now
  estimate: 237536.4188837828, // estimated time (ms) for completing the indexing. If the value is undefined, it means estimation is in progress.
  synced: false // indicates if it is fully cached
}
```

### DIFF
--- a/packages/neuron-ui/src/containers/Main/hooks.ts
+++ b/packages/neuron-ui/src/containers/Main/hooks.ts
@@ -186,10 +186,10 @@ export const useSubscription = ({
       }
     })
 
-    const syncedBlockNumberSubscription = SyncedBlockNumberSubject.subscribe(syncedBlockNumber => {
+    const syncedBlockNumberSubscription = SyncedBlockNumberSubject.subscribe(({ cacheTipNumber }) => {
       dispatch({
         type: NeuronWalletActions.UpdateSyncedBlockNumber,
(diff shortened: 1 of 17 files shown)


Keith-CY:
Does `  estimate: 237536.4188837828, // estimated time (ms) for completing the indexing. If the value is undefined, it means estimation is in progress.
` mean `237536.4188837828 ms`? The decimal may be unnecessary.

yuche (approved):

Keith-CY (approved):

PR #1903, opened 2020-11-12
fix: hardware sign message should use default address to sign

yuche:


packages/neuron-wallet/src/services/hardware/ledger.ts
```diff
@@ -75,7 +75,7 @@ export default class Ledger extends Hardware {
 
   async signMessage (path: string, messageHex: string) {
     const message = HexUtils.removePrefix(messageHex)
-    const signed = await this.ledgerCKB!.signMessage(path, message, false)
+    const signed = await this.ledgerCKB!.signMessage(path === Address.pathForReceiving(0) ? this.defaultAddress : path, message, false)
```
katat, line 78: I guess `this.defaultAddress` should be named as `defaultPath`?

yuche: Yes, the naming is indeed bad.

katat (approved):

Keith-CY (approved):

PR #1904, opened 2020-11-13
Disable sign and export button when the device is not connected

yuche:


packages/neuron-ui/src/components/HardwareSign/index.tsx
```diff
@@ -315,12 +315,16 @@ const HardwareSign = ({
...
+  const dropdownList = useMemo(() => {
+    return [
+      {
+        text: t('offline-sign.sign-and-export'),
+        onClick: signAndExportFromGenerateTx,
+        disabled: status === disconnectStatus,
+        disabledMsg: disconnectStatus,
+      },
+    ]
+  }, [signAndExportFromGenerateTx, status])
```
Keith-CY, line 327: `disconnectStatus` is listed in deps at at https://github.com/nervosnetwork/neuron/blob/refactor/disabled-sign-and-export-button/packages/neuron-ui/src/components/HardwareSign/index.tsx#L236 but not listed in deps here.

yuche: I thought `disconnectStatus` wouldn't change, but after your reminder, it will change if the language setting is changed.

Keith-CY (approved):

katat (approved):

PR #1911, opened 2020-11-17
fix: ensure that the ckb app is open for every device operation

yuche:
Also, if the hardware wallet's original descriptor fails to connect, it will update the wallet's descriptor after reconnection is successful.

### DIFF
--- a/packages/neuron-ui/src/components/HardwareSign/index.tsx
+++ b/packages/neuron-ui/src/components/HardwareSign/index.tsx
@@ -20,9 +20,13 @@ import {
   OfflineSignType,
   OfflineSignJSON,
   signAndExportTransaction,
+  getDeviceCkbAppVersion,
+  DeviceInfo,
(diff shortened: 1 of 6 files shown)


packages/neuron-ui/src/components/HardwareSign/index.tsx
```diff
@@ -61,7 +65,7 @@ const HardwareSign = ({
     onDismiss()
   }, [dispatch, signType, onDismiss])
   const connectStatus = t('hardware-sign.status.connect')
-  const disconnectStatus = t('hardware-sign.status.disconnect')
+  const [unavailableStatus, setUnavailableStatus] = useState<string>(t('hardware-sign.status.disconnect'))
```
Keith-CY, line 68: It's better not to use a translated string as a flag.

katat (approved):

Keith-CY (approved):

PR #1914, opened 2020-11-18
refactor: show ckb default data structure in offline sign dialog

yuche:


Keith-CY:
Why not transform the JSON in neuron-ui directly, or transform it before sending it to neuron-ui?

yuche:
> Why not transform the JSON in neuron-ui directly, or transform it before sending it to neuron-ui?

The way to re-implement the transform of `Transaction` in neuron-ui duplicates neuron-wallet.
Transform it before sending requires modifying neuron-wallet's  almost all offline sign controllers and changing a lot of test cases, for some reason I don't want to spend too much time on this refactoring for now.

katat:
Is using the `ckb.rpc.paramsFormatter.toRawTransaction` in the UI sufficient to resolve the problem that this PR is trying to resolve? In this way, we can avoid implementing an additional backend API for this purpose?

yuche:
> Is using the `ckb.rpc.paramsFormatter.toRawTransaction` in the UI sufficient to resolve the problem that this PR is trying to resolve? In this way, we can avoid implementing an additional backend API for this purpose?

What we really need is the Transaction#toSDKRawTransaction method, ckb.rpc.paramsFormatter.toRawTransaction is not enough. To be honest, I don't think this refactoring is really necessary either, because it doesn't affect the functionality without it.
Further more, Neuron is a general user oriented application, what we should really show to users is the income/outcome/time/transaction type(etc…) of the transaction, instead of a JSON file.
I suggest that we leave the PR unmerged until we decide on a new presentation way in the future.

@Keith-CY @katat

PR #1915, opened 2020-11-19
refactor: sync estimation

katat:
- Provide the `SyncStatus` through the `SyncStateSubject`
 - Refactor both the front end and back end for the `SyncStatus`

### DIFF
--- a/packages/neuron-ui/src/states/init/chain.ts
+++ b/packages/neuron-ui/src/states/init/chain.ts
@@ -24,7 +24,7 @@ export const transactionState: Readonly<State.DetailedTransaction> = {
 export const chainState: Readonly<State.Chain> = {
   networkID: currentNetworkID.load(),
   connectionStatus: ConnectionStatus.Connecting,
-  syncEstimate: {
+  syncState: {
     cacheTipBlockNumber: 0,
     bestKnownBlockNumber: 0,
     bestKnownBlockTimestamp: 0,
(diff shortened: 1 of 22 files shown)


packages/neuron-ui/src/states/init/chain.ts
```diff
@@ -24,11 +24,12 @@ export const transactionState: Readonly<State.DetailedTransaction> = {
 export const chainState: Readonly<State.Chain> = {
   networkID: currentNetworkID.load(),
   connectionStatus: ConnectionStatus.Connecting,
-  syncStatus: {
+  syncEstimate: {
```
Keith-CY, line 27: `estimate` is just 1 of 5 fields in it, so why rename it `syncEstimate`?

katat: Or we should call it the `SyncState`, which comes from the `SyncStateSubject` at the backend and includes the fields: https://github.com/nervosnetwork/neuron/blob/9343de0ddaf0c60a60c852b702286c91e3ffc7c8/packages%2Fneuron-wallet%2Fsrc%2Fcontrollers%2Fsync-api.ts#L20

The main reason I rename it from `SyncStatus` is that the name collides with the other variable, https://github.com/nervosnetwork/neuron/blob/9164b23210f395a2825e3bfa57c445e62fcc7adb/packages%2Fneuron-ui%2Fsrc%2Futils%2Fenums.ts#L109, which is now part of the `SyncState`

Keith-CY: It's nice to be `sync state` in neuron ui and neuron wallet

katat: Sure. Updated

Keith-CY (approved):

yuche (approved):

PR #1928, opened 2020-11-25
Feat/support pw transfer

katat:
- support transfer to pw lock
 - remove the front end check for code hash since the backend also does the check.


packages/neuron-ui/src/utils/validators/sudtAddress.ts
```diff
@@ -29,10 +28,6 @@ export const validateSUDTAddress = ({
     }
 
     const CODE_HASH_LENGTH = 64
-    const codeHashOfAddr = parsed.substr(4, CODE_HASH_LENGTH)
-    if (codeHash && codeHashOfAddr !== codeHash.slice(2)) {
-      throw new FieldInvalidException(FIELD_NAME)
```
Keith-CY, line 34: By removing the validation of code hash, the error message for invalid code hash will be shown at the top but not beneath the input field. Is that expected?

katat: Instead of `Invalid code hash`, it should throw `Invalid anyone-can-pay lock script address` at the top.

I think it is fine since the error only throws when the code hash is not supported, which is rare in practice. It might not be ideal, but it'd better avoid duplicating this code hash check at both front end and back end.

Keith-CY: Got it.

I prefer checking it at front end for user experience and back end for security, but it does attenuate duplication of work to validate it at the back end only.

Keith-CY (approved):

yuche (approved):

PR #1938, opened 2020-12-06
test: increase test coverage in service layer

katat:


packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -74,10 +75,7 @@ const trackingStatus = async () => {
 }
 
 export const register = () => {
-  // every 5 seconds
-  setInterval(async () => {
-    // Disable debug output as it's too annoying.
-    // logger.debug("periodically status tracking ...")
+  interval(5000).pipe().subscribe(async () => {
```
Keith-CY, line 78: What does the empty `pipe` do?

katat: It is actually unnecessary. The empty pipe is now removed.

Keith-CY (approved):

yuche (approved):

PR #1968, opened 2020-12-20
update ui for cheque cells

Keith-CY:
This PR:
1. rebase on development
2. enable sending cheque assets
3. enable claiming cheque assets with/without creating a new account
4. enable withdrawing cheque assets


Keith-CY:
Tests should be fixed in https://github.com/nervosnetwork/neuron/pull/1953

katat:
Thanks @Keith-CY 
I have merged the code here in another PR.

PR #1973, opened 2020-12-30
hotfix 0.34.2

katat:
1. Avoid creating new http agents for every ckb sdk creation. 
2. Fix version updater error when clicks the install button too soon 
3. Update CKB logo

Note for _1_. According to https://nodejs.org/api/http.html#http_new_agent_options, creating new http agents with `keepAlive` option will keep the TCP connection open even it is not being used. This type of agent should be managed to avoid keeping up too many TCP connection, which might eventually lead to `too many open files` error.


Keith-CY:
It's better to use a singleton instead of a new instance

katat:
Right. Updated to use singleton for http(s) agents.

Keith-CY:
It seems that 
```javascript
let httpsAgent: https.Agent
let httpAgent: http.Agent

const getHttpsAgent = () => {
  if (!httpsAgent) {
    httpsAgent = new https.Agent({ keepAlive: true })
  }
  return httpsAgent
}

const getHttpAgent = () => {
  if (!httpAgent) {
    httpAgent = new http.Agent({ keepAlive: true })
  }
  return httpAgent
}
```

has no different from

```javascript
const httpsAgent = new https.Agent({ keepAlive: true })
const httpAgent = new http.Agent({ keepAlive: true })
```

The singleton I've talked about is the `ckb` returned by `generateCKB` actually, codes like:

```javascript
const httpsAgent = new https.Agent({ keepAlive: true })
const httpAgent = new http.Agent({ keepAlive: true })
let ckb: CKB

export const generateCKB = (url: string): CKB => {
  ckb ??= new CKB()
  if (url.startsWith('https')) {
    ckb.rpc.setNode({ url, httpsAgent })
  } else {
    ckb.rpc.setNode({ url, httpAgent })
  }
  return ckb
}
```

After a second thought, I think it's kinda reckless to return a singleton `ckb` since I'm not sure if all `ckb` have the same `url`. So I prefer the previous one if they have no difference.

katat:
There is a difference between the current and previous commits. That is now the agent is only initialized when it is actually needed.

Keith-CY:
> There is a difference between the current and previous commits. That is now the agent is only initialized when it is actually needed.

Gotcha

Keith-CY (approved):

kellyshang (approved):

PR #1980, opened 2021-01-08
chore: fix typo for cheque address hint

kellyshang:


katat:
@kellyshang 
The build failure is probably due to an outdated branch that this PR branch is based on, so it still has the Bluetooth dependency issue on windows build. I have directly cherry picked your commit to the target branch of this PR.

PR #1989, opened 2021-01-20
feat: integration sudt info

solargatsby:
auto fill sudt info such as decimal, name and symbol when create sudt asset


packages/neuron-ui/src/types/Controller/index.d.ts
```diff
@@ -311,4 +311,16 @@ declare namespace Controller {
...
+  namespace GetSUDTTokenInfo {
+    interface Params {
+      tokenId: string
+    }
+    interface Response {
+      tokenId: string
+      symbol: string
+      tokenName: string
+      decimal: string
+    }
```
katat, line 324: Would it better to simplify the following properties?
`tokenId` => `id`
`tokenName` => `name`

packages/neuron-wallet/.env
```diff
@@ -5,6 +5,12 @@ MAINNET_SUDT_DEP_TYPE=code
 MAINNET_SUDT_SCRIPT_CODEHASH=0x5e7a36a77e68eecc013dfa2fe6a23f3b6c344b04005808694ae6dd45eea4cfd5
 MAINNET_SUDT_SCRIPT_HASHTYPE=type
 
+MAINNET_SUDT_INFO_DEP_TXHASH=0x81eeaaedc2909faf471cc17f8aeb66dd5e78d50ad1b7eb56e41ab821ee356330
```
katat, line 8: It would be helpful to add a TODO comment here, so we know these configure will need to be updated once `cell info script` has been deployed on mainnet.

packages/neuron-ui/src/components/SUDTCreateDialog/index.tsx
```diff
@@ -153,10 +153,24 @@ const SUDTCreateDialog = ({
...
+          const { tokenID, ...rest } = tokenInfo
+          dispatch({ type: 'import', payload: { ...rest, tokenId: tokenID } })
+        } else {
+          getSUDTTokenInfo({ tokenId: payload }).then(res => {
+            if (isSuccessResponse(res) && res.result) {
+              dispatch({ type: 'import', payload: { ...res.result } })
+            }
+          })
+          dispatch({ type, payload })
+        }
```
katat, line 173: This looks quite complicated. Can we find a way to simplify these condition checks a bit?

packages/neuron-wallet/src/controllers/sudt.ts
```diff
@@ -0,0 +1,36 @@
...
+import {ResponseCode} from "utils/const";
+
+export interface TokenInfo {
+  tokenId: string
+  symbol: string
+  tokenName: string
+  decimal: string
+}
+
+export default class SUDTController {
```
katat, line 14: Please add unit tests for this controller as well.

packages/neuron-wallet/src/utils/parse_sudt_token_info.ts
```diff
@@ -0,0 +1,18 @@
+const parseSUDTTokenInfo = (hexData : string ) => {
+    let data = hexData
```
Keith-CY, line 2: I'm not quite sure about the encoding of `hexData`, but according to the test cases, I think the decoding can be simplified as follows:

```js
const decode = (hexString) => {
  const decimal = +hexString.substr(0, 4)
  const [_, name, symbol] = Buffer.from(hexString.slice(4), 'hex').toString('utf-8').split(`\n`)
  return { decimal, name, symbol }
}
```

packages/neuron-ui/src/types/Controller/index.d.ts
solargatsby: tokenId and tokenName are also used in other code, so keep consistency maybe better

packages/neuron-wallet/.env
solargatsby: fix

packages/neuron-wallet/src/utils/parse_sudt_token_info.ts
solargatsby: good idea

packages/neuron-wallet/src/controllers/sudt.ts
```diff
@@ -0,0 +1,36 @@
...
+  private liveCellService = new LiveCellService()
+
+  public async getSUDTTokenInfo(params: { tokenId: string }): Promise<Controller.Response<TokenInfo>> {
+    let codeHash = process.env.TESTNET_SUDT_INFO_SCRIPT_CODEHASH!
+    if (!env.isTestMode) {
+      codeHash = process.env.MAINNET_SUDT_INFO_SCRIPT_CODEHASH!
+    }
```
katat, line 21: This would be better to be encapsulated in the existing `AssetAccountInfo` class, so it won't need to do this kind of environment check in the controller level.

solargatsby: fix

packages/neuron-wallet/src/controllers/sudt.ts
```diff
@@ -0,0 +1,36 @@
...
+        status: ResponseCode.Success,
+        result: {...sudtInfo, tokenID: params.tokenID},
+      }
+    }
+
+    const typeScript = Script.fromObject({codeHash: new AssetAccountInfo().sudtInfoCodeHash,
+      args: params.tokenID,
+      hashType: ScriptHashType.Type})
+    const liveCell = await LiveCellService.getInstance().getOneByLockScriptAndTypeScript(null, typeScript)
+    if (liveCell == undefined) {
```
katat, line 24: use `===`

katat:
@solargatsby Please add the unit tests for the newly added `SUDTController`.

packages/neuron-wallet/src/controllers/sudt.ts
solargatsby: fix

packages/neuron-wallet/src/controllers/sudt.ts
solargatsby: fix

katat:
> @solargatsby Please add the unit tests for the newly added SUDTController.

@solargatsby

Keith-CY:
CI failed

packages/neuron-wallet/tests/controllers/sudt.test.ts
```diff
@@ -41,6 +40,7 @@ describe('SUDTController', () => {
     })
 
     it('getSUDTTokenInfo from token list success', async () => {
+      const SUDTController = require("../../src/controllers/sudt").default
```
katat, line 43: This should be able to `require` once at the top, so you don't need to require multiple times.

katat (approved):

Keith-CY (approved):

PR #1994, opened 2021-01-27
fix: wording for insufficient change capacity

katat:


kellyshang (approved):

Keith-CY:
This message will be used in normal sudt transactions so the "Max" related info is necessary.

katat:
> This message will be used in normal sudt transactions so the "Max" related info is necessary.

A new error class has been created to show the `Max` message when the scenario applies, such as making normal CKB transfer.

https://github.com/nervosnetwork/neuron/pull/1994/files#diff-b2ac8829603fb0f46bec76a6ed71baccff0e4218f43369d278eb8758703fba5bR61

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -241,15 +241,22 @@ public static async signSingleMultiSignScript(
...
+        targetOutputs,
+        changeAddress,
+        fee,
+        feeRate
+      )
 
-    return tx
+      return tx
+    } catch (error) {
+      if (error instanceof CapacityNotEnoughForChange) {
```
Keith-CY, line 255: I'm kinda confused here, it's equal to `if error code is 105, then throw 115`, and the messages for `105` and `115` are swapped. So why not throw `115` and keep messages intact?

katat: > I'm kinda confused here, it's equal to if error code is 105, then throw 115, and the messages for 105 and 115 are swapped.

In this context, it needs to add the `Max` notice for users encountering this `CapacityNotEnoughForChange ` error. The `CapacityNotEnoughForChangeByTransfer` differentiates from the `CapacityNotEnoughForChange` by adding while differentiating its error code to `115` from `105`.

> So why not throw 105 and keep messages intact?
Not quire sure what you mean here. But there are different localizations for the error codes as well: https://github.com/nervosnetwork/neuron/pull/1994/files#diff-4a7a6a60a2b2726c0903290ecc7aa86539a0c2a5d33f06a0d09bfba69ba9d034L489-R499

katat: Keep in mind `CapacityNotEnoughForChange` error will be shown in the other places as well, so that is why the original 105 error was renamed by moving the `max` part, which is not always suitable in the other UIs.

Keith-CY: > Keep in mind `CapacityNotEnoughForChange` error will be shown in the other places as well, so that is why the original 105 error was renamed by moving the `max` part, which is not always suitable in the other UIs.

That's what confuses me, `CapacityNotEnoughForChange'(namely error code = 105) is used elsewhere, so why its message is changed in this PR, are they all need to be updated in the PR?

In this PR:
`"105": "You need more capacities for change (at least 61 CKBytes), or click 'Max' button to send all your balance."`
is changed into
`"105": "You need more capacities for change (at least 61 CKBytes).",`

new message `"115": "You need more capacities for change (at least 61 CKBytes), or click 'Max' button to send all your balance."` is added

`CapacityNotEnoughForChange(105)` thrown in `generateTx` is replaced with `CapacityNotEnoughForChangeByTransfer(115)`

Nothing more than keeping `"105": "You need more capacities for change (at least 61 CKBytes), or click 'Max' button to send all your balance."` and throwing `CapacityNotEnoughForChange(105)` in `generateTx`.

Except that it changes the original message of 105 everywhere.

Is this PR going to change error message for 105 everywhere?

Keith-CY: According to the issue, there's a new kind of error should be thrown in cheque transactions, why not just define a new error with its message and throw it directly

katat: > That's what confuses me, CapacityNotEnoughForChange'(namely error code = 105) is used elsewhere, so why its message is changed in this PR, are they all need to be updated in the PR?

Yes. For example, the cheque transfer in the asset account view could trigger the error 105. If `CapacityNotEnoughForChange` message is not renamed and simplified, then the error message would be inappropriate as @kellyshang pointed out.

The `max` additional notice should be added for the contexts appropriate, so that is why I decided to control and transform the original error 105 at a high level.

> Is this PR going to change error message for 105 everywhere?

Since the original `105` is only appropriate for the normal transfer view, then this message rename for the `105` actually fixes for elsewhere.

Keith-CY: Gotcha

Keith-CY (approved):

PR #2178, opened 2021-11-19
feat: bump sdk to adopt new full format of address

Keith-CY:


Keith-CY:
Bumping `typescript` to v4 in `neuron-ui` requires a huge number of deps to be updated, which should be done in an independent PR, so only `typescript` in `neuron-wallet` is updated here for the SDK@0.101.0

yuche (approved):

kellyshang (approved):

PR #2197, opened 2021-11-29
fix: capature capacity too small error

yuche:


packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -93,7 +93,7 @@ const updateTransactionWith = (generator: typeof generateTx | typeof generateSen
           }
           return res.result
         }
-        if (res.status === 0) {
+        if (res.status === 0 || res.status === 114) {
```
Keith-CY, line 96: Is the message `res.message.content` internationalized?

yuche: Yes.
https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/exceptions/transaction.ts#L16

Keith-CY (approved):

kellyshang (approved):

PR #2237, opened 2022-01-17
Feat/update dependencies

Lester-xie:


Keith-CY:
CI Failed

Keith-CY:
Node.js in Azure should also be set 16 by updating https://github.com/nervosnetwork/neuron/blob/develop/azure-pipelines.yml#L29

Keith-CY:
Duplicated with https://github.com/nervosnetwork/neuron/pull/2243

PR #2257, opened 2022-01-28
fix: fix release timeout

Lester-xie:


Keith-CY (approved):

Lester-xie:
> `npx electron-builder --mac -p always`
The command will hang and never returns when I use `npx`.

kellyshang (approved):

PR #2268, opened 2022-02-15
add redist log

yanguoyu:
add redist log

### DIFF
--- a/packages/neuron-wallet/src/controllers/export-debug.ts
+++ b/packages/neuron-wallet/src/controllers/export-debug.ts
@@ -10,6 +10,7 @@ import { ckbDataPath } from 'services/ckb-runner'
 import NetworksService from 'services/networks'
 import SyncedBlockNumber from 'models/synced-block-number'
 import AddressService from 'services/addresses'
+import redistCheck from 'utils/redist-check'
 
 export default class ExportDebugController {
   #I18N_PATH = 'export-debug-info'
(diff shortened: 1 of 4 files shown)


Keith-CY (approved):

Keith-CY (approved):

packages/neuron-wallet/src/services/node.ts
```diff
@@ -145,7 +143,7 @@ class NodeService {
       return false
     } catch (err) {
       logger.info('CKB:\texternal RPC on default uri not detected, starting bundled CKB node.')
-      const isReadyToStart = await this.detectDependency()
+      const isReadyToStart = await redistCheck()
```
Keith-CY, line 146: The function name could be something like `depsCheck` since `isReadyToStart` doesn't have a strong relationship with `redist`

packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -66,7 +67,9 @@ export default class ExportDebugController {
         .catch(() => ''),
       ckb.rpc
         .getPeers()
-        .catch(() => [])
+        .catch(() => []),
+      redistCheck()
+        .catch(() => false),
```
Keith-CY, line 72: `redistCheck` won't throw.

packages/neuron-wallet/src/services/node.ts
yanguoyu: how about rename ```isReadyToStart``` to ```redistReady```?

Keith-CY (approved):

kellyshang (approved):

PR #2281, opened 2022-02-24
add view sudt detail link

yanguoyu:
image

### DIFF
--- a/packages/neuron-wallet/src/controllers/sudt.ts
+++ b/packages/neuron-wallet/src/controllers/sudt.ts
@@ -1,3 +1,4 @@
+import { scriptToHash } from "@nervosnetwork/ckb-sdk-utils"
 import LiveCellService from "services/live-cell-service";
 import AssetAccountInfo from "../models/asset-account-info"
 import Script, {ScriptHashType} from "models/chain/script";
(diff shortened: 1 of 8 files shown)


packages/neuron-wallet/src/controllers/sudt.ts
```diff
@@ -32,5 +33,18 @@ export default class SUDTController {
       result: {tokenID: params.tokenID, symbol: symbol, tokenName: name, decimal: decimal},
     }
   }
+
+  public getSUDTTypeScriptHash(params: { tokenID: string }): Controller.Response<string> {
```
Keith-CY, line 37: This calculation could be done in the renderer process so that the main process can handle other things at the same time.

yanguoyu: so i should add a const `sudt.codeHash` in the renderer process?

Keith-CY: > so i should add a const `sudt.codeHash` in the renderer process?

I see, it's ok to fetch the result if `code hash` and `hash type` are missing in the renderer process.

Keith-CY:
The editor dialog should also have a button to open the explorer
image

Keith-CY:
The entire dialog shakes when the button appears.

https://user-images.githubusercontent.com/7271329/155510133-f0288fa4-04e3-4b54-a97d-f38a5ac483c5.mov

Keith-CY:
If the button has no animation like the one in txns history, it could have a shadow on hovering.

https://user-images.githubusercontent.com/7271329/155510559-2f3d8f61-e4d6-43c0-b0bf-dc81e05ce941.mov

Keith-CY:
CKB Account shouldn't have a dummy button
image

yanguoyu:
> The editor dialog should also have a button to open the explorer image

it's has created account when edit the sudt account , it also need show the explorer?

Keith-CY:
The [SUDTUpdateDialog](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SUDTUpdateDialog/index.tsx) should have the link too.

yanguoyu:
> The [SUDTUpdateDialog](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SUDTUpdateDialog/index.tsx) should have the link too.

ok, i have add it

Keith-CY (approved):

PR #2289, opened 2022-03-02
Chore/prettier format code

Lester-xie:


packages/neuron-wallet/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "trailingComma": "none",
+  "semi": false,
+  "singleQuote": true,
+  "semicolons:": false
```
Keith-CY, line 5: `semicolons` in not a valid config of prettier

PR #2306, opened 2022-03-18
feat: add warn text when send transaction has set locktime

yanguoyu:
image

### DIFF
--- a/packages/neuron-ui/src/components/SendFieldset/index.tsx
+++ b/packages/neuron-ui/src/components/SendFieldset/index.tsx
@@ -5,6 +5,7 @@ import { ckbCore } from 'services/chain'
 
 import TextField from 'widgets/TextField'
 import Button from 'widgets/Button'
+import { MessageBar, MessageBarType } from 'office-ui-fabric-react'
 
 import { ReactComponent as Scan } from 'widgets/Icons/Scan.svg'
 import AddOutput from 'widgets/Icons/AddOutput.png'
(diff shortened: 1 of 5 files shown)


packages/neuron-ui/src/locales/en.json
```diff
@@ -209,7 +209,8 @@
       "scan-screen-qr-code": "Scan the QR Code on the screen",
       "set-locktime": "Set Locktime",
       "locktime-notice-content": "According to the actual running block height, there may be some time variances in locktime.",
-      "release-on": "Release on"
+      "release-on": "Release on",
+      "locktime-warning": "Please ensure receiver's wallet can support expiration unlocking. In general, exchanges do not support expiration unlocking, please use with caution!"
```
Keith-CY, line 213:
```suggestion
      "locktime-warning": "Please ensure that receiver's wallet can support expiration unlocking. In general, exchanges do not support expiration unlocking, please use with caution!"
```

packages/neuron-ui/src/components/SendFieldset/index.tsx
```diff
@@ -5,6 +5,7 @@ import { ckbCore } from 'services/chain'
 
 import TextField from 'widgets/TextField'
 import Button from 'widgets/Button'
+import { MessageBar, MessageBarType } from 'office-ui-fabric-react'
```
Keith-CY, line 8: We'll remove `official ui fabric react` in the future, so better not to use its components here.

Keith-CY (approved):

EmmaZHONG1024 (approved):

PR #2307, opened 2022-03-18
feat:Disable some functions for xpub wallet.

yanguoyu:
1. sign messages
   image
2. send transactions
image
image
3. backup wallets
image


packages/neuron-ui/src/locales/en.json
```diff
@@ -408,7 +408,8 @@
       },
       "destroy-ckb-account": {
         "title": "Destroy CKB Asset Account"
-      }
+      },
+      "xpub-notice": "For Extended Public Key wallets, only export transactions are supported"
```
Keith-CY, line 412:
```suggestion
      "xpub-notice": "For Extended Public Key wallets, only exporting transactions are supported"
```

Keith-CY: Usually, we call the wallet imported by a `public key` a `watch-only wallet`

packages/neuron-ui/src/locales/zh-tw.json
```diff
@@ -401,7 +401,8 @@
       },
       "destroy-ckb-account": {
         "title": "銷毀 CKB 資產賬戶"
-      }
+      },
+      "xpub-notice": "對於 Extended Public Key 錢包，只支持導出交易"
```
Keith-CY, line 405: 通过公钥导入的钱包一般称为 `观察钱包`

packages/neuron-ui/src/services/remote/remoteApiWrapper.ts
```diff
@@ -61,6 +61,7 @@ type Action =
   | 'validate-mnemonic'
   | 'sign-message'
   | 'verify-signature'
+  | 'is-wallet-xpub'
```
Keith-CY, line 64: could be renamed `is-watch-only-wallet` or something concise.

packages/neuron-ui/src/services/remote/wallets.ts
```diff
@@ -19,6 +19,7 @@ export const generateTx = remoteApi<Controller.GenerateTransactionParams>('gener
 export const generateSendingAllTx = remoteApi<Controller.GenerateSendingAllTransactionParams>('generate-send-all-tx')
 export const generateMnemonic = remoteApi<void>('generate-mnemonic')
 export const validateMnemonic = remoteApi<string>('validate-mnemonic')
+export const isWalletXpub = remoteApi<string, boolean>('is-wallet-xpub')
```
Keith-CY, line 22: Could this be returned with `get-current-wallet`?

yanguoyu: Sometimes only need to know current wallet is watch-only-wallet

Keith-CY: Indeed, this property is just used in several cases.

But adding one more API makes the system more complex, meanwhile, the `isWalletXpub` returned along with `get-current-wallet` could be stored in [state](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/states/init/wallet.ts#L12) and there's no need to fetch `isWalletEpub` asynchronously.

yanguoyu: Of cause stored it is good idea.

EmmaZHONG1024:
测试通过
截屏2022-03-22 上午9 47 12
截屏2022-03-22 上午9 48 06
截屏2022-03-22 上午9 52 47

packages/neuron-ui/src/components/SignAndVerify/index.tsx
```diff
@@ -181,11 +183,16 @@ const SignAndVerify = () => {
 
     getAddressesByWalletID(id).then(res => {
       if (isSuccessResponse(res)) {
-        setWallet({ id, addresses: res.result })
+        setWallet(v => ({ ...v, id, addresses: res.result }))
       } else {
         showErrorMessage(t('messages.error'), typeof res.message === 'string' ? res.message : res.message.content!)
       }
     })
+    getCurrentWallet().then(res => {
```
Keith-CY, line 191: With these codes, we can remove the `getAddressesByWalletID` call above, since the addresses are limited to the current wallet and they are returned by `getCurrentWallet`, even more, the `current wallet` has been returned by `useGlobalState()` [line 164]

yanguoyu: should i delete `const currentWallet = useMemo(() => wallets.find(w => w.id === wallet?.id), [wallet, wallets])` by instead of `state.wallet`

Keith-CY: I think it's fine to delete this hook
I've confirmed 2 points:
1. the `id` in URL is bound to the current wallet on opening the window and won't change https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/app/menu.ts#L304;
2. the `wallet` in sign-and-verify window is set by `id` in the URL;
3. the sign-and-verify window will close when the current wallet changes https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L168

so there's no need to update the `wallet` in sign-and-verify window again.

Keith-CY: I've found something new

There are 2 `current wallet` in this component, `wallet` in https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L158 and `currentWallet` in https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L166
They may be different since `currentWallet` is expected to have `device` property https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L339

yanguoyu: `state.wallet` is same as  `currentWallet`,they don't have `addresses`.So i should use `state.wallet` instead of `currentWallet` and retain `wallet` in https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L158 . Or i return `addresses` by `get-current-wallet`, then i can only use `state.wallet` in this component.

Keith-CY: > `state.wallet` is same as `currentWallet`,they don't have `addresses`.So i should use `state.wallet` instead of `currentWallet` and retain `wallet` in [https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L158](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx?rgh-link-date=2022-03-22T06%3A33%3A56Z#L158) . Or i return `addresses` by `get-current-wallet`, then i can only use `state.wallet` in this component.

I'm not sure which one is specified as `state.wallet` that don't have `addresses`, the one returned by `useGlobalState` has `addresses` but has no `device`  property while the `settings.wallets` returned by `useGlobalState` has the `device` property. I think that's the reason `currentWallet` is filtered from `settings.wallets`

yanguoyu: Typeof `state.wallet` include `addresses`, but actual it has no `addresses`. I think why `State.Wallet` has addresses property because used at [this](https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L158). The `state.wallet` returned by `get-current-wallet` or `load-init-data` all don't has `addresses`.Because they all returned by https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-wallet/src/services/wallets.ts#L338. The `Wallet` class has no addresses property.So i think i should add `addresses`.

`state.wallet` actual has `device` property when create Hardware wallet at https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-wallet/src/controllers/wallets.ts#L258

Keith-CY: My point may be misunderstood.

`addresses` could be fetched from `wallet` returned by `useGlobalState()`(https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L162-L164) so there's no need to use `getAddressesByWalletID` to fetch it again.

But the `wallet` returned by `useGlobalState` doesn't have the `device` property. However `settings.wallets` returned by `useGlobalState()` have, so the `device` property should be fetched by `settings.wallets.filter(w => w.id === wallet.id)?.device`

yanguoyu: `The wallet returned by useGlobalState() doesn't have the device property.` I don't have hardware wallet, i think it has device property by `useGlobalState`.The device property has saved when create wallet. https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-wallet/src/controllers/wallets.ts#L258.

Keith-CY: > Typeof `state.wallet` include `addresses`, but actual it has no `addresses`

I missed the first point.

Actually `state.wallet` has `addresses`, I can log in the console

Keith-CY: > `The wallet returned by useGlobalState() doesn't have the device property.` I don't have hardware wallet, i think it has device property by `useGlobalState`.The device property has saved when create wallet.
> 
> https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-wallet/src/controllers/wallets.ts#L258
> 
> .

Indeed, the `wallet` returned by `useGlobalState` has the `device` property and will be instantiated as `HardwareWallet` if `device` exists.

yanguoyu: > 

sorry, I missed the code.`wallet.addresses` assign by https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-ui/src/states/stateProvider/actionCreators/app.ts#L31 or https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-ui/src/states/init/wallet.ts#L16

yanguoyu: > > `The wallet returned by useGlobalState() doesn't have the device property.` I don't have hardware wallet, i think it has device property by `useGlobalState`.The device property has saved when create wallet.
> > https://github.com/nervosnetwork/neuron/blob/33bb20311fdce1e99803104a48492b8863eadaba/packages/neuron-wallet/src/controllers/wallets.ts#L258
> > 
> > .
> 
> Indeed, the `wallet` returned by `useGlobalState` has the `device` property and will be instantiated as `HardwareWallet` if `device` exists.

yes, it's really has `device` property.So i can delete https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SignAndVerify/index.tsx#L166 and use `useGlobalState().wallet` to replace
image

Keith-CY (approved):

Keith-CY:
Conflicted

yanguoyu:
> Conflicted

Resolved

PR #2310, opened 2022-03-21
feat:Show multisig balance

yanguoyu:


Keith-CY:
Please append screenshots or recordings

yanguoyu:
image

packages/neuron-wallet/src/controllers/app/index.ts
```diff
@@ -52,11 +62,21 @@ export default class AppController {
     await Promise.all([stopCkbNode(), IndexerService.getInstance().stop()])
   }
 
+  public registerChannels(winId: number, channels: string[]) {
```
Keith-CY, line 65: There's no need to `unregisterChannel`?

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -1054,4 +1055,39 @@ export default class CellsService {
 
     return outputEntities
   }
+
+  public static async getMultisigBalance(isMainnet: boolean) {
```
Keith-CY, line 1059: If `args` list is not specified in the query, we should add extra logic to guarantee cells of removed addresses are removed at the same time. Or the query may be consuming.

packages/neuron-wallet/tests/controllers/multisig.test.ts
```diff
@@ -195,4 +199,10 @@ describe('test for multisig controller', () => {
       expect(res?.status).toBe(ResponseCode.Success)
     })
   })
+
+  it('getMultisigBalances', async () => {
+    const res = await multisigController.getMultisigBalances(false)
```
Keith-CY, line 204: expected `getMultisigBalance` to be called

packages/neuron-wallet/src/controllers/multisig.ts
```diff
@@ -216,4 +217,12 @@ export default class MultisigController {
       }
     }
   }
+
+  async getMultisigBalances(isMainnet: boolean) {
+    const balances = await CellsService.getMultisigBalance(isMainnet)
```
Keith-CY, line 222: The outside method is named `getMultisigBalances` while the inner one is named `getMultisigBalance`

Keith-CY:
We could adjust the list of multisig addresses left-aligned just as the address book

packages/neuron-wallet/src/controllers/app/index.ts
yanguoyu: if the window destroyed, the window id will not be find.

packages/neuron-wallet/src/services/cells.ts
yanguoyu: I want't to get all live multisig address balance.Even if the addresses has removed,I can also add it again.It's slowly if search all live multisig addresses?

packages/neuron-wallet/src/controllers/app/index.ts
Keith-CY: In this case, a `WeakMap` could be used to release the `channels` automatically

packages/neuron-wallet/src/services/cells.ts
Keith-CY: Without cleaning, the data will accumulate by adding new addresses. It would be disturbing to users who often approve transactions.

yanguoyu: ok, I will add added addresses list as a parameter

packages/neuron-wallet/src/controllers/app/index.ts
yanguoyu: If use `WeakMap`, the WeakMap key type should be `Electron.BrowserWindow`,change like this?
`private windowRegisterChannels = new Map<Electron.BrowserWindow, string[]>()`

Keith-CY: Is `new Map<Electron.BrowserWindow, string[]>()` a typo? I think it's OK to use `new WeakMap<Electron.BrowserWindow, string[]>()` if window instance is a singleton

EmmaZHONG1024 (approved):

packages/neuron-wallet/src/controllers/multisig.ts
```diff
@@ -218,8 +218,9 @@ export default class MultisigController {
     }
   }
 
-  async getMultisigBalances(isMainnet: boolean) {
-    const balances = await CellsService.getMultisigBalances(isMainnet)
+  async getMultisigBalances({ isMainnet, multisigAddresses}: { isMainnet: boolean , multisigAddresses: string[] }) {
+    console.log(CellsService)
```
Keith-CY, line 222: `console.log` should not be committed.

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -1067,9 +1071,8 @@ export default class CellsService {
         from
             output
         where
-            output.lockCodeHash = '${SystemScriptInfo.MULTI_SIGN_CODE_HASH}' AND
-            output.status in ('${OutputStatus.Live}', '${OutputStatus.Sent}') AND
-            output.lockHashType = '${SystemScriptInfo.MULTI_SIGN_HASH_TYPE}'
+            output.lockHash in (${lockHashes.map(v => `'${v}'`).join(',')}) AND
```
Keith-CY, line 1074: For outside data source, we should use data interpolation like
```
output.lockHash IN (:...lockHashes)
...
{ lockHashes: [...data] }
```
to escape data

packages/neuron-ui/src/components/MultisigAddress/index.tsx
```diff
@@ -121,7 +121,7 @@ const MultisigAddress = () => {
                   &nbsp;of&nbsp;
                   {v.n}
                 </td>
-                <td>
```
Keith-CY, line 124: Actually all columns are left-aligned, but we can do this later.

packages/neuron-wallet/src/services/cells.ts
yanguoyu: `EntityManager.query` parameter is different with `SelectQueryBuilder .where`

Keith-CY: Could we use `escapeQueryWithParameters` to avoid this?
Ref: https://github.com/typeorm/typeorm/issues/556#issuecomment-1049587900

Keith-CY (approved):

PR #2317, opened 2022-04-01
fix: Avoid duplicate addresses. Disabled auto resize table in dialog.

yanguoyu:


packages/neuron-ui/src/locales/en.json
```diff
@@ -882,7 +882,8 @@
           "back": "Back",
           "next": "Next",
           "confirm": "Confirm"
-        }
+        },
+        "duplicate-address-forbid": "Address can't be duplicate"
```
Keith-CY, line 886:
```suggestion
        "duplicate-address-forbid": "Address can't be duplicated"
```

Keith-CY (approved):

Keith-CY (approved):

PR #2320, opened 2022-04-02
refactor: refactor script tags into a component

Keith-CY:
This PR adds a component named `ScriptTag` to append a tag next to addresses in the transaction detail page
image

image
image
image

PS: cheque script is ignored

---
Update
Locktime is appended
image


yanguoyu (approved):

packages/neuron-ui/src/components/Transaction/index.tsx
```diff
@@ -178,14 +156,41 @@ const Transaction = () => {
         }
         const capacity = shannonToCKBFormatter(cell.capacity || '0')
 
+        // REFACTOR: should be refacter into a component
```
yanguoyu, line 159: I think it can be move to outside of the function component first.

Keith-CY: Indeed, I'll update it within this PR.

Keith-CY: > I think it can be move to outside of the function component first.

Indeed, I'll update it within this PR.

Keith-CY: Updated

yanguoyu:
Is it necessary to consider transfer with time lock?

yanguoyu:
Is it necessary to consider transfer with time lock?

Keith-CY:
> Is it necessary to consider transfer with time lock?

Is this related to https://github.com/nervosnetwork/neuron/pull/2313 but not this PR?
If yes, we can ignore timelock for now.

yanguoyu:
> > Is it necessary to consider transfer with time lock?
> 
> Is this related to #2313 but not this PR? If yes, we can ignore timelock for now.

No.Send CKB like this.

image

image

Keith-CY:
> > > Is it necessary to consider transfer with time lock?
> > 
> > 
> > Is this related to #2313 but not this PR? If yes, we can ignore timelock for now.
> 
> No.Send CKB like this.
> 
> image
> 
> image

Nice catch, we could append `timelock` here.

Keith-CY:
> > > Is it necessary to consider transfer with time lock?
> > 
> > 
> > Is this related to #2313 but not this PR? If yes, we can ignore timelock for now.
> 
> No.Send CKB like this.
> 
> image
> 
> image

Locktime script tag is appended

yanguoyu (approved):

PR #2323, opened 2022-04-06
feat: Add send max when create multisig transaction

yanguoyu:
https://user-images.githubusercontent.com/12881040/161878584-7dad5ade-f6ea-42a1-886b-e1a6b7162dcf.mov

### DIFF
--- a/packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
+++ b/packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
@@ -1,57 +1,79 @@
 import { useCallback, useState, useMemo, useEffect } from 'react'
-import { useOutputErrors, outputsToTotalAmount, CapacityUnit, validateOutputs, CKBToShannonFormatter } from 'utils'
+import {
+  useOutputErrors,
+  outputsToTotalAmount,
+  CapacityUnit,
+  validateOutputs,
+  CKBToShannonFormatter,
+  shannonToCKBFormatter,
+} from 'utils'
 import { useDispatch } from 'states'
 import { AppActions, StateDispatch } from 'states/stateProvider/reducer'
-import { generateMultisigTx, MultisigConfig } from 'services/remote'
+import { generateMultisigTx, MultisigConfig, generateMultisigSendAllTx } from 'services/remote'
 import { TFunction } from 'i18next'
 
 let generateTxTimer: ReturnType<typeof setTimeout>
+Object.defineProperty(generateMultisigTx, 'type', {
+  value: 'common',
+})
+Object.defineProperty(generateMultisigSendAllTx, 'type', {
+  value: 'all',
+})
 
-const generateMultisigTxWith = ({
+const generateMultisigTxWith = (generator: typeof generateMultisigTx | typeof generateMultisigSendAllTx) => ({
   sendInfoList,
   multisigConfig,
   dispatch,
   setErrorMessage,
   t,
+  isMainnet,
 }: {
   sendInfoList: { address: string | undefined; amount: string | undefined; unit: CapacityUnit }[]
   multisigConfig: MultisigConfig
   dispatch: StateDispatch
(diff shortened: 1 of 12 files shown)


packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
```diff
@@ -118,28 +140,76 @@ export const useSendInfo = ({
   const dispatch = useDispatch()
   const [errorMessage, setErrorMessage] = useState('')
   useEffect(() => {
-    if (outputErrors.some(v => v.addrError || v.amountError)) {
-      return
-    }
     clearTimeout(generateTxTimer)
     setErrorMessage('')
-    const validSendInfoList = sendInfoList.filter(v => v.address && v.amount)
+    const validSendInfoList = sendInfoList.filter(
```
Keith-CY, line 145: If addresses and amounts will be verified in `generateMultisigTxWith`, why should we filter invalid addresses and amounts?

packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
```diff
@@ -118,28 +140,76 @@ export const useSendInfo = ({
...
         })
       }
     }, 300)
-  }, [sendInfoList, setErrorMessage, multisigConfig, dispatch, t, outputErrors])
+  }, [sendInfoList, setErrorMessage, multisigConfig, dispatch, t, isMainnet, outputErrors])
+  const [isSendMax, setIsSendMax] = useState(false)
+  const onSendMaxClick = useCallback(
+    e => {
+      const {
+        dataset: { isOn = 'false' },
```
Keith-CY, line 169: What does `isOn` mean?

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -620,16 +621,21 @@ export default class CellsService {
         output.status IN (:...statuses) AND
         hasData = false AND
         typeHash is null AND
-        output.lockArgs in (
-          SELECT publicKeyInBlake160
-          FROM hd_public_key_info
-          WHERE walletId = :walletId
-        ) AND
+        ${
```
Keith-CY, line 624: Tests should be updated

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -440,6 +440,26 @@ export default class TransactionSender {
     return tx
   }
 
+  public generateMultisigSendAllTx = async (
```
Keith-CY, line 443: Tests are missing

packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
yanguoyu: It equal with the `isSendMax`

packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
yanguoyu: It avoid the valid outputs's length is 0, but it also has transaction fee.

Keith-CY:
image
image
I triggered an error when synchronization is not fully done.

packages/neuron-ui/src/components/SendFromMultisigDialog/hooks.ts
Keith-CY: What does `setIsSendMax(isOn === 'false')` mean if `isSendMax` equals to `isSendMax`

Keith-CY: Ah, I understand it, when `isOn` is `false`, by clicking the button, `isSendMax` is set to `true`

yanguoyu:
> image image I triggered an error when synchronization is not fully done.

Fixed by judging whether `res` is undefined

Keith-CY (approved):

PR #2332, opened 2022-04-15
feat: Sync multisig output cell.

yanguoyu:


Keith-CY:
Balances are not cleared by `clear cache`

packages/neuron-wallet/src/utils/rpc-request.ts
```diff
@@ -0,0 +1,31 @@
...
+    params?: any
+  }[]
+): Promise<any[]> => {
+  const res = await axios({
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json'
+    },
+    data: options.map(v => ({
+      id: 0,
```
Keith-CY, line 16: `id` should be different in the array, it is used to correlate request and response among the request array and response array.

yanguoyu:
> Balances are not cleared by `clear cache`

Fixed it by https://github.com/nervosnetwork/neuron/pull/2332/commits/07038ea658c87a59b0314f3d2df7ffc15da418fa

Keith-CY (approved):

packages/neuron-wallet/src/services/multisig.ts
```diff
@@ -62,16 +71,219 @@ export default class MultisigService {
...
+              addressCursorMap.get(v.fullPayload)
+            ]
+          }
+        })
+      )
+      const nextMultisigConfigs: MultisigConfig[] = []
+      res.forEach((v, idx) => {
+        if (!v.error && v?.result?.objects?.length) {
+          addressCursorMap.set(currentMultisigConfigs[idx].fullPayload, v?.result?.last_cursor)
+          liveCells.push(...v.result.objects.map((output: any) => MultisigOutput.fromIndexer(output)))
```
Keith-CY, line 124: cells with `type script` should be eliminated

yanguoyu: https://github.com/nervosnetwork/neuron/pull/2340

PR #2342, opened 2022-04-24
feat: Mark create and destroy asset account

qiweiii:
Closes: https://github.com/Magickbase/neuron-public-issues/issues/10


packages/neuron-ui/src/components/Overview/index.tsx
```diff
@@ -134,17 +137,26 @@ const Overview = () => {
...
               item.sudtInfo.sUDT.symbol
             }`
           }
         } else {
           amount = `${shannonToCKBFormatter(item.value)} CKB`
-          typeLabel = item.nervosDao ? 'Nervos DAO' : t(`overview.${genTypeLabel(item.type, status)}`)
+          if (item.type === 'create' || item.type === 'destroy') {
+            if (item.assetAccountType === 'CKB') {
+              typeLabel = `${t(`overview.${item.type}`, { name: 'CKB' })}`
+            }
```
Keith-CY, line 156: This else arm is not covered, maybe we can name it `Unknown`(it's reserved) so when `create/destroy Unknown Asset Account` does appear we can ensure this arm is triggered.

packages/neuron-ui/src/components/TransactionList/index.tsx
```diff
@@ -165,7 +169,13 @@ const TransactionList = ({
           name = walletName
           value = `${tx.value} shannons`
           amount = `${shannonToCKBFormatter(tx.value, true)} CKB`
-          typeLabel = tx.nervosDao ? 'Nervos DAO' : t(`history.${tx.type}`)
+          if (tx.type === 'create' || tx.type === 'destroy') {
+            if (tx.assetAccountType === 'CKB') {
+              typeLabel = `${t(`history.${tx.type}`, { name })}`
+            }
```
Keith-CY, line 175: Same as above

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -427,12 +446,17 @@ export class TransactionsService {
           nftInfo = { type: NFTType.Receive, data: receiveNFTCell.typeArgs! }
         }
 
+        if (!txType) {
+          txType = value > BigInt(0) ? 'receive' : 'send'
```
yanguoyu, line 450: `txType` can be initialized at line 361 `let txType: string = ''`

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -358,20 +358,39 @@ export class TransactionsService {
...
-            o => o.outPointTxHash === tx.hash && assetAccountInfo.isSudtScript(o.typeScript()!)
-          )
           if (sudtOutput) {
             typeArgs = sudtOutput.typeArgs
+            txType = 'create'
+            assetAccountType = AssetAccountType.SUDT
           }
         }
 
+        const ckbInput = anyoneCanPayInputs.find(i => i.transactionHash === tx.hash && !i.typeScript())
```
yanguoyu, line 384: `anyoneCanPayInputs` must include `type script`, because it's searching conditions include  `input.typeHash IS NOT NULL`.https://github.com/nervosnetwork/neuron/blob/50e112e67b9de1e679a8e8fd5a4a68c12557f17b/packages/neuron-wallet/src/services/tx/transaction-service.ts#L259

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -358,20 +358,39 @@ export class TransactionsService {
...
+        const ckbOutput = anyoneCanPayOutputs.find(o => o.outPointTxHash === tx.hash && !o.typeScript())
```
yanguoyu, line 385: same as `anyoneCanPayInputs`

packages/neuron-wallet/src/models/chain/transaction.ts
```diff
@@ -72,6 +77,7 @@ export default class Transaction {
   public nftInfo?: NFTInfo
 
   public signatures: Signatures = {}
+  public assetAccountType?: AssetAccountType | string = '' // Default to ''
```
yanguoyu, line 80: How about only use `AssetAccountType`? It can be `AssetAccountType` or undefined.

packages/neuron-ui/src/components/TransactionList/index.tsx
```diff
@@ -165,7 +169,13 @@ const TransactionList = ({
           name = walletName
           value = `${tx.value} shannons`
           amount = `${shannonToCKBFormatter(tx.value, true)} CKB`
-          typeLabel = tx.nervosDao ? 'Nervos DAO' : t(`history.${tx.type}`)
+          if (tx.type === 'create' || tx.type === 'destroy') {
+            if (tx.assetAccountType === 'CKB') {
+              typeLabel = `${t(`history.${tx.type}`, { name })}`
```
yanguoyu, line 174: `CKB`'s label name may be use `CKB`?

packages/neuron-wallet/src/services/tx/transaction-service.ts
qiweiii: Just to clarify, I use `anyoneCanPayInputs` because @Keith-CY's [comment](https://github.com/Magickbase/neuron-public-issues/issues/10#issuecomment-1107784182) says that 

> CKB Asset Account's type script is null, so anyone-can-pay input could have no type scripts.

Maybe I misunderstood this part sentence..😅

So maybe what I should do is to add a query like this:

```js
const assetAccountInputs = await connection
      .getRepository(InputEntity)
      .createQueryBuilder('input')
      .where(
        `
        input.transactionHash IN (:...txHashes) AND
        input.lockArgs in (select publicKeyInBlake160 from hd_public_key_info where walletId = :walletId) AND
        input.lockCodeHash = :lockCodeHash`,
        {
          txHashes,
          walletId: params.walletID,
          lockCodeHash: assetAccountInfo.anyoneCanPayCodeHash
        }
      )
      .getMany()
// then
const anyoneCanPayInputs = assetAccountInputs.filter(i => i.typeScript()) // so it has a typescript
const ckbAssetInput = assetAccountInputs.filter(i => !i.typeScript())

// similar for outputs
```

Is this correct @yanguoyu  @Keith-CY?

Keith-CY: > Just to clarify, I use `anyoneCanPayInputs` because @Keith-CY's [comment](https://github.com/Magickbase/neuron-public-issues/issues/10#issuecomment-1107784182) says that
> 
> > CKB Asset Account's type script is null, so anyone-can-pay input could have no type scripts.
> 
> Maybe I misunderstood this part sentence..😅
> 
> So maybe what I should do is to add a query like this:
> 
> ```js
> const assetAccountInputs = await connection
>       .getRepository(InputEntity)
>       .createQueryBuilder('input')
>       .where(
>         `
>         input.transactionHash IN (:...txHashes) AND
>         input.lockArgs in (select publicKeyInBlake160 from hd_public_key_info where walletId = :walletId) AND
>         input.lockCodeHash = :lockCodeHash`,
>         {
>           txHashes,
>           walletId: params.walletID,
>           lockCodeHash: assetAccountInfo.anyoneCanPayCodeHash
>         }
>       )
>       .getMany()
> // then
> const anyoneCanPayInputs = assetAccountInputs.filter(i => i.typeScript()) // so it has a typescript
> const ckbAssetInput = assetAccountInputs.filter(i => !i.typeScript())
> 
> // similar for outputs
> ```
> 
> Is this correct @yanguoyu @Keith-CY?

Exactly, the previous select clause only gets sUDT Asset Account Input/Output since it requires `type` not to be empty.

In this PR, we're going to get sUDT/CKB Asset Account, so the condition `type hash is not null` should be eliminated.

packages/neuron-wallet/src/services/tx/transaction-service.ts
yanguoyu: I means initialized like this: `txType = value > BigInt(0) ? 'receive' : 'send'`.Then you can delete this code.

qiweiii: yes, that make sense

Keith-CY:
CI failed https://github.com/nervosnetwork/neuron/runs/6170319262?check_suite_focus=true#step:8:1046

yanguoyu (approved):

qiweiii:
Tests fixed. Btw there are lots of changes by prettier, making it a bit difficult to review.

Keith-CY (approved):

PR #2344, opened 2022-04-25
fix: Delete unnecessary conditions. Retry should await.

yanguoyu:


packages/neuron-wallet/src/utils/common.ts
```diff
@@ -10,13 +10,13 @@ export default class CommonUtils {
 
     while (++retryTime < times) {
       try {
-        return callback()
+        return await callback()
```
Keith-CY, line 13: `await` is redundant in `return statement`

yanguoyu: The exception can't be caught if the callback's result is a promise without await.

Keith-CY: gotcha

Keith-CY (approved):

PR #2346, opened 2022-04-26
feat: set new version of full address default

Keith-CY:
1. update addresses in address book
2. update address in receive panel
3. update addresses in transaction detail
4. accept new address in send panel
5. support searching with new address
6. update addresses of asset accounts
7. accept new address in asset account send panel
8. update addresses in sign/verify window
9. keep addresses in multisig addresses window since they are
new already
10. update existing address field in address_description table

Ref: https://github.com/Magickbase/neuron-public-issues/issues/12

### Receive Panel
image

### Address Book
image

### Transaction Detail
image

image

### Accept new address in send panel
image

### Search
image

### Asset Account
image

image

### Accept new address in asset account send
image

### Sign/verify
image


yanguoyu:
should [this address property](https://github.com/nervosnetwork/neuron/blob/05f0f7481abba4c35fef2e291e493d9bdb1ea94a/packages/neuron-wallet/src/database/chain/entities/indexer-tx-hash-cache.ts#L27) migrate?
I gain enlightenment from this PR, we should only save blake160, but not address. Should i refactor save blake160s in `multisig-config` entities.

yanguoyu (approved):

Keith-CY:
> should [this address property](https://github.com/nervosnetwork/neuron/blob/05f0f7481abba4c35fef2e291e493d9bdb1ea94a/packages/neuron-wallet/src/database/chain/entities/indexer-tx-hash-cache.ts#L27) migrate? I gain enlightenment from this PR, we should only save blake160, but not address. Should i refactor save blake160s in `multisig-config` entities.

It would be good to remove `address` from the table since it's flexible in format, ckb explorer also suffers a lot due to storing address in database directly.

PR #2351, opened 2022-04-28
fix: Avoid save duplicate blake160 for one wallet

yanguoyu:


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -27,14 +27,33 @@ export default class AddressService {
...
-      return HdPublicKeyInfo.fromObject({
-        ...addr,
-        publicKeyInBlake160: addr.blake160
+    const walletIds = addresses.map(v => v.walletId)
+    const walletIdMaxAddressIndex = await getConnection()
+      .getRepository(HdPublicKeyInfo)
+      .createQueryBuilder()
+      .select('walletId')
+      .addSelect('group_concat(publicKeyInBlake160)', 'publicKeyInBlake160Array')
+      .where({ walletId: In(walletIds) })
```
Keith-CY, line 36: This could be simplified to `walletId = walletId` since all addresses belong to the same wallet.

To ensure this, we can add an assert in this method.

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -27,14 +27,33 @@ export default class AddressService {
...
+      .getRawMany()
+    const walletIdMaxAddressIndexMap: Record<string, string> = walletIdMaxAddressIndex.reduce(
```
Keith-CY, line 38: Same as above.

Keith-CY:
Could this check https://github.com/nervosnetwork/neuron/blob/50e112e67b9de1e679a8e8fd5a4a68c12557f17b/packages/neuron-wallet/src/services/addresses.ts#L158-L169 be removed in this PR?

yanguoyu:
> Could this check
> 
> https://github.com/nervosnetwork/neuron/blob/50e112e67b9de1e679a8e8fd5a4a68c12557f17b/packages/neuron-wallet/src/services/addresses.ts#L158-L169
> 
> be removed in this PR?

This function is saving an address. I think it should be retained

Keith-CY:
> > Could this check
> > https://github.com/nervosnetwork/neuron/blob/50e112e67b9de1e679a8e8fd5a4a68c12557f17b/packages/neuron-wallet/src/services/addresses.ts#L158-L169
> > 
> > be removed in this PR?
> 
> This function is saving an address. I think it should be retained

Gotcha

Keith-CY (approved):

PR #2352, opened 2022-04-28
fix: Better display for sync status

yanguoyu:
Setting assume-valid-target `0x7544e2a9db2054fbe42215ece2e5d31f175972cfeccaa7597c8ff3ec5c8b7d67`

image

image

image


packages/neuron-wallet/src/controllers/sync-api.ts
```diff
@@ -224,6 +227,14 @@ export default class SyncApiController {
     SyncApiController.emiter.on('cache-tip-block-updated', async states => {
       const newSyncState = await this.#estimate(states)
       this.#syncedBlockNumber.setNextBlock(BigInt(newSyncState.cacheTipNumber))
+      if (
+        env.app.isPackaged &&
+        process.env.CKB_NODE_ASSUME_VALID_TARGET &&
+        (newSyncState.cacheTipNumber === 0 || newSyncState.cacheTipNumber === this.#lastCacheTipNumber)
```
Keith-CY, line 233: Would `newSyncState.cacheTipNumber === this.#lastCacheTipNumber` be true when synchronization is done?

yanguoyu: That's right, it shouldn't equal bestKnownBlockNumber

yanguoyu: If synchronization is done, it wouldn't subscribe `cache-tip-block-updated` event?

Keith-CY: > If synchronization is done, it wouldn't subscribe `cache-tip-block-updated` event?

It keeps subscribing to `cache-tip-block-updated` since `synchronization is done` is temporary.

yanguoyu: I think I should subscribe ckb logs to setting status. I will reimplement it

yanguoyu: I fixed it with validate `bestKnownBlockTimestamp` is within 24 hours

qiweiii (approved):

packages/neuron-wallet/src/services/ckb-runner.ts
```diff
@@ -71,20 +75,40 @@ export const startCkbNode = async () => {
...
     })
+  ckb.stdout &&
+    ckb.stdout.on('data', data => {
+      if (!app.isPackaged || !process.env.CKB_NODE_ASSUME_VALID_TARGET) {
+        return
+      }
+      const dataString: string = data.toString()
+      if (
+        dataString.includes(
+          `can't find assume valid target temporarily, hash: Byte32(${process.env.CKB_NODE_ASSUME_VALID_TARGET})`
```
Keith-CY, line 93: Better to check `can't find assume valid target temporarily, hash: Byte32(\w{66})` and extract the `assume valid target` from logs. Thus users can use their own CKB nodes with their specified `assume valid target` and get the correct info from Neuron.

yanguoyu: Is it has an rpc API to get the node logs or get the node running directory?
If someone starts local nodes with `assume-valid-target`, he should know what's means.

Keith-CY: > Is it has an rpc API to get the node logs or get the node running directory? If someone starts local nodes with `assume-valid-target`, he should know what's means.

My fault, I thought the `ckb` is the running node but it should be undefined when the user uses his own node.

Keith-CY (approved):

PR #2357, opened 2022-05-01
refactor: refine multisig window

Keith-CY:
Before
image

After
image

### DIFF
--- a/packages/neuron-wallet/src/controllers/app/menu.ts
+++ b/packages/neuron-wallet/src/controllers/app/menu.ts
@@ -317,6 +317,8 @@ const updateApplicationMenu = (mainWindow: BrowserWindow | null) => {
             t(`messageBox.multisig-address.title`),
             {
               width: 900,
+              maxWidth: 900,
+              minWidth: 900,
               resizable: true
             },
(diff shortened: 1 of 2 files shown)


packages/neuron-wallet/src/controllers/app/menu.ts
```diff
@@ -317,6 +317,8 @@ const updateApplicationMenu = (mainWindow: BrowserWindow | null) => {
             t(`messageBox.multisig-address.title`),
             {
               width: 900,
+              maxWidth: 900,
+              minWidth: 900,
               resizable: true
```
yanguoyu, line 322: Why not set `resizable` false?

yanguoyu (approved):

packages/neuron-wallet/src/controllers/app/menu.ts
Keith-CY: > Why not set `resizable` false?

With this set, the window is allowed to resize vertically to add more `input fields` but not allowed to resize horizontally.

yanguoyu (approved):

PR #2360, opened 2022-05-01
fix: fix api wrapper

Keith-CY:
error.message is serialized in some cases so the error.message.code
cannot be read directly. This pr try to deserialize the message to
get error code from RPC

### DIFF
--- a/packages/neuron-wallet/src/controllers/api.ts
+++ b/packages/neuron-wallet/src/controllers/api.ts
@@ -292,7 +292,7 @@ export default class ApiController {
       return this.#walletsController.generateSendingAllTx(params)
     })
 
-    handle('generate-multisig-tx', async (_, params: { items: { address: string, capacity: string }[], multisigConfig: MultisigConfigModel}) => {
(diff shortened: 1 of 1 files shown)


packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -620,8 +620,19 @@ export default class ApiController {
           err.code = NODE_DISCONNECTED_CODE
         }
 
-        if (err.message?.code !== undefined && !Number.isNaN(err.message.code)) {
-          err.code = err.message.code
+        try {
+          /**
+           * error.message from ckb node is a stringified error object with code and message
+           */
+          const e = JSON.parse(err.message)
```
yanguoyu, line 627: Is the rpc result `object`?

Keith-CY: Yes, `error.message` consists of `{ code: number, message: string }` but is stringified.

yanguoyu (approved):

PR #2408, opened 2022-05-31
feat: add address format toggle in receive tabs

Keith-CY:
Add a button next to the address in receive tabs to toggle
address foramts between full version and deprecated version.

The switch is designed to be neglectable because the deprecated
format is discouraged.

Ref: https://github.com/Magickbase/neuron-public-issues/issues/22

https://user-images.githubusercontent.com/7271329/171185344-eef47104-d05a-4270-a1b1-e12717430de5.mov

### DIFF
--- a/packages/neuron-ui/src/components/SUDTReceive/index.tsx
+++ b/packages/neuron-ui/src/components/SUDTReceive/index.tsx
@@ -1,7 +1,9 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { useLocation } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
+import { addressToScript, bech32Address, AddressPrefix } from '@nervosnetwork/ckb-sdk-utils'
 import SUDTAvatar from 'widgets/SUDTAvatar'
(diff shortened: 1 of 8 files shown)


packages/neuron-ui/src/components/SUDTReceive/index.tsx
```diff
@@ -13,18 +15,33 @@ import { RoutePath, CONSTANTS } from 'utils'
 import styles from './sUDTReceive.module.scss'
 
 const { DEFAULT_SUDT_FIELDS } = CONSTANTS
+const toShortAddr = (addr: string) => {
```
yanguoyu, line 18: This function looks almost the same as `toShortAddr` at `Receive/index/tsx`

Keith-CY: I left them in component modules respectively because they are simple and will be removed eventually. Without sharing them between components makes sure no codes will be left on removing.

qiweiii (approved):

yanguoyu (approved):

yanguoyu (approved):

PR #2410, opened 2022-06-01
fix: style of release note

Keith-CY:
This commit fixes overflow of release note.

Before:
image

After:
image

### DIFF
--- a/packages/neuron-ui/src/components/GeneralSetting/style.module.scss
+++ b/packages/neuron-ui/src/components/GeneralSetting/style.module.scss
@@ -100,6 +100,8 @@ $action-button-width: 11.25rem;
 
   .releaseNotesStyle {
     grid-area: release-note;
+    width: 100%;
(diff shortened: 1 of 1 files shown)


qiweiii (approved):

yanguoyu (approved):

Cedar67:
After clicking the Check for Updates button, there is an error message. But not every time this error occurs, it is accidental. The trigger condition for this error is not immediately clear.
image

Keith-CY:
> After clicking the Check for Updates button, there is an error message. But not every time this error occurs, it is accidental. The trigger condition for this error is not immediately clear. image

This error is caused by a network error, the request of checking new versions doesn't finished.

This error is not caused by updates in this PR

Cedar67:
> 

Ok, I agree.

Cedar67 (approved):
Verified.
image

PR #2430, opened 2022-06-23
Fix duplicate address

yanguoyu:


packages/neuron-wallet/src/utils/queue.ts
```diff
@@ -0,0 +1,49 @@
...
+        })
+    } catch (error) {
+      callback(error)
+    }
+  }
+  const q = queue<T, R, E>(queueFn, concurrency)
+  return Object.defineProperty(q, 'asyncPush', {
+    value: (item: T) => {
+      if (ignoreSameItem && itemList.length) {
+        if (item === itemList[itemList.length - 1] && q.length()) {
```
Keith-CY, line 27: Is shallow comparison enough in our cases?

yanguoyu: Yeah, it's used to avoid reset sync task many times in a short time like debounce. Most of time, we don't need to ignore the same item.

Keith-CY (approved):

Keith-CY:
Any update? @Cedar67 @FrederLu

Cedar67:
> Any update? @Cedar67 @FrederLu

Verification in progress. Full verification is expected tomorrow. Submitted an issue to deal with previous duplicate addresses, it has been [fixed](https://github.com/nervosnetwork/neuron/pull/2438,) for verification.

PR #2432, opened 2022-06-27
chore: add issue template for synchronization issue

Keith-CY:
Add a user-friendly issue template for synchronization issues
image


yanguoyu:
Does the user know which partial log should be uploaded?

Keith-CY:
> Does the user know which partial log should be uploaded? Can the edit box upload file?

`bundled-ckb.log`, `main.log`, `status.json` are all named the same as exported files so it should be clear enough to pair files with text fields.

GitHub doesn't support a field to upload a file.

yanguoyu (approved):

PR #2434, opened 2022-06-30
fix: Adapt ledger nano s plus.

yanguoyu:
1. If the device is adapt

https://user-images.githubusercontent.com/12881040/176607494-d5755935-c518-48d3-9d78-108faa3b41d2.mov

2. If a new device does not adapt in the future, maybe you can try select other option if the APIs do not change

https://user-images.githubusercontent.com/12881040/176607570-75ed1941-8355-46c1-a205-24bcf5cd1541.mov

### DIFF
--- a/packages/neuron-ui/src/components/ImportHardware/detect-device.tsx
+++ b/packages/neuron-ui/src/components/ImportHardware/detect-device.tsx
@@ -44,7 +44,7 @@ const Info = (
   )
 }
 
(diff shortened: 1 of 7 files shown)


qiweiii (approved):

Keith-CY (approved):

Keith-CY:
Just verify if other ledgers work because we just have on ledger nano s plus. Let's verify ledger nano s plus after release

Cedar67:
> Just verify if other ledgers work because we just have on ledger nano s plus. Let's verify ledger nano s plus after release @Cedar67 @FrederLu

got it.

PR #2441, opened 2022-07-06
feat: Verify short and long address for hard wallet.

yanguoyu:
https://user-images.githubusercontent.com/12881040/177452063-af4eca2f-b4ae-41ea-83b0-f4c6d9df7b8b.mov

### DIFF
--- a/packages/neuron-ui/src/components/VerifyHardwareAddress/index.tsx
+++ b/packages/neuron-ui/src/components/VerifyHardwareAddress/index.tsx
@@ -17,16 +17,18 @@ import {
 import { ErrorCode, errorFormatter, isSuccessResponse, useDidMount } from 'utils'
 import { CkbAppNotFoundException, DeviceNotFoundException } from 'exceptions'
 import CopyZone from 'widgets/CopyZone'
+import { ADDRESS_LENGTH } from 'utils/const'
 import styles from './verifyHardwareAddress.module.scss'
 import VerifyError from './verify-error'
 
 export interface VerifyHardwareAddressProps {
   address: string
+  shortAddress: string
   wallet: State.WalletIdentity
   onDismiss: () => void
 }
 
-const VerifyHardwareAddress = ({ address, wallet, onDismiss }: VerifyHardwareAddressProps) => {
+const VerifyHardwareAddress = ({ shortAddress, address, wallet, onDismiss }: VerifyHardwareAddressProps) => {
   const [t] = useTranslation()
   const dialogRef = useRef<HTMLDialogElement | null>(null)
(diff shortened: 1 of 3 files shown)


packages/neuron-ui/src/components/VerifyHardwareAddress/index.tsx
```diff
@@ -176,9 +178,17 @@ const VerifyHardwareAddress = ({ address, wallet, onDismiss }: VerifyHardwareAdd
...
                   name={t('hardware-verify-address.actions.copy-address')}
-                  style={{ lineHeight: '1.625rem' }}
+                  style={{ lineHeight: '1.625rem', width: '400px', display: 'flex' }}
                 >
-                  {address}
+                  {address.length > ADDRESS_LENGTH ? (
+                    <>
+                      <span className={styles.addressOverflow}>{address.slice(0, -20)}</span>
+                      <span>...</span>
+                      <span>{address.slice(-20)}</span>
```
Keith-CY, line 187: Better not to split the address at the same point
When an address consists of `part_a | part_b` and displayed as `part_a ... part_b`, it indicates that some chars are omitted between `part_a` and `part_b`, which means `part_a | part_b` is definitely different from `part_a ... part_b`

packages/neuron-ui/src/components/VerifyHardwareAddress/index.tsx
```diff
@@ -139,7 +141,7 @@ const VerifyHardwareAddress = ({ address, wallet, onDismiss }: VerifyHardwareAdd
     const res = await getDevicePublicKey()
     if (isSuccessResponse(res)) {
       const { result } = res
-      if (result?.address === address) {
+      if (result?.address === address || result?.address === shortAddress) {
```
Keith-CY, line 144: `shortAddress` could be generated in the callback to avoid an extra prop of this Component

packages/neuron-ui/src/components/VerifyHardwareAddress/index.tsx
yanguoyu: If the `address` is a long address, show it with a wrap? not show with `part_a ... part_b`?
Like this:
image
Rather than like this:
image

Keith-CY: The second one is fine, the point is not to split the address at the same point, we can split it into `address.silce(0, -21)` and `address.slice(-20)` to make sure `...` will take place of at least 1 char

Cedar67 (approved):
plan to verify the issue by draft package.

PR #2443, opened 2022-07-07
fix: Use retry times to replace while.

yanguoyu:
Avoid the promise can't resolve or reject.


Keith-CY (approved):

Cedar67 (approved):
Verified that the CKB and Indexer processes were shut down multiple times as expected.

PR #2450, opened 2022-07-14
fix: Same group cell witness should be 0x. If witness's lock set as 0x. it will be serialized that not equal 0x

yanguoyu:


yanguoyu:
If some unexpected error happens, it should display. https://github.com/nervosnetwork/neuron/pull/2450/commits/88fb8d13286e6c3086bcaf12b62afb8722a4a1c6

image

Keith-CY (approved):

Cedar67 (approved):
The scenario has been verified:
When the sender has multiple input cells, the transaction is expected.

PR #2467, opened 2022-08-12
fix: Fix ckb process kill unexpected.

yanguoyu:
1. Sometime ckb process does not emit close after calling` ckb.kill`, so wait 1 second when stop process monitor.
2. Use kill(2) to replace kill(), [kill(2)](https://man7.org/linux/man-pages/man7/signal.7.html) means `Sends a signal to a specified process, to all members of a specified process group, or to all processes on the system`.


yanguoyu:
@Cedar67 Try this, I use 'SIGKILL' for the kill's parameter. It means kill -9.
image

Cedar67 (approved):
Verified.

Keith-CY (approved):

PR #2485, opened 2022-09-13
chore: update assumed valid target

Keith-CY:
Set block 8067227 as the assumed valid target

Ref: https://explorer.nervos.org/block/8067227


yanguoyu (approved):

Keith-CY:
CI failed due to expiration of the certificate for Windows, has nothing to do with the code changes.

Cedar67 (approved):

PR #2488, opened 2022-09-18
fix: New instance for every test case to avoid side effect

yanguoyu:


yanguoyu:
I have tried to run the test case 11 times. It's all success.

Keith-CY (approved):

PR #2490, opened 2022-09-22
fix: Upgrade lerna to 5.5.2.

yanguoyu:
To fix https://github.com/nervosnetwork/neuron/security/dependabot/107, https://github.com/nervosnetwork/neuron/security/dependabot/116 and https://github.com/nervosnetwork/neuron/security/dependabot/117.
1. upgrade typescript because when running `lerna run build --scope=@nervina-labs/ckb-indexer` some typescript declare files throw exceptions.
2. After upgrading typescript to 4.8.3, fix some type errors.

### DIFF
--- a/packages/neuron-ui/src/components/OfflineSignDialog/index.tsx
+++ b/packages/neuron-ui/src/components/OfflineSignDialog/index.tsx
@@ -16,7 +16,7 @@ import {
   sendSUDTTransaction,
 } from 'states'
 import { OfflineSignJSON, signAndExportTransaction, OfflineSignType } from 'services/remote'
-import { PasswordIncorrectException, isErrorWithI18n } from 'exceptions'
+import { PasswordIncorrectException } from 'exceptions'
 import styles from '../PasswordRequest/passwordRequest.module.scss'
 
 interface SignDialogProps {
@@ -157,7 +157,7 @@ const OfflineSignDialog = ({ isBroadcast, wallet, offlineSignJSON, onDismiss }: 
           }
         }
       } catch (err) {
-        if (isErrorWithI18n(err) && err.code === ErrorCode.PasswordIncorrect) {
+        if (err instanceof PasswordIncorrectException) {
           setError(t(err.message))
         }
       }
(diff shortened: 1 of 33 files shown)


packages/neuron-ui/src/exceptions/index.ts
```diff
@@ -110,3 +110,21 @@ export class ValueReservedException extends Error {
...
   }
 }
+
+export type ErrorWithI18n = {
+  message: string
+  code: ErrorCode
+  i18n?: Record<string, string>
+}
+
+export function isErrorWithI18n(error: unknown): error is ErrorWithI18n {
```
Keith-CY, line 120: `unknown` is for outside users, `any` could be used there and return `true` if `typeof error?.message === 'string' && typeof error.code === 'number'`

packages/neuron-ui/src/components/OfflineSignDialog/index.tsx
```diff
@@ -157,7 +157,7 @@ const OfflineSignDialog = ({ isBroadcast, wallet, offlineSignJSON, onDismiss }:
           }
         }
       } catch (err) {
-        if (err.code === ErrorCode.PasswordIncorrect) {
+        if (isErrorWithI18n(err) && err.code === ErrorCode.PasswordIncorrect) {
```
Keith-CY, line 160: `err instanceof PasswordIncorrectException`

packages/neuron-ui/src/components/PasswordRequest/index.tsx
```diff
@@ -286,7 +286,7 @@ const PasswordRequest = () => {
           }
         }
       } catch (err) {
-        if (err.code === ErrorCode.PasswordIncorrect) {
+        if (isErrorWithI18n(err) && err.code === ErrorCode.PasswordIncorrect) {
```
Keith-CY, line 289: `err instanceof PasswordIncorrectException`

packages/neuron-ui/src/exceptions/index.ts
```diff
@@ -110,3 +110,19 @@ export class ValueReservedException extends Error {
...
+export type ErrorWithI18n = {
+  message: string
+  code: ErrorCode
+  i18n?: Record<string, string>
+}
+
+export function isErrorWithI18n(error: any): error is ErrorWithI18n {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
```
Keith-CY, line 123: `typeof error === 'object' && error !== null` is redundant because `error?.message` covers them.

Keith-CY (approved):

PR #2503, opened 2022-10-12
Bump some deps

Keith-CY:


Keith-CY:
Among these updates, only `electron` and `sqlite3` may have impacts on functions.

I think we can merge these first and have a simple test to see if the app runs and if data are stored correctly. @yanguoyu @JeffreyMa597 @Cedar67

yanguoyu (approved):

yanguoyu:
> Among these updates, only `electron` and `sqlite3` may have impacts on functions, they are so fundamental that it's easy to test.
> 
> I think we can merge these first and have a simple test to see if the app runs and if data are stored correctly. @yanguoyu @JeffreyMa597 @Cedar67

sure.

Keith-CY:
I think this PR could be merged first and rely on the e2e tests to check if the app runs and if data are stored correctly, can we? @Cedar67

Cedar67 (approved):
Basic verification passed.

Cedar67:
> I think this PR could be merged first and rely on the e2e tests to check if the app runs and if data are stored correctly, can we? @Cedar67

OK, It has been verified.

PR #2506, opened 2022-10-17
feat: Update storybook to 6.5.12

yanguoyu:
1. Add `react-transition-group` because `primereact` depend on it. We only use `Calendar` in `primereact` to develop `DatetimePicker`, I think we can remove `primereact` later, it's a too-big package. We don't need add `react-transition-group` before because `@storybook/addon-knobs` depend on it, and now we remove `@storybook/addon-knobs`.
2. Add `webpack` devDependencies because `storybook` need explicitly specify `webpack`.


packages/neuron-ui/.storybook/.babelrc
```diff
@@ -0,0 +1,6 @@
+{
+  "presets": [
+    "@babel/preset-env",
+    "@babel/preset-typescript"
+  ]
+}
```
Keith-CY, line 6: Newline at the end of file is required

Keith-CY (approved):

Cedar67:
> > @yanguoyu Does this pr affect functionality or user experience? If necessary, pls list specific functional impact points.
> 
> It will not affect functions, if `Neuron` can be normally packaged and run, it will be ok.

got it.

Cedar67:
@yanguoyu  The style of the button here is not affected by this change?

image

yanguoyu:
> @yanguoyu The style of the button here is not affected by this change?
> 
> image

Of course, it was caused by `new wallet` with new UI.

Cedar67 (approved):
Verified.

PR #2507, opened 2022-10-18
refactor: use Array.prototype.forEach instead of Array.prototype.map

Keith-CY:
Ref: https://github.com/Magickbase/neuron-public-issues/issues/67


yanguoyu (approved):

Keith-CY:
Please have a review @JeffreyMa597

JeffreyMa597:
It's ok for me.
And it seems I didn't have an access to approve this issue.

PR #2510, opened 2022-10-21
fix:Apple silicon support.

JeffreyMa597:
As the description of the title. This PR is to support packaging for Apple chips.

Ref:https://github.com/Magickbase/neuron-public-issues/issues/62

### DIFF
--- a/packages/neuron-wallet/electron-builder.yml
+++ b/packages/neuron-wallet/electron-builder.yml
@@ -11,14 +11,14 @@ directories:
 afterSign: scripts/notarize.js
 
 files:
-  - from: "../.."
-    to: "."
-    filter: ["!**/*", ".ckb-version", "ormconfig.json"]
+  - from: '../..'
+    to: '.'
+    filter: ['!**/*', '.ckb-version', 'ormconfig.json']
   - package.json
   - dist
-  - ".env"
-  - "!**/*.map"
-  - "!**/src/"
+  - '.env'
+  - '!**/*.map'
+  - '!**/src/'
 
 publish:
   - github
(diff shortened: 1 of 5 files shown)


Keith-CY (approved):

packages/neuron-wallet/electron-builder.yml
```diff
@@ -62,8 +62,11 @@ mac:
   entitlementsInherit: assets/entitlements.plist
   electronLanguages: ["en", "zh", "zh_CN", "zh_TW"]
   target:
-    - dmg
-    - zip
+    - target: default
```
yanguoyu, line 65: Should we retain `dmg` and `zip`? @Keith-CY

Keith-CY: `target: default` will package a `zip` and a `dmg`

Ref: https://www.electron.build/configuration/mac.html

JeffreyMa597: @yanguoyu It seems this value should be a `String`. 
image

Keith-CY: > @yanguoyu It seems this value should be a `String`. image

Yes, it's a string here

```yml
target:
  - target: default
    arch:
      - arm64
      - x64
```

equals to

```json
{
  target: [{
    target: 'default',
    arch: ['arm64', 'x64']
  }]
}
```

yanguoyu: I have tried with `default` to package, it will output zip and dmg.

yanguoyu (approved):

JeffreyMa597:
Cause the Neuron didn't accept commits without verifying, I'll create a new branch to submit these changes based on the latest `develop` branch. Thank you.

PR #2515, opened 2022-10-27
Fix:apple silicon support #2510

JeffreyMa597:
------
apple silicon support
Ref: https://github.com/Magickbase/neuron-public-issues/issues/62

Check List
------

### Test
default
### Task
none

### DIFF
--- a/.github/workflows/package_for_test.yml
+++ b/.github/workflows/package_for_test.yml
@@ -98,14 +98,28 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Mac
-          path: release/Neuron-*-mac.zip
+          path: release/Neuron-*-x64-mac.zip
+
+      - name: Upload Neuron App Zip(arm64)
+        if: matrix.os == 'macos-latest'
+        uses: actions/upload-artifact@v2
+        with:
+          name: Neuron-Mac
+          path: release/Neuron-*-arm64-mac.zip
 
       - name: Upload Neuron Dmg
         if: matrix.os == 'macos-latest'
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Dmg
-          path: release/Neuron-*.dmg
+          path: release/Neuron-*-x64-mac.dmg
+
+      - name: Upload Neuron Dmg(arm64)
+        if: matrix.os == 'macos-latest'
+        uses: actions/upload-artifact@v2
+        with:
+          name: Neuron-Dmg
+          path: release/Neuron-*-arm64-mac.dmg
 
       - name: Upload Neuron Win
         if: matrix.os == 'windows-2019'
--- a/.github/workflows/package_for_test.yml
+++ b/.github/workflows/package_for_test.yml
@@ -98,28 +98,28 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Mac
-          path: release/Neuron-*-x64-mac.zip
+          path: release/Neuron-*-mac-x64.zip
 
       - name: Upload Neuron App Zip(arm64)
         if: matrix.os == 'macos-latest'
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Mac
-          path: release/Neuron-*-arm64-mac.zip
+          path: release/Neuron-*-mac-arm64.zip
 
       - name: Upload Neuron Dmg
         if: matrix.os == 'macos-latest'
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Dmg
-          path: release/Neuron-*-x64-mac.dmg
+          path: release/Neuron-*-x64.dmg
 
       - name: Upload Neuron Dmg(arm64)
         if: matrix.os == 'macos-latest'
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Dmg
-          path: release/Neuron-*-arm64-mac.dmg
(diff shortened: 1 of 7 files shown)


yanguoyu (approved):

packages/neuron-wallet/electron-builder.yml
```diff
@@ -48,7 +48,7 @@ win:
         - x64
 
 mac:
-  artifactName: "${productName}-v${version}-${os}.${ext}"
+  artifactName: "${productName}-v${version}-${arch}-${os}.${ext}"
```
Keith-CY, line 51: I would prefer `os` prior to `arch` because `arch` here is a subset of `os`

Keith-CY:
The task(https://github.com/nervosnetwork/neuron/blob/develop/scripts/release-checksums.rb) to verify checksums of released packages should be updated because there are more packages to generate and package names are changed

JeffreyMa597:
> The task(https://github.com/nervosnetwork/neuron/blob/develop/scripts/release-checksums.rb) to verify checksums of released packages should be updated because there are more packages to generate and package names are changed
 
@Keith-CY  Have Fixed it. Please take a look.

.github/workflows/package_for_test.yml
```diff
@@ -98,14 +98,28 @@ jobs:
...
+          name: Neuron-Mac
+          path: release/Neuron-*-arm64-mac.zip
 
       - name: Upload Neuron Dmg
         if: matrix.os == 'macos-latest'
         uses: actions/upload-artifact@v2
         with:
           name: Neuron-Dmg
-          path: release/Neuron-*.dmg
+          path: release/Neuron-*-x64-mac.dmg
```
yanguoyu, line 115: The dmg output seems not to end with os?

.github/workflows/package_for_test.yml
```diff
@@ -98,14 +98,28 @@ jobs:
...
           name: Neuron-Mac
-          path: release/Neuron-*-mac.zip
+          path: release/Neuron-*-x64-mac.zip
+
+      - name: Upload Neuron App Zip(arm64)
+        if: matrix.os == 'macos-latest'
+        uses: actions/upload-artifact@v2
+        with:
+          name: Neuron-Mac
+          path: release/Neuron-*-arm64-mac.zip
```
Keith-CY, line 108: `path` should be updated as https://github.com/nervosnetwork/neuron/pull/2515/files#diff-170ebc8e4dc40acf23cbe0ecce5f3e2aef1652511f59860db704106b197e1d52R113

.github/workflows/package_for_test.yml
```diff
@@ -98,14 +98,28 @@ jobs:
...
           name: Neuron-Dmg
-          path: release/Neuron-*.dmg
+          path: release/Neuron-*-x64-mac.dmg
+
+      - name: Upload Neuron Dmg(arm64)
+        if: matrix.os == 'macos-latest'
+        uses: actions/upload-artifact@v2
+        with:
+          name: Neuron-Dmg
+          path: release/Neuron-*-arm64-mac.dmg
```
Keith-CY, line 122: `path` should be updated

scripts/release-checksums.rb
```diff
@@ -25,8 +30,10 @@ def get_sha256_checksum(url)
 OS | Arch | Package | SHA256 Checksum
 -- | -- | -- | --
 Windows | x64 | [exe](#{windows_exe}) | <code>#{windows_exe_sha256}</code>
-macOS | x64 | [zip](#{macos_zip}) | <code>#{macos_zip_sha256}</code>
-macOS | x64 | [DMG](#{macos_dmg}) | <code>#{macos_dmg_sha256}</code>
+macOS | x64 | [zip](#{macos_x64_zip}) | <code>#{macos_x64_zip_sha256}</code>
+macOS | arm64 | [zip](#{macos_arm64_zip}) | <code>#{macos_arm64_zip_sha256}</code>
+macOS | x64 | [DMG](#{macos_x64_dmg}) | <code>#{macos_x64_dmg_sha256}</code>
+macOS | arm64 | [DMG](#{macos_arm64_dmg}) | <code>#{macos_arm64_dmg_sha256}</code>
 Linux | x64 |  [AppImage](#{linux_appimage}) | <code>#{linux_appimage_sha256}</code>
```
Keith-CY, line 37: It's my fault that I missed mentioning the counterpart in CI https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/check_checksums.yml#L22, since there're 2 more records, the truncation should cover 2 more lines of the release note.

.github/workflows/package_for_test.yml
JeffreyMa597: Fixed.

scripts/release-checksums.rb
Keith-CY: Sorry for adding more details late.

The code
```
          python3 -c "import sys, json; print(\"\n\".join(json.load(sys.stdin)['body'].splitlines()[-9:]))" |\
```
in https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/check_checksums.yml#L22 will truncate last 9 lines of a release page, which is
image

The code will be executed after a release to check if the checksums listed in the release note are consistent with the attached packages.

Now we're going to have 2 more attached packages, we will add 2 more checksums in the release note, so the code should truncate last 11 lines instead of 9 lines.

Keith-CY (approved):

PR #2530, opened 2022-11-19
fix: Use batch rpc to request result. Optimization sql search.

yanguoyu:
1. For the nerovs dao page, there are many rpc requests, use batch request.
2. Because some users have many relations with others cells, left join twice to search inputs and outputs.
3. Find unlock transactions with the status that equal Dead or Pending.

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAO/hooks.ts
+++ b/packages/neuron-ui/src/components/NervosDAO/hooks.ts
@@ -451,9 +451,7 @@ export const useUpdateWithdrawList = ({
       .then(txs => {
         const committedTx = txs.filter(v => v.txStatus.status === 'committed')
         committedTx.forEach((tx, idx) => {
-          if (tx.txStatus.status === 'committed') {
-            txMap.set(depositOutPointHashes[idx], tx)
-          }
+          txMap.set(depositOutPointHashes[idx], tx)
         })
         const blockHashes = [
           ...(committedTx.map(v => v.txStatus.blockHash).filter(v => !!v) as string[]),
@@ -468,9 +466,7 @@ export const useUpdateWithdrawList = ({
           .exec()
           .then(blockHeaders => {
             blockHeaders.forEach((header, idx) => {
-              if (header?.dao) {
-                hashHeaderMap.set(blockHashes[idx], header.dao)
-              }
+              hashHeaderMap.set(blockHashes[idx], header.dao)
             })
             const withdrawList = new Map()
             records.forEach(record => {
--- a/packages/neuron-ui/src/components/NervosDAO/hooks.ts
+++ b/packages/neuron-ui/src/components/NervosDAO/hooks.ts
@@ -20,8 +20,9 @@ import {
   generateDaoDepositTx,
   generateDaoClaimTx,
 } from 'services/remote'
-import { ckbCore, getHeaderByNumber, calculateDaoMaximumWithdraw } from 'services/chain'
+import { ckbCore, getHeaderByNumber } from 'services/chain'
 import { isErrorWithI18n } from 'exceptions'
+import { calculateMaximumWithdraw } from '@nervosnetwork/ckb-sdk-utils'
 
 const {
   MIN_AMOUNT,
@@ -91,7 +92,7 @@ export const useInitData = ({
     updateNervosDaoData({ walletID: wallet.id })(dispatch)
     const intervalId = setInterval(() => {
       updateNervosDaoData({ walletID: wallet.id })(dispatch)
-    }, 3000)
+    }, 10000)
     updateDepositValue(
       `${
         BigInt(wallet.balance) > BigInt(CKBToShannonFormatter(`${MIN_DEPOSIT_AMOUNT}`))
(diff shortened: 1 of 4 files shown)


packages/neuron-ui/src/components/NervosDAO/hooks.ts
```diff
@@ -436,35 +437,79 @@ export const useUpdateWithdrawList = ({
...
+    }
+    const depositOutPointHashes = records.map(v => v.depositOutPoint?.txHash ?? v.outPoint.txHash)
+    const txMap = new Map<CKBComponents.Hash, CKBComponents.TransactionWithStatus>()
+    ckbCore.rpc
+      .createBatchRequest<'getTransaction', string[], CKBComponents.TransactionWithStatus[]>(
+        depositOutPointHashes.map(v => ['getTransaction', v])
+      )
+      .exec()
+      .then(txs => {
+        const committedTx = txs.filter(v => v.txStatus.status === 'committed')
```
Keith-CY, line 452: Transactions should not be filtered because `txMap.set(depositOutPointHashes[idx], tx)` is related to the order, the order would be incorrect if some txs are filtered out.

Or maybe `txMap.set(tx.hash, tx)` is enough here. If so, the `if(tx.txStatus.status === 'committed')` is redundant because `committedTx.forEach` has been filtered before.

packages/neuron-ui/src/components/NervosDAO/hooks.ts
```diff
@@ -436,35 +437,79 @@ export const useUpdateWithdrawList = ({
...
+        const hashHeaderMap = new Map<CKBComponents.Hash, string>()
+        return ckbCore.rpc
+          .createBatchRequest<'getHeader', string[], CKBComponents.BlockHeader[]>(
+            blockHashes.map(v => ['getHeader', v])
+          )
+          .exec()
+          .then(blockHeaders => {
+            blockHeaders.forEach((header, idx) => {
+              if (header?.dao) {
+                hashHeaderMap.set(blockHashes[idx], header.dao)
```
Keith-CY, line 472: maybe `hashHeaderMap.set(header.hash, header.dao)` is more clear

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -484,16 +484,22 @@ export class TransactionsService {
...
       .orderBy({
         'input.id': 'ASC'
       })
       .getOne()
+    const txwithOutput = await getConnection()
+      .getRepository(TransactionEntity)
+      .createQueryBuilder('transaction')
+      .where('transaction.hash is :hash', { hash })
+      .leftJoinAndSelect('transaction.outputs', 'output')
+      .getOne()
```
Keith-CY, line 496: Why not query `outputs` directly

packages/neuron-wallet/src/services/cells.ts
```diff
@@ -131,6 +131,65 @@ export default class CellsService {
     return uniqueLockArgs
   }
 
+  private static async addUnLockInfo(cells: Cell[]): Promise<Cell[]> {
```
Keith-CY, line 134: `UnLock` should be `Unlock`

Keith-CY: This refactor should be done along with tests

packages/neuron-ui/src/components/NervosDAO/hooks.ts
yanguoyu: I refer to this [function](https://github.com/nervosnetwork/ckb-sdk-js/blob/674dfa575571a835911410b1d6a12f8cf5999170/packages/ckb-sdk-core/src/index.ts#L423). If the transaction is not committed, it will throw an exception and we will set it null.
And of course `txMap.set(tx.hash, tx)` is enough here. I will fix it.

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -484,16 +484,23 @@ export class TransactionsService {
       .createQueryBuilder('transaction')
       .where('transaction.hash is :hash', { hash })
       .leftJoinAndSelect('transaction.inputs', 'input')
-      .leftJoinAndSelect('transaction.outputs', 'output')
       .orderBy({
         'input.id': 'ASC'
       })
       .getOne()
+    const txOutpus = await getConnection()
```
Keith-CY, line 491:
```suggestion
    const txOutputs = await getConnection()
```

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -484,16 +484,23 @@ export class TransactionsService {
...
+      .where({
+        outPointTxHash: hash
+      })
+      .getMany()
 
     if (!tx) {
       return undefined
     }
 
+    tx.outputs = txOutpus
```
Keith-CY, line 503:
```suggestion
    tx.outputs = txOutputs
```

Keith-CY (approved):

packages/neuron-ui/src/components/NervosDAO/hooks.ts
Keith-CY: Seems not to be updated but is marked resolved, is it incorrect to use `header.hash` directly?

Cedar67 (approved):
Due to the changes to the ckb architecture, this PR needs to be validated after being merged into the `develop` branch.

yanguoyu:
> Due to the changes to the ckb architecture, this PR needs to be validated after being merged into the `develop` branch.

I have rebased develop.

packages/neuron-ui/src/components/NervosDAO/hooks.ts
```diff
@@ -436,35 +437,75 @@ export const useUpdateWithdrawList = ({
...
+    const txMap = new Map<CKBComponents.Hash, CKBComponents.TransactionWithStatus>()
+    ckbCore.rpc
+      .createBatchRequest<'getTransaction', string[], CKBComponents.TransactionWithStatus[]>(
+        depositOutPointHashes.map(v => ['getTransaction', v])
+      )
+      .exec()
+      .then(txs => {
+        const committedTx = txs.filter(v => v.txStatus.status === 'committed')
+        committedTx.forEach((tx, idx) => {
+          txMap.set(depositOutPointHashes[idx], tx)
```
Keith-CY, line 454: `depositOutPointHashes[idx]` and `tx` may be mismatched because `committedTx` is a subset of `depositOutPointHashes`(filtered by `'status === 'committed'`)

It should be `txMap.set(tx.transaction.hash, tx)`

packages/neuron-ui/src/components/NervosDAO/hooks.ts
```diff
@@ -436,35 +437,75 @@ export const useUpdateWithdrawList = ({
...
+        })
+        const blockHashes = [
+          ...(committedTx.map(v => v.txStatus.blockHash).filter(v => !!v) as string[]),
+          ...(records.map(v => (v.depositOutPoint ? v.blockHash : null)).filter(v => !!v) as string[]),
+          tipBlockHash,
+        ]
+        const hashHeaderMap = new Map<CKBComponents.Hash, string>()
```
Keith-CY, line 461: The definition of `hashHeaderMap` could be moved into `then` of the `batchRequest` because it's not used outside.

JeffreyMa597 (approved):

packages/neuron-ui/src/components/NervosDAO/hooks.ts
yanguoyu: I fixed it by use `txs` to loop. And moving it nearby when it was used. https://github.com/nervosnetwork/neuron/pull/2530/commits/95d3bbe5a9d6bcaa507fa6fe5f651b9ddb06285b

Cedar67:
Verified on `develop` branch.

PR #2543, opened 2022-12-13
Chore: call rimraf with npx

JeffreyMa597:
------
As title.

This error(https://github.com/Magickbase/neuron/actions/runs/3627132547/jobs/6116742113#step:8:54) has been found several times in CI.
we used `npx` to replace installing `rimraf` to avoid `not recognized as an internal or external command`.

Ref: https://github.com/Magickbase/neuron-public-issues/issues/86

Check List
------

### Test
e2e Test
### Task
none

### DIFF
--- a/packages/ckb-indexer/package.json
+++ b/packages/ckb-indexer/package.json
@@ -26,7 +26,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "clean": "rimraf lib",
+    "clean": "npx rimraf lib",
     "prepublishOnly": "yarn run clean && yarn run build",
     "release": "npm publish"
   },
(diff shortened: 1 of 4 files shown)


Keith-CY:
This PR simply removes `rimraf` from dependencies but dependencies are updated in the `yarn.lock`, is it as expected?

yanguoyu (approved):

JeffreyMa597:
> This PR simply removes `rimraf` from dependencies but dependencies are updated in the `yarn.lock`, is it as expected?
 
@Keith-CY Yeah, actually I've also noticed this before, but this dependency is all from other dependency packages. So, if we just try to solve the error we got, I think it is acceptable. am I right?
 
image

JeffreyMa597:
After removing this dependency I reinstalled all of the dependencies to check our project was normal. So there are some other dependencies that may be upgraded.
image

Keith-CY (approved):

PR #2574, opened 2023-01-18
fix: Sometimes RPC timeout but ckb is living

yanguoyu:
1. Try to check ckb is alive every 30s.
2. When startCkb try to check the ckb is null.


Cedar67 (approved):
Verified. No reproduction in five times.

Keith-CY (approved):

Keith-CY:
I didn't find the reason for point 1 `Try to check ckb is alive every 30s.` in the related issue, please add a brief explanation.

yanguoyu:
> I didn't find the reason for point 1 `Try to check ckb is alive every 30s.` in the related issue, please add a brief explanation.

I think 10s is too faster, It's unnecessary.

PR #2587, opened 2023-02-13
Feat/implement dynamic fee rate based on fee rate statistics

JeffreyMa597:
------
As title.

An RPC to fetch fee_rate_statistics is added since CKB@v0.106.0, which returns mean fee rate and median fee rate of the specified block range.

With the historic data, Neuron could offer more precise fee rate recommendations for users.

PRD: https://github.com/Magickbase/neuron-public-issues/issues/87
UI: https://www.figma.com/file/6XNoimRDbFTTNm016rbIdU/Magickbase?node-id=319%3A4539&t=oQrWTpg2GhW4obRJ-0

Ref: https://github.com/Magickbase/neuron-public-issues/issues/95

Check List
------

### Test
e2e Test
### Task
none


Keith-CY:
Most commits in this PR are not signed

JeffreyMa597:
> Most commits in this PR are not signed

Yes, it is. I'll create another PR to fix this problem. It maybe cause I use `git push --force` to push the code.

JeffreyMa597:
@Keith-CY  The new PR is over here. https://github.com/nervosnetwork/neuron/pull/2599

PR #2590, opened 2023-02-20
feat: Adapt ckb light node

yanguoyu:
**How to switch to light client**
1. ./script/download-ckb.sh
2. start `Neuron`
3. Click `Network`, choose `Light Testnet`

**Light node sync way**
Create a `sync_progress` table to save sync progress.
1. Call `set_script` to sync scripts by light node
3. Every 5s to check if there exists a queue or else create some sync queue
4. In a sync queue there will check wheater exists txs between `blockStart` and `blockEnd`, if exists, it will send tx hashes to the `queue` class, or else update the `sync_progress`
5. Update the `sync_progress` after `queue` class using the tx hashes.

**Something needs to do before ready**
~~1. Add test cases for added code. I will add test cases later.~~

**Something needs to do later**
1. Support change light node data path.


packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -47,9 +50,12 @@ export default class Queue {
   start = async () => {
     logger.info("Queue:\tstart")
     try {
-      this.#indexerConnector = new IndexerConnector(this.#addresses, this.#url, this.#indexerUrl)
-
-      await this.#indexerConnector.connect()
+      if (this.#url === 'http://localhost:9000') {
```
Keith-CY, line 53: The url could be replaced by https://github.com/nervosnetwork/neuron/pull/2590/files#diff-486b3b6eb77724b4fb3b13a4b64f038277a6a98d78a62d12959ef22a3dd5bbb6R4

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -62,36 +68,36 @@ export default class Queue {
...
     })
 
     this.#checkAndSaveQueue.error((err: any, task: any) => {
       logger.error(err, JSON.stringify(task, undefined, 2))
     })
 
     this.#indexerConnector.transactionsSubject
-      .subscribe(transactions => {
-        const task = { transactions: transactions.map(t => t.transaction) }
+      .subscribe(({ txHashes, params }) => {
```
Keith-CY, line 92: Unnecessary destructure

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -100,7 +106,36 @@ export default class Queue {
...
+  private async fetchTxsWithStatus(txHashes: string[]) {
+    const rpc = generateRPC(this.#url)
+    const txsWithStatus = await rpc.createBatchRequest<'getTransaction', string[], CKBComponents.TransactionWithStatus[]>(
+      txHashes.map(v => ['getTransaction', v])
+    ).exec()
+    const txs: Transaction[] = []
+    const blockHashes = []
+    for (let index = 0; index < txsWithStatus.length; index++) {
+      if (txsWithStatus[index]?.transaction) {
+        const tx = Transaction.fromSDK(txsWithStatus[index]?.transaction)
```
Keith-CY, line 118:
```suggestion
        const tx = Transaction.fromSDK(txsWithStatus[index].transaction)
```

packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts
```diff
@@ -100,7 +106,36 @@ export default class Queue {
...
+        tx.blockHash = txsWithStatus[index].txStatus.blockHash!
+        blockHashes.push(tx.blockHash)
+        txs.push(tx)
+      } else if ((txsWithStatus[index].txStatus as any) === 'rejected') {
+        logger.warn(`Transaction[${txHashes[index]}] was rejected`)
+        throw new Error(`failed to fetch transaction for hash ${txHashes[index]}`)
+      }
```
Keith-CY, line 125: Better to handle the `else` branch

packages/neuron-wallet/src/services/live-cell-service.ts
```diff
@@ -1,7 +1,8 @@
 import Script from 'models/chain/script'
 import LiveCell from 'models/chain/live-cell'
 import { queryIndexer } from 'block-sync-renderer/index'
-import { LumosCellQuery, LumosCell } from 'block-sync-renderer/sync/indexer-connector'
+import { LumosCell } from 'block-sync-renderer/sync/connector'
+import { LumosCellQuery } from 'block-sync-renderer/sync/connector'
```
Keith-CY, line 5: Two modules could be imported by one line

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -23,9 +24,21 @@ const presetNetworks: { selected: string; networks: Network[] } = {
...
 }
 
+const lightClientNetwork: Network[] = [
+  {
+    id: 'testnet_light',
+    name: 'Light Testnet',
+    remote: BUNDLED_LIGHT_CKB_URL,
+    genesisHash: TESTNET_GENESIS_HASH,
+    type: NetworkType.Light,
+    chain: LIGHT_CHAIN
```
Keith-CY, line 34: The name could be more exact, like `Light Client Testnet`

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -23,9 +24,21 @@ const presetNetworks: { selected: string; networks: Network[] } = {
...
+  }
+]
+
 enum NetworksKey {
   List = 'networks',
-  Current = 'selected'
+  Current = 'selected',
+  AddedLight = 'AddedLight'
```
Keith-CY, line 41: The label naming convention is inconsistent

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -0,0 +1,225 @@
...
+      blockStartNumber: result.lastCursor === '0x' ? parseInt(blockRange[1]) : parseInt(blockRange[0]),
+      blockEndNumber: parseInt(blockRange[1]),
+      cursor: result.lastCursor === '0x' ? undefined : result.lastCursor
+    })
+  }
+
+  private async collectLiveCellsByScript(query: LumosCellQuery) {
+    const { lock, type, data } = query
+    if (!lock && !type) {
+      throw new Error('at least one parameter is required')
```
Keith-CY, line 159: This error message is a bit obscure, how about `at least one script is required`

packages/neuron-wallet/src/database/chain/entities/sync-progress.ts
```diff
@@ -0,0 +1,48 @@
...
+  blockStartNumber: number = 0
+
+  @Column()
+  blockEndNumber: number = 0
+
+  @Column({ type: 'varchar' })
+  cursor?: HexString
+
+  @Column({ type: 'boolean' })
+  softDelete: boolean = false
```
Keith-CY, line 35: Not sure if `deleted` is better, I think any `delete` field in the table is `soft`

packages/neuron-wallet/src/services/settings.ts
```diff
@@ -54,6 +62,9 @@ export default class SettingsService extends Store {
     if (!this.ckbDataPath) {
       this.ckbDataPath = path.resolve(app.getPath('userData'), 'chains/mainnet')
     }
+    if (!this.lightDataPath) {
+      this.lightDataPath = path.resolve(app.getPath('userData'), 'chains/light')
```
Keith-CY, line 66: Neuron only supports mainnet full node, so the path could be `chains/mainnet`. Now light client will be supported for mainnet and testnet, maybe `chains/light/testnet` is more clear in this case

packages/neuron-wallet/src/types/controller.d.ts
```diff
@@ -91,7 +91,7 @@ declare module Controller {
 
     name: string
     remote: string
-    type: 0 | 1 // 0 for the default type, 1 for the normal type
+    type: 0 | 1 | 2 // 0 for the default type, 1 for the normal type
```
Keith-CY, line 94: The comment should be updated too.

packages/neuron-wallet/src/services/settings.ts
yanguoyu: Good idea, and I think we need to rename `lightDataPath ` to `testnetLightDataPath` because I think we will add a feature to set it as editable. So we can edit `testnetLightDataPath` and `mainnetLightDataPath` separately.

packages/neuron-wallet/src/services/light-runner.ts
```diff
@@ -0,0 +1,148 @@
...
+  protected networkType: NetworkType = NetworkType.Full
+
+  static getInstance(): CKBLightRunner {
+    if (!CKBLightRunner.instance) {
+      CKBLightRunner.instance = new CKBLightRunner()
+    }
+    return CKBLightRunner.instance as CKBLightRunner
+  }
+
+  protected get ckbBinary() {
```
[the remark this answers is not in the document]
yanguoyu: I think it's better to rename it to `binary` because I will refactor `packages/neuron-wallet/src/services/ckb-runner.ts` to extend `Runner`. How about `binary`?

packages/neuron-ui/src/components/NetworkSetting/index.tsx
```diff
@@ -71,11 +71,12 @@ const NetworkSetting = ({ chain = chainState, settings: { networks = [] } }: Sta
...
-                  <button type="button" data-action="edit" aria-label={t('common.edit')} title={t('common.edit')}>
-                    <EditNetwork />
-                  </button>
-                  {isDefault ? null : (
+                  {network.chain === 'light_testnet' || (
+                    <button type="button" data-action="edit" aria-label={t('common.edit')} title={t('common.edit')}>
+                      <EditNetwork />
+                    </button>
+                  )}
+                  {isDefault || network.chain === 'light_testnet' ? null : (
```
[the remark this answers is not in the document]
Keith-CY: > `isDefault || network.chain === 'light_testnet'` is a boolean, there will not renderer others.

Sure, but it's prone to be rendered with unexpected element along with developing for a long time

packages/neuron-wallet/src/services/settings.ts
```diff
@@ -42,6 +42,14 @@ export default class SettingsService extends Store {
     this.writeSync('ckbDataPath', dataPath)
   }
 
+  get testnetLightDataPath() {
+    return this.readSync('testnetLightDataPath')
```
Keith-CY, line 46: I would suggest using a constant as the key so a typo won't break it

packages/neuron-wallet/src/services/sync-progress.ts
```diff
@@ -0,0 +1,81 @@
...
+      .insert()
+      .orIgnore()
+      .values(params.map(v => SyncProgress.fromObject(v)))
+      .execute()
+    const walletIds = [...new Set(params.map(v => v.walletId))]
+    await getConnection()
+      .createQueryBuilder()
+      .update(SyncProgress)
+      .set({ delete: true })
+      .where({ walletId: Not(In(walletIds)) })
```
Keith-CY, line 22: I didn't get this one, the method looks to reset progress of scripts belonging to specific wallets, but the wallets other than the specified ones are deleted?

packages/neuron-wallet/src/services/sync-progress.ts
```diff
@@ -0,0 +1,81 @@
...
+      .where({ blake160s: In(blake160s) })
+      .execute()
+  }
+
+  static async getSyncStatus(script: CKBComponents.Script) {
+    const scriptHash = scriptToHash(script)
+    const res = await getConnection()
+      .getRepository(SyncProgress)
+      .createQueryBuilder()
+      .where({ softDelete: false, hash: scriptHash })
```
Keith-CY, line 40: The `softDelete` field was renamed to `delete`

packages/neuron-wallet/src/services/sync-progress.ts
```diff
@@ -0,0 +1,81 @@
...
+      .getOne()
+    return res
+  }
+
+  static async getAllSyncStatusToMap() {
+    const result: Map<CKBComponents.Hash, SyncProgress> = new Map()
+    const syncProgresses = await getConnection()
+      .getRepository(SyncProgress)
+      .createQueryBuilder()
+      .where({ softDelete: false })
```
Keith-CY, line 50: The `softDelete` field was renamed to `delete`

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -0,0 +1,225 @@
...
+    super()
+    this.indexer = new CkbIndexer(nodeUrl, nodeUrl)
+    this.lightRpc = new LightRPC(nodeUrl)
+    this.addressMetas = addresses.map(address => AddressMeta.fromObject(address))
+
+    this.indexerQueryQueue = queue(this.collectLiveCellsByScript)
+  }
+
+  private async synchronize() {
+    if (this.syncQueue.idle()) {
```
Keith-CY, line 47: It's better to return early

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -0,0 +1,225 @@
...
+      })
+    }
+  }
+
+  private async subscribeSync() {
+    const minSyncBlockNumber = await SyncProgressService.getMinBlockNumber()
+    const header = await this.lightRpc.getTipHeader()
+    this.blockTipsSubject.next({
+      cacheTipNumber: minSyncBlockNumber,
+      indexerTipNumber: parseInt(header.number, 16)
```
Keith-CY, line 90: Why not use `+header.number` to transform an integer from hex format to number format

`parseInt` accepts unrecognizable char, as `parseInt('0xfzz', 16)` will be `15`

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -0,0 +1,225 @@
...
+    const result = await this.lightRpc.getTransactions({ script, blockRange, scriptType }, 'asc', '0x64', cursor!)
+    if (!result.txs.length) {
+      await SyncProgressService.updateSyncStatus(syncProgress.hash, {
+        blockStartNumber: parseInt(blockRange[1]),
+        blockEndNumber: parseInt(blockRange[1]),
+        cursor: undefined
+      })
+      return
+    }
+    const scriptHash = scriptToHash(script)
```
Keith-CY, line 147: `scriptHash` can be fetched from `syncProgress.hash`

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -0,0 +1,225 @@
...
+    this.transactionsSubject.next({ txHashes: result.txs.map(v => v.txHash), params: scriptHash })
+    this.syncInQueue.set(scriptToHash(script), {
```
Keith-CY, line 149: Redundant computation of script hash

packages/neuron-wallet/src/services/light-runner.ts
```diff
@@ -0,0 +1,148 @@
...
+}
+
+abstract class NodeRunner {
+  protected constructor() {}
+  protected abstract networkType: NetworkType
+  protected runnerProcess: ChildProcess | undefined
+  protected static instance: NodeRunner | undefined
+  static getInstance(): NodeRunner {
+    throw new Error('should be called by instance')
+  }
```
Keith-CY, line 23: The error message is confusing, maybe refer to `concrete class`?

packages/neuron-wallet/src/services/light-runner.ts
```diff
@@ -0,0 +1,148 @@
...
+        this.runnerProcess = undefined
+      } else {
+        resolve()
+      }
+    })
+  }
+}
+
+export class CKBLightRunner extends NodeRunner {
+  protected networkType: NetworkType = NetworkType.Full
```
Keith-CY, line 56: Is the network type `full` while class is of `light runner`?

packages/neuron-wallet/src/services/light-runner.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  protected get binary() {
+    const appPath = app.isPackaged
+      ? path.join(path.dirname(app.getAppPath()), '..', './bin')
+      : path.join(__dirname, '../../bin')
+    const binary = app.isPackaged
+      ? path.resolve(appPath, './ckb-light-client')
+      : path.resolve(appPath, `./${this.platform()}`, './ckb-light-client')
+    return this.platform() === 'win' ? binary + '.exe' : binary
+  }
```
Keith-CY, line 73: There should be a similar logic to download `ckb binary`, maybe they can be encapsulated

packages/neuron-wallet/src/services/light-runner.ts
```diff
@@ -0,0 +1,148 @@
...
+    })
+    if (!fs.existsSync(SettingsService.getInstance().testnetLightDataPath)) {
+      fs.mkdirSync(SettingsService.getInstance().testnetLightDataPath, { recursive: true })
+    }
+    fs.writeFileSync(this.configFile, newValues.join('\n'))
+  }
+
+  async start() {
+    if (this.runnerProcess) {
+      logger.info(`CKBLightRunner:\tckb is not closed, close it before start...`)
```
Keith-CY, line 121: `CKBLightRunner` could be `CKB Light Runner` for readability, and the message refer to `ckb`, maybe a typo?

packages/neuron-wallet/src/services/sync-progress.ts
yanguoyu: Yeah, the wallets that are not in sync will be set to delete true. But it may be better to split these into two functions.
And the two functions will be called when init sync.

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
yanguoyu:
```suggestion
    if (!this.syncQueue.idle()) return
```
Is that so?

Keith-CY: LGTM

Keith-CY (approved):

Keith-CY:
CKB light client v0.2.2 has been released and is suggested to be used

Ref: https://github.com/nervosnetwork/ckb-light-client/releases/tag/v0.2.2

PR #2592, opened 2023-02-23
Refactored Calendar component to remove dependency of primereact. (#2592

mortoys:
Hi everyone,

I've reimplemented the new Calendar component as described below issue, and removed the dependency on primereact.

It generally follows the previous interactive logic, keeps the current system's colour and style, and adds a check for the available range of dates.

Any suggestions for improvement would be appreciated.

https://github.com/Magickbase/neuron-public-issues/issues/66

Kapture 2023-02-23 at 14 56 09


packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+                  data-type="button"
+                  aria-label={date.label}
+                  aria-pressed={dateEqual(date.instance, value)}
+                  title={date.label}
+                  className={`
+                    ${styles.calDateItem}
+                    ${date.isToday ? 'today' : ''}
+                    ${dateEqual(date.instance, value) ? 'active' : ''}
+                  `}
+                  disabled={!date.curMonth || !date.selectable}
```
yanguoyu, line 79: I think it's better to set `disabled` as a callback function. Currently, It can only control by `minDate` and `maxDate`.
```javascript
disabled={!date.curMonth || disabledTime(date.instance) }
```

packages/neuron-ui/src/widgets/Calendar/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  return true
+}
+
+export function monthInRange(year: number, month: number, range: DateRange) {
+  return dayInRange(new Date(year, month + 1, 1), range)
+}
+
+export function yearInRange(year: number, range: DateRange) {
+  return dayInRange(new Date(year + 1, 1, 1), range)
```
yanguoyu, line 41: I think we can judge by `getFullYear`. Otherwise: `yearInRange(2022, { minDate: new Date(2023, 0, 1), maxDate: new Date(2023, 0, 31) } )` will be false.
And I don't know why we should increase the year with 1

packages/neuron-ui/src/widgets/Calendar/utils.ts
```diff
@@ -0,0 +1,115 @@
+import { useTranslation } from 'react-i18next'
```
yanguoyu, line 1: Utils functions should add test case.

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  if (prevProps.minDate?.toDateString() !== nextProps.minDate?.toDateString()) {
+    return false
+  }
+  if (prevProps.maxDate?.toDateString() !== nextProps.maxDate?.toDateString()) {
+    return false
+  }
+  if (prevProps.onChange !== nextProps.onChange) {
+    return false
+  }
+  return true
```
yanguoyu, line 185:
```suggestion
  return prevProps.value?.toDateString() === nextProps.value?.toDateString()
  && prevProps.minDate?.toDateString() === nextProps.minDate?.toDateString()
  && prevProps.maxDate?.toDateString() === nextProps.maxDate?.toDateString()
  && prevProps.onChange === nextProps.onChange
```

On the other hand, I think we can let whoever uses this component keep the reference equal.

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  )
+
+  const onChangeMonth = (monthOptionItem: Option) => {
+    setMonth(monthOptionItem.value)
+    setStatus('date')
+  }
+  const onChangeYear = (yearOptionItem: Option) => {
+    setYear(yearOptionItem.value)
+    setStatus('month')
+  }
```
yanguoyu, line 160: use `useCallback` to cache functions.

packages/neuron-ui/src/widgets/Calendar/utils.ts
mortoys: really a bug, my fault

packages/neuron-ui/src/widgets/Calendar/index.tsx
mortoys: From my understanding of useCallback, it is not necessary inside the component, only useful when passing values to child components.

yanguoyu: ```<Selector options={yearOptions} onChange={onChangeYear} />```
Does the `onChangeYear` pass to `Selector` component?
And `useCallback` can memory function with dependence, avoid to create function repeatedly when the function component execute

mortoys: You're right, I missed it.

packages/neuron-ui/src/widgets/Calendar/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+    firstDayOfWeek: 0,
+    dayNames: dayNames.map(dayname => t(`datetime.${dayname}.full`)),
+    dayNamesShort: dayNames.map(dayname => t(`datetime.${dayname}.short`)),
+    dayNamesMin: dayNames.map(dayname => t(`datetime.${dayname}.tag`)),
+    monthNames: monthNames.map(monname => t(`datetime.${monname}.short`)),
+    monthNamesShort: monthNames.map(monname => t(`datetime.${monname}.short`)),
+  }
+
+  return locale
+}
```
Keith-CY, line 113: Local names can be exported by 

E.g.
```
const date = new Date(Date.UTC(2020, 11, 20, 3, 23, 16, 738));
console.log(new Intl.DateTimeFormat('zh-CN', { weekday: 'short', month: 'short', calendar: 'chinese' }).format(date));
```

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  </ol>
+)
+
+export interface CalendarProps {
+  value: Date | undefined
+  onChange: (value: Date) => void
+  minDate?: Date
+  maxDate?: Date
+}
+const Calendar: React.FC<CalendarProps> = ({ value, onChange, minDate = null, maxDate = null }) => {
```
Keith-CY, line 35: `minDate` and `maxDate` are declared as optional Date, but `null` is assigned as the default value

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const [month, setMonth] = useState(new Date().getMonth() + 1)
+  const [status, setStatus] = useState<'year' | 'month' | 'date'>('date')
+
+  useEffect(() => {
+    setYear(value?.getFullYear() ?? new Date().getFullYear())
+    setMonth((value?.getMonth() ?? new Date().getMonth()) + 1)
+  }, [value])
+
+  const locale = useLocalNames()
+  const weeknames = useMemo(() => [...Array(7).keys()].map(_ => locale.dayNamesMin[(_ + locale.firstDayOfWeek) % 7]), [
```
Keith-CY, line 46: First, don't instantiate an array in the format `Array(length)`, I cannot find a reference, but it's not a good practice.

Second, `[...Array(7).keys()]` could be replaced by `Array.from({ length: 7 })`

Last, `_` implies the variable is not used, `i` or `index` should be used in this case.

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+    locale,
+  ])
+  const monthname = locale.monthNames[month - 1]
```
Keith-CY, line 49: `monthname` should be `monthName`

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const calendar = useMemo(() => getMonthCalendar(year, month), [year, month])
+  function disabledTime(date: Date) {
```
Keith-CY, line 52: Use method name `is~` to imply it's a condition checking

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+              <td key={`${date.month}${date.date}`}>
+                <button
+                  type="button"
+                  data-type="button"
+                  aria-label={date.label}
+                  aria-pressed={dateEqual(date.instance, value)}
+                  title={date.label}
+                  className={`
+                    ${styles.calDateItem}
+                    ${date.isToday ? 'today' : ''}
```
Keith-CY, line 79: `isToday` could be marked by `aria-*` too, as `aria-pressed`, besides, `aria` has its standard now, `attr-*` would be better for a custom attribute`

Ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    ${dateEqual(date.instance, value) ? 'active' : ''}
```
Keith-CY, line 80: The style of `active` could be added by `aria-pressed` above

packages/neuron-ui/src/widgets/Calendar/index.tsx
mortoys: add `aria-current` property

packages/neuron-ui/src/widgets/Calendar/utils.ts
mortoys: This function is used to localized the date. The one I have is to get the localized names of the month and week. Not quite the same

packages/neuron-ui/src/widgets/Calendar/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (a === undefined || b === undefined) {
+    return false
+  }
+  return a?.toDateString() === b?.toDateString()
+}
+
+/**
+ * @description Generate monthly calendar 2D table data
+ */
+export function getMonthCalendar(year: number, month: number): Day[][] {
```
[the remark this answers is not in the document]
mortoys: The primereact uses table, and I also think it might be better semantically to use table as opposed to line wrap layout?

packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
```diff
@@ -155,14 +128,7 @@ const DatetimePicker = ({
...
         )}
-        <Calendar
-          value={selected}
-          minDate={new Date()}
-          onChange={onCalendarChange}
-          inline
-          locale="es"
-          className={styles.calendar}
-        />
+        <Calendar value={selected} minDate={tomorrow} onChange={onCalendarChange} />
```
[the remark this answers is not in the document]
mortoys: I specifically designed the `minDate` to include the day of the `minDate`, otherwise it would feel pretty odd. So using new Date() would include today

yanguoyu (approved):

Keith-CY:
CI failed

packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
```diff
@@ -155,14 +129,7 @@ const DatetimePicker = ({
...
         )}
-        <Calendar
-          value={selected}
-          minDate={new Date()}
-          onChange={onCalendarChange}
-          inline
-          locale="es"
-          className={styles.calendar}
-        />
+        <Calendar value={selected} minDate={tomorrow} onChange={onCalendarChange} lang={i18n.language} />
```
Keith-CY, line 132: The `language` could be fetched by `useTranslation` instead of a global instance.

Besides, the `language` could be fetched inside `Calendar`

Ref: https://github.com/nervosnetwork/neuron/blob/1f19cca93b3247447ab6082eca16021dd6656db6/packages/neuron-ui/src/containers/Settings/index.tsx#L43

Keith-CY (approved):

Keith-CY:
@Cedar67 please have a review on the functionality

Keith-CY:
How is it going @Cedar67

Cedar67:
> How is it going @Cedar67

Two other high-priority issues are currently under validation. This PR will be arranged to complete the verification next Monday.

Cedar67:
@mortoys The display style for the current day and the selected date should remain the same as before.

image

Cedar67:
Need to add a period at the end of each month abbreviation.

image

Cedar67:
One exception: When selecting another month by the arrow symbol, it automatically jumps back to the current month.


https://user-images.githubusercontent.com/64718893/223000131-f6e9bb52-50ce-41d2-a75a-f7c990d1c899.mov

mortoys:
> Need to add a period at the end of each month abbreviation.
> 
> image

According to the current i18n method, this cannot be controlled

JeffreyMa597 (approved):

Cedar67:
> One exception: When selecting another month by the arrow symbol, it automatically jumps back to the current month.

Verified.

Cedar67:
> @mortoys The display style for the current day and the selected date should remain the same as before.
> 
> image

Verified.

Keith-CY:
> > Need to add a period at the end of each month abbreviation.
> > image
> 
> According to the current i18n method, this cannot be controlled

It can be done by updating here https://github.com/nervosnetwork/neuron/pull/2592/files#diff-227821703ffb93fc0998891374bc9394fb37fb7bfb38486fd284b9b0c5aaef69R98-R101

```javascript
const getLocalMonthNames = (lang: string) => {
  const formater = new Intl.DateTimeFormat(lang, { month: 'short' })
  return Array.from({ length: 12 }, (_, i) => `${formater.format(new Date(Date.UTC(2023, i, 1)))}${lang.startsWith('en') ? '.' : ''}`)
}
```

mortoys:
I found some calendar accessibility examples here：https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/
https://react-spectrum.adobe.com/react-aria/useCalendar.html

and a great wai-aria lesson: https://www.udacity.com/course/web-accessibility--ud891
I plan to upgrade the component more accessibility

mortoys:
Add focus style
Add wai-aria labels and related i18n language
Add focus fully controlled with the keyboard: 
1. focus can be controlled by `Tab` and `Shift + Tab` to move forward and backward
2. `Space` and `Enter` to select
3. `Up Arrow` `Down Arrow` `Right Arrow` `Left Arrow` control the direction of movement
4. `Home` first day of the current month. `End` last day of the current month.
5. `Page Up` previous month `Page Down` next month

It can be tested use Mac Voiceover(press Command-F5), and feel whether the blind can be used properly
Also you can check accessibility tree in Chrome DevTools.

It's still a lot to do, e.g.
- [ ] The modal must trap the focus

Kapture 2023-03-13 at 17 54 25

Cedar67:
Code scanning and Unit test results indicate exceptions that need to be confirmed.

packages/neuron-ui/src/widgets/Calendar/focusControl.tsx
```diff
@@ -0,0 +1,162 @@
+import React, { useState, useRef, useEffect, KeyboardEvent } from 'react'
+import { isMonthInRange, isDayInRange } from './utils'
+
+type ButtonHasFocusProps = React.ButtonHTMLAttributes<HTMLButtonElement> & { isFocus: boolean }
+export const ButtonHasFocus = ({ isFocus, children, ...props }: ButtonHasFocusProps) => {
+  const ref = useRef<HTMLButtonElement | null>(null)
+  useEffect(() => {
+    if (isFocus && ref.current) {
+      ref.current.focus()
+    }
```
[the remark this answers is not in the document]
Keith-CY: The `else` branch should also be declared even if it's ignored intentionally, as

```javascript
} else {
  // ignore
}
```

So it won't be missed in future development, and reviewers can understand it clearly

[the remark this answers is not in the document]
mortoys: All right

Keith-CY:
CI failed at https://github.com/nervosnetwork/neuron/actions/runs/4403595487/jobs/7713534899#step:9:478

packages/neuron-ui/src/widgets/Calendar/focusControl.tsx
```diff
@@ -0,0 +1,162 @@
...
+  title: string
+  label: string
+  selectable: boolean
+}
+export const useSelectorFocusControl = (value: number, options: Option[], onChange: (option: Option) => void) => {
+  const [focusIndex, setFocusIndex] = useState(-1)
+
+  useEffect(() => {
+    setFocusIndex(options.findIndex(option => option.value === value))
+  }, [value])
```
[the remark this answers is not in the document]
mortoys: Do add dependency of a function of this type `setFocusIndex` is necessary? it never change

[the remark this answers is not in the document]
yanguoyu: That's right. It's unnecessary.

Keith-CY (approved):

Cedar67:
Cannot be saved after the year is changed.


https://user-images.githubusercontent.com/64718893/226126765-6a65020e-192e-481a-874c-09dbc1b0e253.mov

Cedar67:
> Need to add a period at the end of each month abbreviation.
> 
> image

This issue has been Solved and Verified.

Keith-CY:
> > Need to add a period at the end of each month abbreviation.
> > image
> 
> This issue has been Solved and Verified.

Please mark this PR approved if tests are all green

Cedar67:
> Please mark this PR approved if tests are all green

Found a new issue to be confirmed by the developer.

https://github.com/nervosnetwork/neuron/pull/2592#issuecomment-1474938203

mortoys:
@Cedar67 
Changes will only be submitted after the date is selected in the monthly calendar
this interaction is same as the past

Keith-CY:
> @Cedar67 Changes will only be submitted after the date is selected in the monthly calendar this interaction is same as the past

Appreciate your contribution!

PR #2595, opened 2023-02-28
fix: Restart or stop sync process by connectionStatusSubject.

yanguoyu:
This pr will fix https://github.com/Magickbase/neuron-public-issues/issues/96. Here is why I changed the code like this:
When starting the sync process in `ckb-runner`, the network `genesisHash` has not refresh, so it will start a wrong sync process that connects the `mainnet` database and use the `testnet` data, so I remove the start and stop from `ckb-runner`. But after I remove the stop sync process from `ckb-runner`, `NetworksController` may delay stop the sync process because of `debounceTime`. We should stop the sync process right now when the `ckb` node is stopped otherwise the sync process may use the wrong net data. So I remove `debounceTime` from `NetworksController` otherwise the sync process may delay stopping.

### Here are the pr changes
1. Remove `debounceTime` because it has `distinctUntilChanged` unless the node connects or disconnects frequently.
3. Remove starting or stopping the sync process from `ckb-runner`, because it has control by `connectionStatusSubject`, when the `ckb` connection status changes, it will stop or start.
4. Add start and stop App logs to find the latest running log.

### DIFF
--- a/packages/neuron-wallet/src/controllers/networks/index.ts
+++ b/packages/neuron-wallet/src/controllers/networks/index.ts
@@ -1,13 +1,13 @@
 import { dialog } from 'electron'
 import { t } from 'i18next'
 import env from 'env'
-import { skip, distinctUntilChanged, debounceTime } from 'rxjs/operators'
+import { distinctUntilChanged, skip } from 'rxjs/operators'
 import { NetworkType, Network } from 'models/network'
 import NetworksService from 'services/networks'
 import NodeService from 'services/node'
 import { ResponseCode } from 'utils/const'
 import { IsRequired, InvalidName, NetworkNotFound, CurrentNetworkNotSet } from 'exceptions'
-import { switchToNetwork } from 'block-sync-renderer'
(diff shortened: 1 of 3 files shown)


Keith-CY (approved):

JeffreyMa597 (approved):

Cedar67:
https://github.com/Magickbase/neuron-public-issues/issues/96#issuecomment-1465327577

It can be reproduced very occasionally.

image

Log:
[neuron_debug_1678663081451.zip](https://github.com/Magickbase/neuron-public-issues/files/10952353/neuron_debug_1678663081451.zip)

yanguoyu:
From the image your pr number is 2592, but my pr number is 2595.

Cedar67:
> From the image, your pr number is 2592, but my pr number is 2595.

@yanguoyu Sorry, I read it wrong branch number.
Verified and No exceptions have been found in the target branch validation. https://github.com/nervosnetwork/neuron/pull/2595

Cedar67 (approved):

PR #2599, opened 2023-03-04
Feat/implement dynamic fee rate based on fee rate statistics

JeffreyMa597:
------
As title.

An RPC to fetch fee_rate_statistics is added since CKB@v0.106.0, which returns mean fee rate and median fee rate of the specified block range.

With the historic data, Neuron could offer more precise fee rate recommendations for users.

PRD: https://github.com/Magickbase/neuron-public-issues/issues/87
UI: https://www.figma.com/file/6XNoimRDbFTTNm016rbIdU/Magickbase?node-id=319%3A4539&t=oQrWTpg2GhW4obRJ-0

Ref: https://github.com/Magickbase/neuron-public-issues/issues/95

Check List
------
1、The `customer price` and `dropdown price` can be switched.

image

2、The fee rate dropdown list is displayed with 3 options, and the `suggested_fee_rate` will be updated every `30 seconds`.
    1. Fast: fee rate is `2 * suggested_fee_rate - 1000`
    2. Standard: fee rate is `suggested_fee_rate`
    3. Slow: fee rate is `1000`

image

3、The fee in every single option will update when the fee rate is updated. 
Of course, It's calculated by `account` and `fee rate`.
The standard one is selected by default and also when the selected value is not matched with any option.

image

4、The `suggest_fee_rate` will update every 30 seconds, also the fee is. The fee is based on the `fee rate` which we type in the text field. When the `fee rate` is set below 1000, there will be a warning tip instead of the suggest tip.

image
image

5、The fee in `Nervos Dao`, `Customized assets` and `Assets account` are all based on the `suggest fee rate` instead of using the `medium value`.

image
image

### Test
e2e Test
### Task
none


yanguoyu:
Texts should be internationalization, and they should not use directly.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  // render selected value in options
+  const parseValue = (value: string, options: OptionProps[]) => {
+    let option
+
+    for (let i = 0, num = options.length; i < num; i++) {
+      if (typeof options[i].value !== 'undefined' && options[i].value === value) {
+        option = options[i]
+      }
+    }
```
yanguoyu, line 46:
```suggestion
    let option = options.find(v => v.value !== 'undefined' && v.value === value)
```

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+    return (
+      option || {
+        label: typeof placeholder === 'undefined' ? DEFAULT_PLACEHOLDER_STRING : placeholder,
```
yanguoyu, line 50:
```suggestion
        label: placeholder ?? DEFAULT_PLACEHOLDER_STRING,
```

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+        value: '',
+      }
+    )
+  }
+
+  const [selectedValue, setSelectedValue] = useState<OptionProps>(
+    parseValue(valueFromProps, optionsFromProps) || {
```
yanguoyu, line 57: `parseValue` will not return undefined, it may be no need to add `||`

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,200 @@
+import React, { ReactNode, useEffect, useMemo, useRef, useState } from 'react'
```
yanguoyu, line 1: Can we use `packages/neuron-ui/src/widgets/Select/index.tsx` or `packages/neuron-ui/src/widgets/Dropdown/index.tsx`?

packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -264,6 +266,51 @@ const clear = (dispatch: StateDispatch) => {
...
+          return res.result
+        }
+        if (res.status === 0 || res.status === 114) {
+          throw new Error(res.message.content)
+        }
+        return res
+      })
+      .catch((err: Error) => {
+        // eslint-disable-next-line no-console
+        console.log(err, 'err')
```
yanguoyu, line 291: `console` should not be used in the release.

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+            <RingProgressBar percents={percent} color="#00C891" strokeWidth="3px" size="16px" backgroundColor="#CCC" />
+            <span>{`单价将在${countDown}秒后刷新`}</span>
+          </div>
+        ) : null}
+      </div>
+
+      {isStandard ? (
+        <div className={styles['dropdown-box']}>
+          <DropdownWithCustomRender
+            onChange={e => handleDropdownChange(e)}
```
yanguoyu, line 152:
```suggestion
            onChange={handleDropdownChange}
```

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    app: {
+      send = appState.send,
+      feeRateStatics: { suggestFeeRate = 0 },
+      countDown,
+    },
+    wallet: { id: walletID = '' },
+  } = useGlobalState()
+
+  const isStandard = type === PriceTypeEnum.Standard
+  const label = isStandard ? '单价' : '自定义价格'
```
Keith-CY, line 44: Internationalization is required

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+            <span>慢速</span>
+            <span className={styles['label-comment']}>{`单价${feeValuesMap?.[0]?.feeRateValue ||
+              priceArray[0]}shannons/byte | 费用${feeValuesMap?.[0]?.value || 0}CKB`}</span>
+          </div>
+        ),
+      },
+      {
+        value: priceArray[1],
+        label: (
+          <div className={styles['label-wrap']}>
```
Keith-CY, line 87: Set class name in `camelCase` style

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+import styles from './pricePanel.module.scss'
+
+interface PricePanelProps {
+  field: string
+  price: string
+  onPriceChange: React.EventHandler<React.SyntheticEvent>
+}
+enum PriceTypeEnum {
+  Customer = 'customer',
```
Keith-CY, line 21: `Customer` or `Custom`

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+  Standard = 'standard',
+}
+const DEFAULT_PRICE_ARRAY = ['1000', '2000', '3000']
+
+const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, onPriceChange }: PricePanelProps) => {
+  const [t] = useTranslation()
+  const [type, setType] = useState<PriceTypeEnum>(PriceTypeEnum.Standard)
+  const [inputError, setInputError] = useState<string | null>(null)
+  const [inputHint, setInputHint] = useState<string | null>(null)
```
Keith-CY, line 30: These two values can be computed on the fly without storing them in a state

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  const isStandard = type === PriceTypeEnum.Standard
+  const label = isStandard ? '单价' : '自定义价格'
+  const percent = (countDown / DEFAULT_COUNT_DOWN) * 100
+
+  const handleDropdownChange = useCallback((e: any) => {
+    const { value: dropdownChangedValue } = e
+    setDropdowValue(dropdownChangedValue)
+    onPriceChange?.({ target: { value: dropdownChangedValue } } as any)
+  }, [])
```
Keith-CY, line 51: Deps are missing

packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -264,6 +266,51 @@ const clear = (dispatch: StateDispatch) => {
 
 const useClear = (dispatch: StateDispatch) => useCallback(() => clear(dispatch), [dispatch])
 
+export const useGetbatchGeneratedTx = async ({
```
Keith-CY, line 269: `useGetbatchGeneratedTx` should be `useGetBatchGeneratedTx`

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+}
+const DEFAULT_PRICE_ARRAY = ['1000', '2000', '3000']
+
+const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, onPriceChange }: PricePanelProps) => {
+  const [t] = useTranslation()
+  const [type, setType] = useState<PriceTypeEnum>(PriceTypeEnum.Standard)
+  const [inputError, setInputError] = useState<string | null>(null)
+  const [inputHint, setInputHint] = useState<string | null>(null)
+  const [dropdownValue, setDropdowValue] = useState(DEFAULT_PRICE_ARRAY[1])
+  const [feeMap, setFeeMap] = useState([])
```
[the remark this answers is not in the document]
JeffreyMa597: > Why `Map` set an Array?

I'll change the name.

packages/neuron-ui/src/components/PricePanel/index.tsx
JeffreyMa597: ok.

packages/neuron-ui/src/components/PricePanel/index.tsx
JeffreyMa597: > Set class name in `camelCase` style

ok. I'll change all the class name.

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    })
+  }, [send.outputs, priceArray])
+
+  useEffect(() => {
+    if (suggestFeeRate === 0) {
+      setPriceArray(['1000', '2000', '3000'])
+    } else {
+      setPriceArray(['1000', `${suggestFeeRate}`, `${Number(suggestFeeRate) * 2 - 1000}`])
+    }
+    setInputHint(`建议价格 ${suggestFeeRate} shannons/byte`)
```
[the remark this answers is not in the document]
JeffreyMa597: > I think the `inputHint` can directly show by `suggestFeeRate`.

Make sense.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
JeffreyMa597: > Can we use `packages/neuron-ui/src/widgets/Select/index.tsx` or `packages/neuron-ui/src/widgets/Dropdown/index.tsx`?

Just to be different from the original dropdown, because this is a new component and the original dropdown may be used elsewhere.

I can change the name of this componnet.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
JeffreyMa597: already remove `| |`

packages/neuron-ui/src/components/PricePanel/index.tsx
JeffreyMa597: good idea.

packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -394,6 +394,10 @@ export default class ApiController {
       return this.#daoController.withdrawFromDao(params)
     });
 
+    handle('get_fee_rate_statics', async () => {
```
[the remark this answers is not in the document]
JeffreyMa597: Named `get_fee_rate_statis` instead.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
JeffreyMa597: It seems both the `DropDown` and `Select` are in use in some other components, so I'll still call it by this name.

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const feeValuesMap = feeMap?.map((item: any) => ({
+    feeRateValue: item.feeRateValue,
+    value: shannonToCKBFormatter(item.feeValue),
+  }))
+
+  useEffect(() => {
+    if (!priceArray.includes(dropdownValue)) {
+      handleDropdownChange({ value: priceArray[1] })
+    }
+  }, [priceArray])
```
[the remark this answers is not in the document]
JeffreyMa597: Acctually, it will be updated when the suggestFeeRate is updated.

packages/neuron-ui/src/containers/Navbar/index.tsx
```diff
@@ -147,6 +155,46 @@ const Navbar = () => {
     bestBlockNumber >= 0 ? localNumberFormatter(bestBlockNumber) : '-'
   }`
 
+  useEffect(() => {
+    intervalHandle.current = setTimeout(() => {
```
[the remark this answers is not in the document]
JeffreyMa597: I'll optimize the logic to clear the Timeout before setting a new one, to avoid memory leaks.  
Using `setTimeout` may be better at this moment, when using setInterval, the execution time of the timer is fixed, and the next callback function will be called after the set time interval, regardless of whether the previous callback function has been completed. This may cause the execution time of the timer to be inaccurate due to competition with other codes.

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -828,4 +829,11 @@ export default class TransactionSender {
       privateKey: `0x${masterKeychain.derivePath(path).privateKey.toString('hex')}`
     }))
   }
+
+  public getFeeRateStatics = async (): Promise<any> => {
+    const network = await NetworksService.getInstance().getCurrent()
+    const res = await rpcRequest(network.remote, { method: 'get_fee_rate_statics', params: [] })
+
+    return res
```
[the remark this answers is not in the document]
JeffreyMa597: Have been removed.

packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -394,6 +394,10 @@ export default class ApiController {
...
+      return this.#daoController.getFeeRateStatics()
```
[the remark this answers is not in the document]
JeffreyMa597: Have changed the request way.

packages/neuron-ui/src/components/PricePanel/index.tsx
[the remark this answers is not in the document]
yanguoyu: It's strange that when the `suggestFeeRate` changes, it will change the selected option.
I think if the user selects `fast`, it will also be `fast`. 
Or, to put it another way, the `suggestFeeRate` only affects the display, but should not affect the selected option.
And in current logic, it may exist one exception, after the user selects the fast, and the current fee is 3000, but later, the fast fee become 2000, it will change to the slow option, it's strange, the user selects fast, but it becomes slow.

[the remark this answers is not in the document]
JeffreyMa597: Ok, have fixed the FeeRateMode which is currently selected.

packages/neuron-ui/src/containers/Navbar/index.tsx
```diff
@@ -23,6 +22,7 @@ import {
 } from 'utils'
 
 import { Migrate } from 'services/subjects'
+
```
Keith-CY, line 25: No update in this file

packages/neuron-ui/src/locales/en.json
```diff
@@ -991,6 +991,22 @@
...
         "view-raw-data": "View raw data"
       }
+    },
+    "dropdown": {
+      "placeholder": "Please select..."
+    },
+    "price-switch": {
+      "price": "Price",
+      "customPrice": "Custom Price",
+      "errorTip": "Price must not be less than {{minPrice}} shannons/byte, please re-enter",
```
Keith-CY, line 1001:
```suggestion
      "errorTip": "Price cannot be less than {{minPrice}} shannons/byte, please re-enter",
```

packages/neuron-ui/src/locales/en.json
```diff
@@ -991,6 +991,22 @@
...
+      "hintTip": "Suggested price is {{suggestFeeRate}} shannons/byte",
```
Keith-CY, line 1002: `shannons/kB` is used widely, we'd better use it too

image

image

packages/neuron-ui/src/services/remote/remoteApiWrapper.ts
```diff
@@ -80,6 +80,7 @@ type Action =
   | 'generate-dao-deposit-all-tx'
   | 'start-withdraw-from-dao'
   | 'withdraw-from-dao'
+  | 'get_fee_rate_statis'
```
Keith-CY, line 83: `get_fee_rate_statis` should be `get_fee_rate_stats`

packages/neuron-ui/src/services/remote/wallets.ts
```diff
@@ -28,6 +28,7 @@ export const generateDaoDepositAllTx = remoteApi<Controller.GenerateDepositAllTr
 )
 export const generateDaoWithdrawTx = remoteApi<Controller.WithdrawParams>('start-withdraw-from-dao')
 export const generateDaoClaimTx = remoteApi<Controller.ClaimParams>('withdraw-from-dao')
+export const getFeeRateStatis = remoteApi<void>('get_fee_rate_statis')
```
Keith-CY, line 31: `getFeeRateStatis` should be `getFeeRateStats`

packages/neuron-ui/src/styles/color.scss
```diff
@@ -1,4 +1,8 @@
 :root {
   --nervos-green: #3cc68a;
   --nervos-green-light: #31a572; // most recent color to nervos green that pairs with white background
+
+  --avtive-color: #00c891;
```
Keith-CY, line 5: Maybe `active-color`

packages/neuron-ui/src/styles/index.scss
```diff
@@ -5,7 +5,8 @@
 * {
   outline: none !important;
   -webkit-tap-highlight-color: none !important;
-  cursor: default !important;
+  // why we need this property with value of '!important'
```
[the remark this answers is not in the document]
Keith-CY: It was required to act as a native app, and `pointer` is in web app exclusively.

packages/neuron-ui/src/utils/calculateFee.ts
```diff
@@ -1,5 +1,5 @@
 export const calculateFee = (tx: any) => {
-  if (!tx) {
+  if (!tx?.inputs || !tx?.outputs) {
```
Keith-CY, line 2: In what case `!tx` is not enough as a condition

packages/neuron-ui/src/states/stateProvider/reducer.ts
```diff
@@ -89,6 +92,8 @@ export type StateAction =
   | { type: AppActions.Ignore; payload?: any }
   | { type: AppActions.UpdateExperimentalParams; payload: { tx: any; assetAccount?: any } | null }
   | { type: AppActions.UpdateLoadedTransaction; payload: { filePath?: string; json: OfflineSignJSON } }
+  | { type: AppActions.GetFeeRateStatis; payload: any }
```
Keith-CY, line 95: Can we type the payload

packages/neuron-ui/src/components/SpecialAssetList/index.tsx
```diff
@@ -3,6 +3,7 @@ import { useTranslation } from 'react-i18next'
 import { useHistory, useLocation } from 'react-router-dom'
 import { Pagination } from '@uifabric/experiments'
 import SpecialAsset, { AssetInfo } from 'components/SpecialAsset'
+import { MEDIUM_FEE_RATE } from 'utils/const'
```
Keith-CY, line 6: Unnecessary update

packages/neuron-ui/src/components/SpecialAssetList/index.tsx
```diff
@@ -100,7 +101,11 @@ const SpecialAssetList = () => {
   }, [closeNewAccount, closeExistAccount, setMigrateCell])
 
   const {
-    app: { epoch, globalDialog },
+    app: {
+      epoch,
+      globalDialog,
+      feeRateStatics: { suggestFeeRate = MEDIUM_FEE_RATE },
```
Keith-CY, line 107: IMO, the default value can be set on initialization, it's repeated multiple times

packages/neuron-ui/src/components/NervosDAO/index.tsx
```diff
@@ -244,6 +252,28 @@ const NervosDAO = () => {
     }
   }, [maxDepositAmount, depositValue, setDepositValue])
 
+  // use to update fee with suggestFeeRate
```
Keith-CY, line 255: Unnecessary comment

packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -264,6 +266,50 @@ const clear = (dispatch: StateDispatch) => {
...
+        return res
+      })
+      .catch(() => {
+        return undefined
+      })
+
+  const realParams = {
+    walletID,
+    items: items.map(item => ({
+      address: item.address || '',
```
Keith-CY, line 296: Is an empty address allowed here?

packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -264,6 +266,50 @@ const clear = (dispatch: StateDispatch) => {
...
+      capacity: CKBToShannonFormatter(item.amount, item.unit),
+      date: item.date,
+    })),
+  }
+
+  const requestArray = priceArray.map(itemPrice => getUpdateGeneratedTx({ ...realParams, feeRate: itemPrice }))
+  const allPromiseResult = await Promise.allSettled(requestArray)
+
+  const feeRateValueArray = allPromiseResult?.map((batchItem: any, index: number) => ({
```
Keith-CY, line 305: Avoid using `any`

packages/neuron-ui/src/containers/Main/index.tsx
```diff
@@ -147,14 +149,20 @@ const MainContent = () => {
   const history = useHistory()
   const location = useLocation()
   const {
-    app: { isAllowedToFetchList = true },
+    app: {
+      isAllowedToFetchList = true,
+      countDown,
```
Keith-CY, line 154: Is `countDown` necessary to be global?

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const label = isStandard ? t('price-switch.price') : t('price-switch.customPrice')
+  const percent = (countDown / DEFAULT_COUNT_DOWN) * 100
+
+  const handleDropdownChange = useCallback(
+    (e: { value: string; index?: number }) => {
+      const { value: dropdownChangedValue, index = 1 } = e
+      setDropdowValue(dropdownChangedValue)
+      setCurrentModeIdx(index)
+
+      onPriceChange?.({ target: { value: dropdownChangedValue } } as any)
```
Keith-CY, line 53: Avoid using `any`

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    },
+    [setDropdowValue, onPriceChange]
```
Keith-CY, line 55: `setCurrentModeIdx` is missing in the deps

packages/neuron-ui/src/services/remote/remoteApiWrapper.ts
JeffreyMa597: Have been removed.

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+  )
+
+  const handleInputChange = useCallback(
+    (e: React.SyntheticEvent<HTMLInputElement>) => {
+      const { value: inputChangedValue } = e.currentTarget
+      const handledValue = inputChangedValue.replace(/,/g, '')
+
+      setInputError(Number(handledValue) < 1000 ? t('price-switch.errorTip', { minPrice: 1000 }) : null)
```
Keith-CY, line 63: `handledValue` should be a valid number(expecially integer), or `0xxxxx` and `0xffff` pass

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+      onPriceChange?.(e)
+    },
+    [onPriceChange]
+  )
+
+  const feeValuesArray = useMemo(
+    () =>
+      feeRateValueArray?.map((item: any) => ({
```
Keith-CY, line 71: `any` should be avoided

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        value: shannonToCKBFormatter(item.feeValue),
+      })),
+    [feeRateValueArray]
+  )
+
+  useEffect(() => {
+    if (isStandard && !priceArray.includes(dropdownValue)) {
+      handleDropdownChange({ value: priceArray[currentModeIdx], index: currentModeIdx })
+    }
+  }, [priceArray])
```
Keith-CY, line 82: `isStandard`, `handleDropdownChange` and `currentModeIdx` are missing in the deps

packages/neuron-ui/src/components/NervosDAO/index.tsx
JeffreyMa597: Removed.

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+      setPriceArray(['1000', `${suggestFeeRate}`, `${Number(suggestFeeRate) * 2 - 1000}`])
+    }
+  }, [suggestFeeRate])
+
+  const inputHint = t('price-switch.hintTip', { suggestFeeRate })
+
+  return (
+    <div>
+      <div className={styles.pricePanel}>
+        <div style={{ display: 'flex' }}>
```
Keith-CY, line 123: inline style can be set in css file

packages/neuron-ui/src/containers/Main/index.tsx
JeffreyMa597: yep, it will be used to dispatch a request in another place.

packages/neuron-ui/src/containers/Navbar/index.tsx
Keith-CY: Don't resolve the conversation until the commit is submitted, and please append the commit in the thread

JeffreyMa597: ok

packages/neuron-ui/src/utils/calculateFee.ts
JeffreyMa597: Reset. I mean `tx.inputs` maybe a little bit more precise.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+  onChange: (changeValue: OptionProps) => void
+  options: OptionProps[]
+  value?: string
+  placeholder?: string
+  onFocus?: (isOpen: boolean) => void
+  disabled?: boolean
+  className?: string
+}
+
+const DEFAULT_PLACEHOLDER_STRING = 'Please select...'
```
Keith-CY, line 24: i18n is required

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+  options: optionsFromProps,
+  onChange,
+  value: valueFromProps = '',
+  placeholder,
+  onFocus,
+  disabled,
+  className,
+}: DropdownWithCustomRenderProps) => {
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [isMounted, setIsMounted] = useState<boolean>(true)
```
Keith-CY, line 36: initial value of `isMounted` should be `false` according to https://github.com/nervosnetwork/neuron/pull/2599/files#diff-c69edb9c0ce9ae6d9eb94be1cf1f617a43cfe450ecf1dffc51dd5cce636cc1c3R167

```
  useEffect(() => {
    document.addEventListener('click', handleDocumentClick, false)
    document.addEventListener('touchend', handleDocumentClick, false)
    setIsMounted(true)

    return () => {
      setIsMounted(false)
      document.removeEventListener('click', handleDocumentClick, false)
      document.removeEventListener('touchend', handleDocumentClick, false)
    }
  }, [])
```

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const dropdownRef = useRef<HTMLDivElement | null>(null)
+
+  // render selected value in options
+  const parseValue = (value: string, options: OptionProps[]) => {
+    const option = options.find(v => v.value !== 'undefined' && v.value === value)
```
Keith-CY, line 41: In what case `option.value` is `'undefined'`

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+      value,
+      label,
+      index,
+    }
+    fireChangeEvent(newState)
+    setSelectedValue(newState)
+    setIsOpen(false)
+  }
+
+  const handleDocumentClick = (event: any) => {
```
Keith-CY, line 70: `any` should be avoided

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+    if (isMounted) {
+      if (!dropdownRef.current?.contains(event.target)) {
+        setIsOpen(false)
+      }
+    }
```
Keith-CY, line 75: Return early to avoid nested `if` statements

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  const handleMouseDown = (event: any) => {
```
Keith-CY, line 78: `any` should be avoided

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+    if (onFocus && typeof onFocus === 'function') {
```
Keith-CY, line 79:
```suggestion
    if (typeof onFocus === 'function') {
```

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
+import React, { ReactNode, useEffect, useMemo, useRef, useState } from 'react'
```
Keith-CY, line 1: It looks too complex as a dropdown component

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+    document.addEventListener('click', handleDocumentClick, false)
+    document.addEventListener('touchend', handleDocumentClick, false)
+    setIsMounted(true)
+
+    return () => {
+      setIsMounted(false)
+      document.removeEventListener('click', handleDocumentClick, false)
+      document.removeEventListener('touchend', handleDocumentClick, false)
+    }
+  }, [])
```
Keith-CY, line 174: `setIsMounted` is missing in the deps

packages/neuron-ui/src/containers/Main/index.tsx
Keith-CY: The `fee rate` can be fetched on demand, then `count down` starts. And `count down` ends on leaving the page.

JeffreyMa597: 1、The countdown is associated with the interface call. They are synchronized, and now the interface call is on the main page, so the countdown accordingly is there. And even if the countdown is not displayed, the interface needs to be called periodically on other pages to update `fee_rate_statistics`.
2、The display of the countdown is on the send page. I use a global variable to pass this value, not context.

packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -264,6 +266,52 @@ const clear = (dispatch: StateDispatch) => {
...
+    walletID,
+    items: items.map(item => ({
+      address: item.address!,
+      capacity: CKBToShannonFormatter(item.amount, item.unit),
+      date: item.date,
+    })),
+  }
+
+  const requestArray = priceArray.map(itemPrice => getUpdateGeneratedTx({ ...realParams, feeRate: itemPrice }))
+  const allPromiseResult = await Promise.allSettled<State.Transaction>(requestArray)
```
yanguoyu, line 305: `getUpdateGeneratedTx` seems will not reject. And `Promise.allSettled` result will be `{status: 'fulfilled' | 'rejected', value: any, reason: any }`. 
It can return `State.Transaction[]`?

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const memoedMenu = useMemo(() => buildMenu(), [optionsFromProps])
+  const inputValue = <div className={placeholderClass}>{placeHolderValue}</div>
+  const menu = useMemo(
+    () =>
+      isOpen ? (
+        <div className={menuClass} aria-expanded="true">
+          {memoedMenu}
+        </div>
+      ) : null,
+    [isOpen, memoedMenu]
```
yanguoyu, line 154: Maybe you can compose the menu-created process into a function? It will be clear.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+      document.removeEventListener('click', handleDocumentClick, false)
+      document.removeEventListener('touchend', handleDocumentClick, false)
+    }
+  }, [])
+
+  return (
+    <div ref={dropdownRef} className={dropdownClass}>
+      <div
+        className={controlClass}
+        onMouseDown={(e: React.MouseEvent) => handleMouseDown(e)}
```
yanguoyu, line 175:
```suggestion
        onMouseDown={handleMouseDown}
```

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        onTouchEnd={(e: React.TouchEvent) => handleMouseDown(e)}
```
yanguoyu, line 176:
```suggestion
        onTouchEnd={handleMouseDown}
```

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const isStandard = type === PriceTypeEnum.Standard
+  const label = isStandard ? t('price-switch.price') : t('price-switch.customPrice')
+  const percent = (countDown / DEFAULT_COUNT_DOWN) * 100
+
+  const handleDropdownChange = useCallback(
+    (e: { value: string; index?: number }) => {
+      const { value: dropdownChangedValue, index = 1 } = e
+      setDropdowValue(dropdownChangedValue)
+      setCurrentModeIdx(index)
```
yanguoyu, line 51: Can we only save the mode idx, and create the dropdown value by idx and `priceArray`
Example:
When the user selects 'slow', if `priceArray` is [1000, 2000, 5000], the dropdown value will be `{key: 'slow', label: '1000'}`, and if `priceArray` is [2000, 4000, 10000], the dropdown value will be `{key: 'slow', label: '2000'}`.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { useTranslation } from 'react-i18next'
+import { LineDownArrow } from 'widgets/Icons/icon'
+
+import styles from './index.module.scss'
+
+export interface OptionProps {
+  label: string | ReactNode
+  value: string
+  className?: string
+  data?: Record<string, string | number>
```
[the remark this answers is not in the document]
JeffreyMa597: ok

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        label: placeholder ?? DEFAULT_PLACEHOLDER_STRING,
+        value: '',
+      }
+    )
+  }
+
+  const [selectedValue, setSelectedValue] = useState<OptionProps>(parseValue(valueFromProps, optionsFromProps))
+
+  const fireChangeEvent = (newState: OptionProps) => {
+    if (newState !== selectedValue && onChange) {
```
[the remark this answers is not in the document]
JeffreyMa597: updated.

packages/neuron-ui/src/widgets/DropdownWithCustomRender/index.tsx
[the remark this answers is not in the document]
JeffreyMa597: ?

yanguoyu (approved):

packages/neuron-ui/src/containers/Main/index.tsx
JeffreyMa597: > 



> Maybe you can wrapper get the fee rate as a hook, like this, then every want get `feeRate` and `countDown` by call this hook
> 
> ```ts
> const useGetFeeRate = () => {
>     const [countDown, setCountDown] = useState(30)
>     const [feeRate, setFeeRate] = useState()
>     useEffect(() => {
>         const timeout = setTimeout(() => {
>           setCountDown(v => v - 1)
>         }, 1000)
>         return () => clearTimeout(timeout)
>     }, [])
>     useEffect(() => {
>      if (countDown === 0) {
>         getFeeRateStats()
>               .then(res => {
>                 const { mean, median } = res
>                 const suggested = res ? Math.max(1000, Number(mean), Number(median)) : 0
>                 setFeeRate({ ...res, suggestFeeRate: suggested })
>               })
>               .finaly {
>                   setCountDown(30)
>                }
>         }
>      }
>     }, [countDown])
>     return { feeRate, countDown }
> }
> ```

Thx.  I submitted a new version with a hook before I saw your response. Thanks for your help again.

packages/neuron-ui/src/utils/hooks/useGetCountDownAndFeeRateStats.ts
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      setCountDown(seconds)
+    }
+
+    return () => {
+      if (timerId) {
+        clearTimeout(timerId)
+      }
+    }
+  }, [countDown, setCountDown, setTimerId])
```
yanguoyu, line 64: Every time the `count` changes, there will call setTimeout and clearTimeout, I think it's not good.

packages/neuron-ui/src/utils/calculateFee.ts
JeffreyMa597: When we didn't give the correct `address` or `account`, we cannot earn the right data. That may cause some errors like this. 
image

packages/neuron-ui/src/utils/hooks/useGetCountDownAndFeeRateStats.ts
JeffreyMa597: Update with a setInterval.

packages/neuron-ui/src/utils/calculateFee.ts
Keith-CY: The tx should be `null` if `address` is invalid, or it will be hazardous as others believe the data and skip validation.

packages/neuron-ui/src/components/PricePanel/index.tsx
JeffreyMa597: The `handledValue` has been instead by the `price` which is from props.

packages/neuron-ui/src/utils/calculateFee.ts
Keith-CY: It’s a potential risk due to legacy codebase and will be improved in future

Keith-CY (approved):

yanguoyu:
This PR may need a tester to test and approve it.

Keith-CY:
> This PR may need a tester to test and approve it.

I’ve tested the functionality yesterday.

packages/neuron-ui/src/utils/calculateFee.ts
JeffreyMa597: > The tx should be `null` if `address` is invalid, or it will be hazardous as others believe the data and skip validation.

Comments have been added.

PR #2610, opened 2023-03-16
feat: Add node version and start parameter judgement.

yanguoyu:
**The Code changes**
1. Add node version and start parameter judgment.
2. Disable the data path set button if ckb node is started externally.

**The display when start without indexer or start a different version**

1. Start ckb node external without --indexer
image
2. Start a different version ckb node
image

### DIFF
--- a/packages/neuron-wallet/src/locales/en.ts
+++ b/packages/neuron-wallet/src/locales/en.ts
@@ -217,10 +217,10 @@ export default {
         }
       },
       'node-version-different': {
-        message: '节点版本与Neuron不一致，请确认后使用'
+        message: 'The node version is inconsistent with Neuron, please use after confirmation'
       },
       'ckb-without-indexer': {
-        message: "Please add '--indexer' parameter to start local node"
+        message: "Please add '--indexer' option to start local node"
       }
     },
     prompt: {
--- a/packages/neuron-wallet/src/locales/en.ts
+++ b/packages/neuron-wallet/src/locales/en.ts
@@ -217,7 +217,7 @@ export default {
         }
       },
       'node-version-different': {
-        message: 'The node version is inconsistent with Neuron, please use after confirmation'
+        message: 'The node version is inconsistent with Neuron(v {{ version }}), please use after confirmation'
       },
       'ckb-without-indexer': {
         message: "Please add '--indexer' option to start local node"
--- a/packages/neuron-wallet/src/locales/en.ts
+++ b/packages/neuron-wallet/src/locales/en.ts
@@ -215,6 +215,12 @@ export default {
           ok: 'OK',
           cancel: 'Cancel'
         }
+      },
+      'node-version-different': {
+        message: '节点版本与Neuron不一致，请确认后使用'
+      },
+      'ckb-without-indexer': {
+        message: "Please add '--indexer' parameter to start local node"
       }
     },
     prompt: {
--- a/packages/neuron-wallet/src/locales/zh-tw.ts
+++ b/packages/neuron-wallet/src/locales/zh-tw.ts
@@ -205,10 +205,10 @@ export default {
         }
       },
       'node-version-different': {
-        message: '節點版本與Neuron不壹致，請確認後使用'
+        message: '節點版本與 Neuron 不壹致，請確認後使用'
       },
       'ckb-without-indexer': {
-        message: "請添加 '--indexer'參數來啟動本地節點"
+        message: "請添加 '--indexer' 參數來啟動本地節點"
       }
     },
     prompt: {
--- a/packages/neuron-wallet/src/locales/zh-tw.ts
+++ b/packages/neuron-wallet/src/locales/zh-tw.ts
@@ -203,6 +203,12 @@ export default {
           ok: '確認',
           cancel: '取消'
         }
+      },
+      'node-version-different': {
+        message: '節點版本與Neuron不壹致，請確認後使用'
+      },
+      'ckb-without-indexer': {
+        message: "請添加 '--indexer'參數來啟動本地節點"
       }
     },
     prompt: {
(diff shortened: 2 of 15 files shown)


packages/neuron-wallet/src/locales/en.ts
```diff
@@ -215,6 +215,12 @@ export default {
           ok: 'OK',
           cancel: 'Cancel'
         }
+      },
+      'node-version-different': {
+        message: '节点版本与Neuron不一致，请确认后使用'
```
Keith-CY, line 220:
```suggestion
        message: '节点版本与 Neuron 不一致，请确认后使用'
```

What if the desired version is prompted in the message, e.g. `Neuron (CKB Node@v0.108.1)`

packages/neuron-wallet/src/locales/en.ts
```diff
@@ -215,6 +215,12 @@ export default {
...
+      },
+      'ckb-without-indexer': {
+        message: "Please add '--indexer' parameter to start local node"
```
Keith-CY, line 223:
```suggestion
        message: "Please add '--indexer' option to start local node"
```

packages/neuron-wallet/src/locales/zh-tw.ts
```diff
@@ -203,6 +203,12 @@ export default {
           ok: '確認',
           cancel: '取消'
         }
+      },
+      'node-version-different': {
+        message: '節點版本與Neuron不壹致，請確認後使用'
```
Keith-CY, line 208:
```suggestion
        message: '節點版本與 Neuron 不一致，請確認後使用'
```

packages/neuron-wallet/src/locales/zh.ts
```diff
@@ -204,6 +204,12 @@ export default {
           ok: '确认',
           cancel: '取消'
         }
+      },
+      'node-version-different': {
+        message: '节点版本与Neuron不一致，请确认后使用'
```
Keith-CY, line 209:
```suggestion
        message: '节点版本与 Neuron 不一致，请确认后使用'
```

packages/neuron-wallet/src/locales/zh-tw.ts
```diff
@@ -203,6 +203,12 @@ export default {
...
+      },
+      'ckb-without-indexer': {
+        message: "請添加 '--indexer'參數來啟動本地節點"
```
Keith-CY, line 211:
```suggestion
        message: "請添加 '--indexer' 參數來啟動本地節點"
```

packages/neuron-wallet/src/locales/zh.ts
```diff
@@ -204,6 +204,12 @@ export default {
...
+      },
+      'ckb-without-indexer': {
+        message: "请添加 '--indexer'参数来启动本地节点"
```
Keith-CY, line 212:
```suggestion
        message: "请添加 '--indexer' 参数来启动本地节点"
```

packages/neuron-wallet/src/services/node.ts
```diff
@@ -187,6 +200,50 @@ class NodeService {
...
+        logger.error('App\t: get ckb node version failed')
+      }
+    }
+  }
+
+  private async verifyNodeVersion() {
+    const network = NetworksService.getInstance().getCurrent()
+    const localNodeInfo = await new RpcService(network.remote).getLocalNodeInfo()
+    const internalNodeVersion = this.getInternalNodeVersion()
+    if (internalNodeVersion && !localNodeInfo.version?.startsWith(internalNodeVersion.slice(1))) {
```
Keith-CY, line 220: It requires the exact same version. Maybe the version range could be accepted.

yanguoyu: It's just a notice, maybe not affecting the Neuron running.

packages/neuron-wallet/src/locales/en.ts
yanguoyu: Do you mean we need to display the internal node version and the external node version?

Keith-CY: I think showing versions in comparison would be more friendly and easy to debug with a screenshot, any idea from @Danie0918

packages/neuron-wallet/src/services/node.ts
Keith-CY: CKB node follows the semantic versioning so `major.minor` could be treated as the same one.

Usually `major` could be regarded as the same one because only `major` could have breaking changes, but now `major` is `0`, `minor` should be considers. But `patch` shouldn't break the workflow

packages/neuron-wallet/src/locales/en.ts
Danie0918: This will be better and help users understand the current wallet version.Supplementary instructions can be added as follows：

[英]The node version is inconsistent with Neuron(v 0.108.1), please use after confirmation

[中]节点版本与Neuron(v 0.108.1)不一致，请确认后使用

packages/neuron-wallet/src/services/node.ts
```diff
@@ -187,6 +200,56 @@ class NodeService {
...
+  }
+
+  private async verifyNodeVersion() {
+    const network = NetworksService.getInstance().getCurrent()
+    const localNodeInfo = await new RpcService(network.remote).getLocalNodeInfo()
+    const internalNodeVersion = this.getInternalNodeVersion()
+    const [internalMajor, internalMinor] = internalNodeVersion?.split('.') ?? []
+    const [externalMajor, externalMinor] = localNodeInfo.version?.split('.') ?? []
+
+    if (internalMajor !== externalMajor || internalMinor !== externalMinor) {
```
Keith-CY, line 226: Good enough for now, but once `major > 0`, a higher `minor` should be acceptable because it won't change breakingly.

Keith-CY (approved):

PR #2629, opened 2023-04-10
fix: Fix when enter over 8 decimal crash.

yanguoyu:
When the `depositValue` is invalid, it displays the error in the textField, but it will throw an exception when valid wheater it's big than the max deposit value, and when the `depositValue` is invalid, it's no need to set max `depositValue`.

https://user-images.githubusercontent.com/12881040/230840833-6f981c6f-22ba-4c8f-9985-88e35fa5495c.mov

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAO/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAO/index.tsx
@@ -245,6 +245,8 @@ const NervosDAO = () => {
       }
     } catch (error) {
       // ignore error
+      // When the depositValue is invalid, it displays the error in the textField, but it will throw an exception when valid wheater it's big than the max deposit value
+      // and when the depositValue is invalid, it's no need to set max depositValue.
     }
   }, [maxDepositAmount, depositValue, setDepositValue])
 
--- a/packages/neuron-ui/src/components/NervosDAO/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAO/index.tsx
@@ -239,8 +239,12 @@ const NervosDAO = () => {
   }, [])
 
   useEffect(() => {
-    if (BigInt(CKBToShannonFormatter(depositValue)) > maxDepositAmount) {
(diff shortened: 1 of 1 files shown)


Keith-CY (approved):

Keith-CY:
Please have a review @JeffreyMa597 @WhiteMinds @devchenyan

packages/neuron-ui/src/components/NervosDAO/index.tsx
```diff
@@ -239,8 +239,12 @@ const NervosDAO = () => {
...
 
   useEffect(() => {
-    if (BigInt(CKBToShannonFormatter(depositValue)) > maxDepositAmount) {
-      setDepositValue(shannonToCKBFormatter(`${maxDepositAmount}`, false, ''))
+    try {
+      if (BigInt(CKBToShannonFormatter(depositValue)) > maxDepositAmount) {
+        setDepositValue(shannonToCKBFormatter(`${maxDepositAmount}`, false, ''))
+      }
+    } catch (error) {
+      // ignore error
```
WhiteMinds, line 247: It is recommended to explicitly state that this is to ignore errors when converting to BigInt.

WhiteMinds (approved):

packages/neuron-ui/src/components/NervosDAO/index.tsx
yanguoyu: Do you mean to add this statement for the annotation?
`When the depositValue is invalid, it displays the error in the textField, but it will throw an exception when valid wheater it's big than the max deposit value, and when the depositValue is invalid, it's no need to set max depositValue.`

JeffreyMa597 (approved):

packages/neuron-ui/src/components/NervosDAO/index.tsx
Keith-CY: Will we add more statement or merge this PR first

yanguoyu: Okay, I have added.

PR #2631, opened 2023-04-12
fix: use linux portable for compatibility

Keith-CY:
ckb light client portable@v0.2.2 works on ubuntu 20.04 while ckb
light client@v0.2.2 doesn't due to openssl compatibility


yanguoyu (approved):

Keith-CY:
This PR blocks testing on linux, please have a review @WhiteMinds @devchenyan @JeffreyMa597

WhiteMinds (approved):

PR #2632, opened 2023-04-13
fix: Some optimize for light client.

yanguoyu:
1. Support export light client log
2. Support multisig for light client

image

3. If create a wallet from a new mnemonic start block with the current tip block number.
4. When creating addresses, start syncing from the last synced block number
5. Hide the data set and support clear node data when the node is light client.

image

### DIFF
--- a/packages/neuron-ui/src/utils/const.ts
+++ b/packages/neuron-ui/src/utils/const.ts
@@ -65,3 +65,4 @@ export const DEPRECATED_CODE_HASH: Record<string, string> = {
 }
 
 export const LIGHT_CLIENT_TESTNET = 'light_client_testnet'
+export const BUNDLED_LIGHT_CKB_URL = 'http://localhost:9000'
--- a/packages/neuron-ui/src/utils/const.ts
+++ b/packages/neuron-ui/src/utils/const.ts
@@ -65,4 +65,4 @@ export const DEPRECATED_CODE_HASH: Record<string, string> = {
 }
 
 export const LIGHT_CLIENT_TESTNET = 'light_client_testnet'
-export const BUNDLED_LIGHT_CKB_URL = 'http://localhost:9000'
+export const LIGHT_NETWORK_TYPE = 2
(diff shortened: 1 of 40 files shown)


packages/neuron-ui/src/utils/const.ts
```diff
@@ -65,3 +65,4 @@ export const DEPRECATED_CODE_HASH: Record<string, string> = {
 }
 
 export const LIGHT_CLIENT_TESTNET = 'light_client_testnet'
+export const BUNDLED_LIGHT_CKB_URL = 'http://localhost:9000'
```
Keith-CY, line 68: Should we use `127.0.0.1` instead of `localhost`

packages/neuron-wallet/src/listeners/main/index.ts
```diff
@@ -21,7 +21,7 @@ export const register = () => {
 
   MultisigConfigDbChangedSubject.getSubject()
     .pipe(debounceTime(500))
-    .subscribe(async (event: string) => {
+    .subscribe(async event => {
```
Keith-CY, line 24: Why is the type removed

packages/neuron-wallet/src/services/ckb-runner.ts
```diff
@@ -92,7 +92,6 @@ export const startCkbNode = async () => {
     ckb.stderr.on('data', data => {
       const dataString: string = data.toString()
       logger.error('CKB:\trun fail:', dataString)
-      ckb = null
```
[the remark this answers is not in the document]
yanguoyu: Because the `process.stderr` output log does not mean the process is closed, it may be an error log but can run continuously.

packages/neuron-wallet/src/listeners/main/index.ts
yanguoyu: Because the type can be inference by https://github.com/nervosnetwork/neuron/pull/2632/files#diff-0eadefcf991daa7cae52f65fbea7dd80eabf9f4b5cd3f0392d3dc0ec42b348f5R5

Keith-CY:
CI failed at https://github.com/nervosnetwork/neuron/actions/runs/4711528254/jobs/8355954662#step:9:1708

Keith-CY (approved):

devchenyan (approved):

PR #2635, opened 2023-04-14
fix: fix parsing deposit time in nervos dao

Keith-CY:
0 in the condition hits the negative branch but it should be treated as a positive one because 0 is a valid index of an array


yanguoyu (approved):

Keith-CY:
Please have a review @JeffreyMa597 @devchenyan @WhiteMinds

JeffreyMa597 (approved):

WhiteMinds (approved):

PR #2637, opened 2023-04-16
feat:update new ui design for history

JeffreyMa597:
------
As title.

UI: https://www.figma.com/file/6XNoimRDbFTTNm016rbIdU/Magickbase?node-id=767-8143&t=3or59J7zyaYq1XGs-0

Ref: https://github.com/Magickbase/neuron-public-issues/issues/112

Check List
------
1、List
2、Detail

### Test
e2e Test
### Task
none


yanguoyu:
It's better to upload a video to show this feature.

JeffreyMa597:
> It's better to upload a video to show this feature.

I'll upload the video after the sync is finished.

packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -46,51 +73,343 @@ const History = () => {
...
-        <div>
-          {totalCount ? (
-            <TransactionList
-              walletID={id}
-              walletName={walletName}
-              items={items as State.Transaction[]}
-              bestKnownBlockNumber={bestBlockNumber}
-              isMainnet={isMainnet}
-              dispatch={dispatch}
+      <tr>
```
yanguoyu, line 282: I think if you use `grid` layout, the element may be more clear.

packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -46,51 +73,343 @@ const History = () => {
...
+          default: {
+            // ignore
+          }
+        }
+      }
+    },
+    [isMainnet]
+  )
+
+  const rowExtendRender = (column: State.Transaction) => {
```
yanguoyu, line 275: Can we extract the `rowExtendRender` as a function component out of the current component?

packages/neuron-ui/src/components/HistoryDetailPage/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  if (error.code) {
+    return (
+      <div className={styles.error}>
+        {error.message || t(`messages.codes.${ErrorCode.FieldNotFound}`, { fieldName: 'transaction' })}
+      </div>
+    )
+  }
+
+  const columns: TableProps<InputOrOutputType>['columns'] = [
```
yanguoyu, line 189: If the `columns` is not dependent on others' variables, it's better to define it out of the component.

packages/neuron-ui/src/components/History/index.tsx
```diff
@@ -1,18 +1,44 @@
-import React, { useState, useCallback, useMemo } from 'react'
+import React, { useState, useCallback } from 'react'
 import { useNavigate, useLocation } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
-import { Stack, SearchBox } from 'office-ui-fabric-react'
+import { Stack } from 'office-ui-fabric-react'
```
devchenyan, line 4: It's unnecessary to use 'office-ui-fabric-react'

packages/neuron-ui/src/widgets/Table/table.module.scss
```diff
@@ -17,6 +17,9 @@ $head-height: 52px;
     }
   }
 }
+.noBorder {
+  border: none !important;
```
WhiteMinds, line 21: Try to avoid using "important" as much as possible. Using this keyword often implies that there are issues with our code design, and we need to carefully consider a reasonable solution.
Each "important" can become a ticking time bomb of technical debt.

WhiteMinds (approved):
I haven't carefully checked the functionality and UI, but I will verify them in the DAO's PR after this PR is merged.

packages/neuron-ui/src/widgets/Table/index.tsx
```diff
@@ -48,6 +68,7 @@ const Table = <T extends Record<string, any>>(props: TableProps<T>) => {
                   data-field={dataIndex}
                   align={align ?? 'left'}
                   className={headClassName}
+                  style={{ width, minWidth }}
```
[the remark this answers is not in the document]
JeffreyMa597: > How about using `className`?

I actually want to make these two properties a convenient generic property. So you don't have to use a specific class every time you use it. Because each column might have a different width, it would take a lot of classes to handle it, especially if I only had to set the width.

packages/neuron-ui/src/widgets/TextField/textField.module.scss
```diff
@@ -40,10 +42,10 @@
     justify-content: flex-start;
     align-items: stretch;
     box-sizing: border-box;
-    background: #FFFFFF;
+    background: #ffffff;
     border: 1px solid var(--divide-line-color);
     border-radius: 8px;
-    width: 480px;
+    width: 100%;
```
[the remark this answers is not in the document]
JeffreyMa597: image

Have been make a double-check with this change, it makes the same function in other features.
Also, `width: 100%` seems better. Is that right?

devchenyan:
image

The transfer amount should be highlighted

image

devchenyan:
font-family: 'JetBrains Mono'

image

image

image

devchenyan:
image

"widgets/Dialog" component implements the close by default. we can use it directly

image

packages/neuron-ui/src/styles/theme.scss
```diff
@@ -34,6 +34,18 @@ body {
   // table
   --table-head-background-color: #FAFAFA;
   --table-head-border-color: #F0F0F0;
+  // history
+  --history-button-border: #e5e5e5;
+  --history-info-title-color: #999;
+  --history-button-color: #8da394;
+  --history-button-hover: #00c891;
```
devchenyan, line 41: I think we can use "--primary-color" directly

packages/neuron-ui/src/widgets/Tabs/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  const tabStyle = (tab: string | number) => {
+    return value === tab ? `${styles.tab} ${styles.active}` : styles.tab
+  }
+
+  return (
+    <div className={styles.tabs}>
+      {tabs.map(({ label, value: val }) => (
+        <button type="button" key={val} className={tabStyle(val)} onClick={() => handleTabClick(val)}>
```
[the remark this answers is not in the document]
devchenyan: we can use data-value and get value with dataset

packages/neuron-ui/src/components/HistoryDetailPage/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+  const {
+    app: { pageNotice },
+  } = useGlobalState()
+  const [t] = useTranslation()
+  const [transaction, setTransaction] = useState(transactionState)
+  const [isMainnet, setIsMainnet] = useState(false)
+  const [error, setError] = useState({ code: '', message: '' })
+  const [lockInfo, setLockInfo] = useState<CKBComponents.Script | null>(null)
+  const hash = useMemo(() => window.location.href.split('/').pop(), [])
+
```
devchenyan, line 49: maybe we can use "useParams" to get hash

packages/neuron-ui/src/widgets/Table/index.tsx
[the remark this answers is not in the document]
yanguoyu: Gotcha

packages/neuron-ui/src/widgets/TextField/textField.module.scss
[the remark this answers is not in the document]
yanguoyu: I think it's better to change the container's width to `480px` which includes `TextField`.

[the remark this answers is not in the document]
yanguoyu: My fault, I misread it was changing from `100%` to `480px`.

packages/neuron-ui/src/components/History/index.tsx
JeffreyMa597: Maybe to be a function it will be more convenient to pass the data from the `Table`.

JeffreyMa597:
image

The `padding` of the content in the component of `widget/Dialog` looks so strange. IMO, we can give it the same padding or unset it. I'll change it to the same value and check all the `paddings` where we use this `widget/Dialog`.

WhiteMinds:
> The `padding` of the content in the component of `widget/Dialog` looks so strange. IMO, we can give it the same padding or unset it. I'll change it to the same value and check all the `paddings` where we use this `widget/Dialog`.

The new design for this dialog has been implemented in https://github.com/nervosnetwork/neuron/pull/2645. Some other similar parts can also be copied and pasted. I suggest that we wait until we finish the entire new version UI before considering refactoring and reusing the cod

devchenyan:
> image
> 
> The `padding` of the content in the component of `widget/Dialog` looks so strange. IMO, we can give it the same padding 


> image
> 
> The `padding` of the content in the component of `widget/Dialog` looks so strange. IMO, we can give it the same padding or unset it. I'll change it to the same value and check all the `paddings` where we use this `widget/Dialog`. @yanguoyu @devchenyan @WhiteMinds

I don't think the padding needs to be set to the same value.

devchenyan (approved):

packages/neuron-ui/src/components/History/index.tsx
JeffreyMa597: I had the same idea before, but in practice, it’s difficult to get the same width as each column in this list.

packages/neuron-ui/src/components/HistoryDetailPage/index.tsx
```diff
@@ -0,0 +1,282 @@
...
+          if (!network) {
+            throw new Error('Cannot find current network in the network list')
+          }
+
+          setIsMainnet(network.chain === MAINNET_TAG)
+        }
+      })
+      .catch(err => console.warn(err))
+
+    const currentWallet = currentWalletCache.load()
```
yanguoyu, line 64: `currentWallet` also can get from `useGlobalState`'s `wallet`

packages/neuron-ui/src/components/LockInfoDialog/index.tsx
```diff
@@ -66,30 +84,22 @@ const LockInfoDialog = ({ lockInfo, isMainnet, onDismiss }: LockInfoDialogProps)
...
-            onClick={() => handleCopy(shortAddr)}
-            onKeyPress={onEnter(() => handleCopy(shortAddr))}
-            role="none"
-          >
-            {copied ? <Check /> : <Copy />}
+          <div className={styles.lock}>
+            <div
+              className={styles.copyBtn}
+              onClick={handleCopy}
+              onKeyPress={e => (e.key === 'enter' ? handleCopy : undefined)}
```
yanguoyu, line 95: Use `onEnter`  https://github.com/nervosnetwork/neuron/blob/new-ui-design/packages/neuron-ui/src/utils/inputDevice.ts#L3

packages/neuron-ui/src/widgets/SimpleTabs/index.tsx
yanguoyu, line 1: Do you have a plan to use the widget tabs to replace this?

JeffreyMa597: I've checked the widget Tabs, it's different from this SimpleTabs. And, as a common component, the widget Tabs may need some optimization. I think it's better to replace it after we finished them.

yanguoyu (approved):

PR #2643, opened 2023-04-21
ci: fix the add-label action

Keith-CY:
This action(https://github.com/nervosnetwork/neuron/actions/runs/4760813903)
was triggered by a comment in pull request because pull request
is also regarded as an issue in GitHub. And this action threw
an exception that `actions-ecosystem/action-add-label` is not
found.

This commit fixed the typo of `actions-ecosystem/action-add-labels`
And add a step to filter pull request out

Ref:
actions-ecosystem/action-add-labels: https://github.com/actions-ecosystem/action-add-labels

### DIFF
--- a/.github/workflows/add-replied-label.yml
+++ b/.github/workflows/add-replied-label.yml
@@ -22,9 +22,23 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - id: check-issue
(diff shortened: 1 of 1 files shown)


Keith-CY:
This comment is used to test if the comments in pull requests are filtered out by this action

devchenyan (approved):

yanguoyu (approved):

PR #2651, opened 2023-04-25
feat: Add dialog warn for migrate date to 0.108

yanguoyu:
When we get the `migrate` log from `ckb` node log.
If the user has clicked migrate, it will continue to migrate. If the user has not clicked migrate, there will show the migrate dialog.
After clicking `Cancel`, the dialog will hide until the user refreshes the page or reopen `Neuron`.
When clicking `Backup`, `Neuron` will open the data dictionary, when clicking `migrate`, `Neuron` will record the migrated version. Then If we need to migrate in the future, we only need to change the migrated version.

https://user-images.githubusercontent.com/12881040/234271807-2201f1a4-c3c7-42d7-9c5c-d67359bc7ebf.mov

### DIFF
--- a/packages/neuron-ui/src/containers/Notification/index.tsx
+++ b/packages/neuron-ui/src/containers/Notification/index.tsx
@@ -80,8 +80,8 @@ export const NoticeContent = () => {
   const dispatch = useDispatch()
   const [t, i18n] = useTranslation()
   useOnLocaleChange(i18n)
-  const [hasDismiss, setHasDismiss] = useState(false)
-  useGlobalNotifications(dispatch, hasDismiss)
+  const [hasDismissMigrate, setHasDismissMigrate] = useState(false)
+  useGlobalNotifications(dispatch, hasDismissMigrate)
 
   const notificationsInDesc = useMemo(() => [...notifications].reverse(), [notifications])
   const notification: State.Message | undefined = notificationsInDesc[0]
@@ -103,8 +103,10 @@ export const NoticeContent = () => {
 
   const onGlobalDialogDismiss = useCallback(() => {
     dismissGlobalDialog()(dispatch)
-    setHasDismiss(true)
-  }, [dispatch])
+    if (globalDialog === 'rebuild-sync') {
+      setHasDismissMigrate(true)
+    }
+  }, [dispatch, globalDialog])
 
   const onOk = useCallback(() => {
     migrateData().then(res => {
--- a/packages/neuron-ui/src/containers/Notification/index.tsx
+++ b/packages/neuron-ui/src/containers/Notification/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useCallback, MouseEventHandler } from 'react'
+import React, { useMemo, useCallback, MouseEventHandler, useState } from 'react'
 import { createPortal } from 'react-dom'
 import { useTranslation } from 'react-i18next'
 import { Stack, MessageBar, MessageBarType, IconButton, Panel, PanelType, Text } from 'office-ui-fabric-react'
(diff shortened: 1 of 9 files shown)


packages/neuron-ui/src/containers/Notification/index.tsx
```diff
@@ -101,17 +103,34 @@ export const NoticeContent = () => {
 
   const onGlobalDialogDismiss = useCallback(() => {
     dismissGlobalDialog()(dispatch)
+    setHasDismiss(true)
```
Keith-CY, line 106: It works in this `dismiss migration` case because `migration` notification always prompts first.

But it's kind of sloppy based on the assumption that `migration` notification always shows first.

I notice that `global notification` only works for the migration for now, maybe we can revise it by renaming the component.

packages/neuron-ui/src/containers/Notification/index.tsx
```diff
@@ -79,7 +80,8 @@ export const NoticeContent = () => {
   const dispatch = useDispatch()
   const [t, i18n] = useTranslation()
   useOnLocaleChange(i18n)
-  useGlobalNotifications(dispatch)
+  const [hasDismiss, setHasDismiss] = useState(false)
+  useGlobalNotifications(dispatch, hasDismiss)
```
Keith-CY, line 84: `hasDismiss` can be used to hide the `notification component` so the logic of `onBackUp` can be intact.

packages/neuron-ui/src/utils/hooks/index.ts
```diff
@@ -433,27 +433,31 @@ export const useOnHandleNetwork = ({ history }: { history: ReturnType<typeof use
...
-      dispatch({
-        type: AppActions.SetGlobalDialog,
-        payload: 'rebuild-sync',
-      })
-    }
     const migrateSubscription = Migrate.subscribe(migrateStatus => {
-      if (!isVersionUpdate && migrateStatus === 'need-migrate') {
-        migrateData()
-        migrateSubscription.unsubscribe()
+      if (migrateStatus === 'need-migrate') {
```
Keith-CY, line 443: Return early to avoid nested `if` statements

devchenyan (approved):

packages/neuron-ui/src/containers/Notification/index.tsx
yanguoyu: The `hasDismiss` variable is only used to judge whether the user have clicks dismiss for migrate, so how about rename `hasDismiss` to `hasDismissMigrate`?

packages/neuron-ui/src/containers/Notification/index.tsx
yanguoyu: I didn't get the point about `onBackUp can be intact`

packages/neuron-ui/src/containers/Notification/index.tsx
Keith-CY: From the codebase, we can see `global notification` is only for `migration`, but its name implies that it is target-agonistic(no business keywords in the component name). So it may deliver other messages in the future.

However, `dismiss/dismissMigrate` is triggered 100% by `onGlobalDialogDismiss`.

Imagine that there're 2 messages on launch
1. `There's an update of Neuron`
2. `A migration is required`

When the global notification of an update is dismissed, the `hasDismissMigrate` will be set `true` unexpectedly.

packages/neuron-ui/src/containers/Notification/index.tsx
Keith-CY: My fault, `useGlobalNotifications` can be intact because `dismiss` will not be triggered second time if the dialog is hidden by `hasDismiss`

packages/neuron-ui/src/containers/Notification/index.tsx
yanguoyu: Currently, `global notification` also includes `Asset was released, view details in History.`
So, maybe I need to rename `onGlobalDialogDismiss ` to `onDismissMigrate`.

Keith-CY: > Currently, `global notification` also includes `Asset was released, view details in History.` So, maybe I need to rename `onGlobalDialogDismiss ` to `onDismissMigrate`.

It would be clear

Keith-CY (approved):

packages/neuron-ui/src/containers/Notification/index.tsx
yanguoyu: I found that maybe other global dialog needs to dismiss when clicking the mask, so I add a check whether the global dialog type is `rebuild-sync`  when calling `setHasDismissMigrate`. The commit is https://github.com/nervosnetwork/neuron/pull/2651/commits/d1c0d598a68425d6c245bf7a2a64bf4634bbcc33

PR #2652, opened 2023-04-27
fix: failed to connect to ckb rpc after upgrading electron to v23

WhiteMinds:
fix: failed to connect to ckb rpc after upgrading electron to v23


yanguoyu:
CI failed.

yanguoyu (approved):

WhiteMinds:
> CI failed.

Solved, I used `src` to filter the search results when processing before and forgot to have `tests`.

packages/neuron-wallet/tests/services/node.test.ts
Keith-CY, line 1: Should add a test case to check if `localhost` is resolved to ipv4 instead of ipv6

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -176,3 +183,17 @@ export default class NetworksService extends Store {
     return network
   }
 }
+
+function applyLocalhostIPv4Resolve(url: string): string {
```
Keith-CY, line 187: How about using `dns.setDefaultResultOrder('ipv4first')`

Keith-CY: Ah I see the discussion at https://github.com/Magickbase/neuron-public-issues/issues/122#issuecomment-1524625828

WhiteMinds: > How about using `dns.setDefaultResultOrder('ipv4first')`

If we use this API, it is highly likely to become a technical debt because it may affect other DNS resolutions, which may lead us to make decisions to write other code with a bias towards using IPv4, while the future should embrace IPv6.

Keith-CY: Got it, please add this consideration in the issue https://github.com/Magickbase/neuron-public-issues/issues/122#issuecomment-1526892427

packages/neuron-wallet/tests/services/node.test.ts
WhiteMinds: Makes sense, I added it

Keith-CY (approved):

PR #2645, opened 2023-04-24
Implement the new UI for the NervosDAO

WhiteMinds:
Implement the new UI for the NervosDAO

Preview:

https://user-images.githubusercontent.com/9160743/234505195-90d7dfd9-56fa-4598-9148-4f1618d90eb8.mp4

PR #2537, opened 2022-11-29
Chore: remove uifabric experiments and uifabric styling

JeffreyMa597:
What problem does this PR solve?
------
As title.

Ref: https://github.com/Magickbase/neuron-public-issues/issues/68

Check List
------

### Test
e2e Test
### Task
none

PR #2654, opened 2023-05-06
Feat: adopt new UI

devchenyan:
resolve the conflicts between develop and new-ui-design


packages/neuron-ui/package.json
```diff
@@ -40,24 +40,25 @@
     "last 2 chrome versions"
   ],
   "dependencies": {
-    "@nervosnetwork/ckb-sdk-core": "0.103.1",
-    "@nervosnetwork/ckb-sdk-utils": "0.103.1",
+    "@nervosnetwork/ckb-sdk-core": "0.107.0",
+    "@nervosnetwork/ckb-sdk-utils": "0.107.0",
+    "@uifabric/experiments": "7.45.14",
+    "@uifabric/styling": "7.25.1",
```
yanguoyu, line 46: These have removed by https://github.com/nervosnetwork/neuron/pull/2537/files

packages/neuron-ui/package.json
```diff
@@ -71,33 +72,42 @@
     "@storybook/preset-create-react-app": "4.1.2",
     "@storybook/react": "6.5.12",
     "@storybook/testing-library": "0.0.13",
+    "@storybook/addon-knobs": "5.3.21",
+    "@storybook/addon-storyshots": "5.3.18",
+    "@storybook/addon-viewport": "5.3.18",
+    "@storybook/addons": "5.3.18",
+    "@storybook/cli": "5.3.18",
```
yanguoyu, line 79: These `@storybook`'s plugins should be removed. `@storybook` has been updated to 6.5.12 by https://github.com/nervosnetwork/neuron/pull/2506/files#diff-0563a7b3e27fdf2d3e488aa5c618a241c5ba435a435c120acfa80a357a307a20

packages/neuron-ui/package.json
yanguoyu, line 1: Some dependencies and devDependencies about `storybook` should refer to this [commit](https://github.com/nervosnetwork/neuron/pull/2506/files#diff-0563a7b3e27fdf2d3e488aa5c618a241c5ba435a435c120acfa80a357a307a20)

packages/neuron-ui/src/components/NervosDAO/hooks.ts
```diff
@@ -539,7 +554,7 @@ export const useUpdateDepositEpochList = ({
           const epochList = new Map()
           records.forEach(record => {
             const key = getRecordKey(record)
-            epochList.set(key, recordKeyIdxMap.get(key) ? res[recordKeyIdxMap.get(key)!]?.epoch : null)
+            epochList.set(key, recordKeyIdxMap.get(key) !== undefined ? res[recordKeyIdxMap.get(key)!]?.epoch : null)
```
yanguoyu, line 557:
```suggestion
            epochList.set(key, recordKeyIdxMap.has(key) ? res[recordKeyIdxMap.get(key)!]?.epoch : null)
```

yanguoyu:
After the dao feature https://github.com/nervosnetwork/neuron/pull/2645 merges into the new-ui branch, you need to resolve some conflicts again.

WhiteMinds:
I suggest using `git rebase` instead of `git merge` as much as possible to handle conflicts, because `merge` will make it difficult for `reviewers` to distinguish what changes the `merge-commit` has made to handle the conflict. Even using `git diff-tree --cc [merge-commit]` will have a lot of redundant information. In contrast, the difference between before and after resolving the conflict using `rebase` can be compared using `git range-diff`.

If you choose `merge` because of the branch that is developed based on this branch, it is also possible to use `rebase`. You can handle it on the child branch through `git rebase --onto`.

Of course, `rebase` may be more troublesome than `merge` because it requires processing one commit at a time, but the benefits it brings are that the `reviewer`'s efficiency is improved, and it is easier to identify problems. Therefore, I think the time cost of the `rebase` solution is acceptable.

yanguoyu (approved):

PR #2657, opened 2023-05-08
docs: proxy when install

homura:
Add some description for speed up electron download

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ $ yarn global add lerna # install lerna globally
 
 #### Install Dependencies
 
-> TIP: If you find that it takes a long time to download electron, you can try [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) by
-> 
-> export ELECTRON_GET_USE_PROXY=http://username:password@host:port
+> Notes: 
+> 1. Please set [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) if you are using a proxy locally and find the downloading is halted
+> 2. Please set [CDN](https://www.electronjs.org/docs/latest/tutorial/installation#mirror) if the downloading is slow
 
 After lerna has been installed, run this to install and link dependencies:
 
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@ $ yarn global add lerna # install lerna globally
 
 #### Install Dependencies
 
+> TIP: If you find that electron takes a long time to download, you can try proxy by 
+> export ELECTRON_GET_USE_PROXY=http://username:password@host:port
+
 After lerna has been installed, run this to install and link dependencies:
 
 ```shell
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ $ yarn global add lerna # install lerna globally
 
 #### Install Dependencies
 
-> TIP: If you find that electron takes a long time to download, you can try [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) by
+> TIP: If you find that it takes a long time to download electron, you can try [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) by
 > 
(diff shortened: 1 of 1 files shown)


WhiteMinds:
Perhaps providing a link to https://www.electronjs.org/docs/latest/tutorial/installation#proxies would be more detailed?

WhiteMinds (approved):

yanguoyu (approved):

README.md
```diff
@@ -23,6 +23,10 @@ $ yarn global add lerna # install lerna globally
 
 #### Install Dependencies
 
+> TIP: If you find that electron takes a long time to download, you can try [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) by
```
Keith-CY, line 26:
```suggestion
> TIP: If you find that it takes a long time to download electron, you can try [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) by
```

devchenyan (approved):

JeffreyMa597 (approved):

README.md
Keith-CY: > For me yes, I think most developers use proxy to bypass a firewall

Then how about mentioning it clearly that `proxy` is specific for `proxy` and `firewall` instead of the downloading speed? By doing this, `proxy` user will notice it soon while other users won't care about it.

zhangyouxin: Actually I do have a proxy with ClashX, and I tried to set the proxy by `export ELECTRON_GET_USE_PROXY=http://localhost:7890`, but it's weird that electron-downloader says the port refuses to connect, then i had to use CDN mirror

yanguoyu: Actually, I use  `ELECTRON_MIRROR ` with `electron` at the previous company. I also think `ELECTRON_MIRROR` is enough to download the `electron` binary file.

Keith-CY: So can we regard the expression as imprecise? `proxy` is for users blocked by the firewall, while CDN is used to boost download. @homura

homura: The firewall is not only ban related sites, it also has the function of speed limit, in fact, when I download Electron, it should be speed limit, otherwise if it is inaccessible, the download connection should be disconnected

Keith-CY: > The firewall is not only ban related sites, it also has the function of speed limit, in fact, when I download Electron, it should be speed limit, otherwise if it is inaccessible, the download connection should be disconnected

`proxy` accelerates download in some cases, like users blocked by the firewall. But it doesn't accelerate for all users. So why not annotate it clear, to tell `blocked users` to use `proxy` while other users to user CDN?

homura: So I say `if you find ..., you can try`

Keith-CY: > So I say `if you find ..., you can try`, and as a "Quick Start", I personally think we can cover the scenario with a common solution. As for CDN, the mirror site we give out does not always have coverage to the user's location

However, `proxy` is a really special case specifically for users, mainly those blocked by GFW. And most users all over the world are not blocked by GFW.

If only `proxy` is mentioned here along with `download slow`, and our users are not using a proxy, they won't get helped if they want to speed download up.

I would prefer an expression as follows.

```
1. Please set [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) if you are using a proxy locally and find the downloading is halted
2. Please set [CDN](https://www.electronjs.org/docs/latest/tutorial/installation#mirror) if the downloading is slow
```

BTW, I haven't used `proxy`, too.

homura: > 1. Please set [proxy](https://www.electronjs.org/docs/latest/tutorial/installation#proxies) if you are using a proxy locally and find the downloading is halted
> 2. Please set [CDN](https://www.electronjs.org/docs/latest/tutorial/installation#mirror) if the downloading is slow

make sense, I'll change it

Keith-CY (approved):

PR #2659, opened 2023-05-09
fix: fix some bugs

yanguoyu:
1. When append scripts are empty means remove append scripts, so remove empty judgment
2. resultFormatters should not return null. Because the Method will ignore null or undefined.
3. When clear cache start monitor should start node right now and only delete current wallet sync progress.
4. Save decimal number to SQLite when the field is `timestamp` or `blockNumber`


Keith-CY:
I was reviewing commit [fix: Save number to sqlite when field is timestamp or blockNumber](https://github.com/nervosnetwork/neuron/pull/2659/commits/101826517d7b02e494ecd36712df6ef550599226) and found that it just removed an `non-nullable assertion`, how did it save `number` to sqlite

Keith-CY (approved):

devchenyan (approved):

PR #2661, opened 2023-05-10
fix: Add min-height avoid dialog truncation

yanguoyu:
1. before the change

image

2. after the change

image

### DIFF
--- a/packages/neuron-ui/src/components/SUDTUpdateDialog/sUDTUpdateDialog.module.scss
+++ b/packages/neuron-ui/src/components/SUDTUpdateDialog/sUDTUpdateDialog.module.scss
@@ -7,6 +7,7 @@
   left: 0;
   height: 100%;
   width: 100%;
+  min-height: 550px;
   display: flex;
   justify-content: center;
   align-items: center;


Keith-CY:
Seems that the modal is not 100% full of the width and height

yanguoyu:
> Seems that the modal is not 100% full of the width and height

Because this main page https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/containers/Main/index.tsx#L188 is not 100% height, If we change the main page to 100%, we may need to check other functions.
I think we can fix the main page to 100% in the new-ui branch.

Keith-CY:
> > Seems that the modal is not 100% full of the width and height
> 
> Because this main page [`develop`/packages/neuron-ui/src/containers/Main/index.tsx#L188](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/containers/Main/index.tsx?rgh-link-date=2023-05-10T03%3A17%3A05Z#L188) is not 100% height, If we change the main page to 100%, we may need to check other functions. I think we can fix the main page to 100% in the new-ui branch.

Reasonable

Keith-CY (approved):

JeffreyMa597 (approved):

PR #2663, opened 2023-05-10
fix: Fix deposit dao

yanguoyu:
1. When refreshing in DAO page, the network is not ready, so ignore 404.
2. Reset deposited DAO capacity when opening dialog to fix https://github.com/Magickbase/neuron-public-issues/issues/146.
3. Create a common `generateDaoDepositTx` hook to create a deposit tx whether `despositeValue` change or `suggestFeeRate` change.

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAO/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAO/index.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState, useMemo, useCallback } from 'react'
-import { useState as useGlobalState, useDispatch, AppActions } from 'states'
+import { useState as useGlobalState, useDispatch } from 'states'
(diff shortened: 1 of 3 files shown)


packages/neuron-ui/src/components/NervosDAO/index.tsx
```diff
@@ -407,7 +389,7 @@ const NervosDAO = () => {
         <Button
           type="primary"
           disabled={connectionStatus === 'offline' || sending || !maxDepositTx}
-          onClick={() => setShowDepositDialog(true)}
+          onClick={onOpenDepositeDialog}
```
Keith-CY, line 392:
```suggestion
          onClick={onOpenDepositDialog}
```

Keith-CY (approved):

JeffreyMa597 (approved):

PR #2666, opened 2023-05-13
feat: Multisig Address with new ui

devchenyan:
- Light Theme

https://github.com/nervosnetwork/neuron/assets/11756869/3b4c5fc6-b8ad-46bc-96b8-8cba2867e0dc

- Dark Theme

https://github.com/nervosnetwork/neuron/assets/11756869/b2e864ee-b646-42e4-9b60-35acdce08786


yanguoyu:
If we do not show `multisig` and `sign-verify` in a new window, how about only using a dialog component in <App />, but not using a router to show these dialogs?
What is the difference between `using router` and `use dialog component`?

packages/neuron-wallet/src/controllers/app/menu.ts
```diff
@@ -64,14 +63,15 @@ const showAbout = () => {
...
       type: 'info',
       title: app.name,
       message: app.name,
       detail: applicationVersion,
       buttons: ['OK'],
-      cancelId: 0,
-    })
+      cancelId: 0
+    }
+    dialog.showMessageBox(options)
```
yanguoyu, line 74: Here should use develop branch's code.

silySuper:
@devchenyan please handle this review ，thank you～

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -45,7 +37,7 @@ const UpdateDownloadStatus = ({
     return (
       <Dialog
         show={show}
-        onConfirm={handleDownload}
+        onConfirm={() => downloadUpdate()}
```
yanguoyu, line 40:
```suggestion
        onConfirm={downloadUpdate}
```

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -67,7 +59,7 @@ const UpdateDownloadStatus = ({
         show={show}
         onCancel={onCancel}
         showCancel={false}
-        onConfirm={handleInstall}
+        onConfirm={() => installUpdate()}
```
yanguoyu, line 62:
```suggestion
        onConfirm={installUpdate}
```

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -37,6 +37,7 @@ const UpdateDownloadStatus = ({
     return (
       <Dialog
         show={show}
+        // @ts-ignore
```
yanguoyu, line 40: `ts-ignore` should not be added unless we can only do this, and add it with annotation

packages/neuron-ui/src/widgets/Table/index.tsx
```diff
@@ -50,6 +51,8 @@ const Table = <T extends Record<string, any>>(props: TableProps<T>) => {
     setExpandedRow(prevIndex => (prevIndex === index ? null : index))
   }
 
+  const theColumns = useMemo(() => columns.filter(item => !item.hidden), [columns])
```
yanguoyu, line 54: `theColumns` has no clear meaning

packages/neuron-ui/src/components/ScriptTag/index.tsx
```diff
@@ -41,7 +43,7 @@ const ScriptTag = ({
   }
 
   return (
-    <button type="button" className={styles.tag} onClick={onClick}>
+    <button type="button" className={className || styles.tag} onClick={onClick}>
```
yanguoyu, line 46: Why should we cover all of the styles of `styles.tag`?

packages/neuron-ui/src/components/MultisigAddress/hooks.ts
```diff
@@ -68,8 +80,9 @@ export const useConfigManage = ({ walletId, isMainnet }: { walletId: string; isM
     })
   }, [setEntities, walletId])
   const updateConfig = useCallback(
-    (id: number) => (alias: string | undefined) => {
-      updateMultisigConfig({ id, alias: alias || '' }).then(res => {
+    (id: number) => (e: React.SyntheticEvent<any>) => {
+      const { value } = e.target as HTMLInputElement
```
WhiteMinds, line 84:
```suggestion
    (id: number) => (e: React.SyntheticEvent<unknown>) => {
      if (!(e.target instanceof HTMLInputElement)) return
      const { value } = e.target
```

WhiteMinds (changes requested):

packages/neuron-ui/src/components/ScriptTag/index.tsx
devchenyan: When merging the code, it was found that the style of ScriptTag was rewritten in Nervos DAO. So I moved the className into Mutilsig Address and passed the prop to ScriptTag.

yanguoyu: You have covered all of the styles in packages/neuron-ui/src/components/ApproveMultisigTxDialog/approveMultisigTx.module.scss, I guess we can use ``{`${styles.tag} ${className}`}``

yanguoyu: image
Kirl has set them to the same style.

packages/neuron-ui/src/components/MultisigAddress/hooks.ts
JeffreyMa597: What's the difference between removing `any` and updating in `unknown`?

packages/neuron-ui/src/components/MultisigAddress/index.tsx
```diff
@@ -1,43 +1,51 @@
...
+import { useOnLocaleChange, isMainnet as isMainnetUtil, shannonToCKBFormatter, useExitOnWalletChange } from 'utils'
 import { useState as useGlobalState, withProvider } from 'states'
 import MultisigAddressCreateDialog from 'components/MultisigAddressCreateDialog'
-import CopyZone from 'widgets/CopyZone'
-import { More } from 'widgets/Icons/icon'
-import { CustomizableDropdown } from 'widgets/Dropdown'
 import MultisigAddressInfo from 'components/MultisigAddressInfo'
 import SendFromMultisigDialog from 'components/SendFromMultisigDialog'
-import { EditTextField } from 'widgets/TextField'
+import { ReactComponent as Edit } from 'widgets/Icons/Edit.svg'
```
JeffreyMa597, line 9: How about putting these `widgets/` images together?

packages/neuron-ui/src/components/MultisigAddress/multisigAddress.module.scss
```diff
@@ -1,115 +1,197 @@
...
-    word-break: break-all;
-    text-align: right;
-    line-height: 1.625rem;
-    overflow: hidden;
+    @include checkbox;
   }
 
-  & > span {
-    line-height: 1.625rem;
+  .addressTr {
```
JeffreyMa597, line 92: Sounds strange.

packages/neuron-ui/src/components/MultisigAddress/multisigAddress.module.scss
```diff
@@ -1,115 +1,197 @@
...
+    cursor: pointer;
+    padding-right: 30px;
     &:hover {
-      path {
-        fill: var(--nervos-green);
-        stroke: var(--nervos-green);
+      color: var(--primary-color);
+    }
+  }
+  .actionTr {
```
JeffreyMa597, line 142: This name sounds strange.

packages/neuron-ui/src/components/MultisigAddressInfo/index.tsx
```diff
@@ -1,9 +1,11 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import TextField from 'widgets/TextField'
 import CopyZone from 'widgets/CopyZone'
+import { Copy } from 'widgets/Icons/icon'
 import { ErrorWithI18n } from 'exceptions'
-
+import Table from 'widgets/Table'
```
JeffreyMa597, line 7: How about moving them together?

packages/neuron-ui/src/components/ScriptTag/index.tsx
devchenyan: thanks, i will fix it

packages/neuron-ui/src/components/MultisigAddressInfo/index.tsx
```diff
@@ -22,57 +24,62 @@ export const MultisigAddressTable = ({
...
-                  name={t('multisig-address.create-dialog.copy-address')}
-                >
-                  <span className={styles.overflow}>{v.slice(0, -6)}</span>
-                  <span>...</span>
-                  <span>{v.slice(-6)}</span>
-                </CopyZone>
-              ) : (
+    <div className={styles.tableWrap}>
+      <Table
+        columns={[
```
JeffreyMa597, line 31: How about moving the columns out?

packages/neuron-ui/src/components/SendFieldset/index.tsx
```diff
@@ -42,6 +43,7 @@ const SendFieldset = ({
   onSendMaxClick,
   onItemChange,
   isTimeLockable = true,
+  className = '',
```
JeffreyMa597, line 46: The default value is unnecessary.

packages/neuron-ui/src/components/SendFieldset/index.tsx
```diff
@@ -64,7 +66,7 @@ const SendFieldset = ({
   }
 
   return (
-    <div className={styles.container}>
+    <div className={`${styles.container} ${className}`}>
```
JeffreyMa597, line 69: We can use `clsx` from the `utils`.

packages/neuron-ui/src/widgets/Dialog/dialog.module.scss
```diff
@@ -43,12 +46,28 @@
 }
 
 .content {
-  padding: 20px 16px 0;
+  padding: 20px 16px;
   position: relative;
+  max-height: calc(90vh - 200px);
```
JeffreyMa597, line 51: Where does the `max-height` come from?

packages/neuron-ui/src/components/ScriptTag/index.tsx
JeffreyMa597: > You have covered all of the styles in packages/neuron-ui/src/components/ApproveMultisigTxDialog/approveMultisigTx.module.scss, I guess we can use `` {`${styles.tag} ${className}`} ``

We can use `clsx` from our `utils`.

packages/neuron-ui/src/components/ApproveMultisigTxDialog/approveMultisigTx.module.scss
```diff
@@ -0,0 +1,180 @@
...
+    padding: 0 16px;
+    font-size: 14px;
+    display: flex;
+    align-items: center;
+    background: var(--table-head-background-color);
+    color: var(--secondary-text-color);
+    border-bottom: 1px solid var(--table-head-border-color);
+
+    .tip {
+      left: calc(100% - 240px) !important;
```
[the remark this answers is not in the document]
devchenyan: It is necessary

yanguoyu (approved):

packages/neuron-ui/src/components/MultisigAddress/hooks.ts
WhiteMinds: > What's the difference between removing `any` and updating in `unknown`?

`any` and `as` can break the type system, while `unknown` is also a special case in the type system. It accepts any type, but when using or transferring the value to another type that is not `unknown` or `any`, it needs to be converted to a valid type first. This avoids breaking the type system.

packages/neuron-ui/src/components/ApproveMultisigTxDialog/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+            </button>
+            <div className={styles.underline} />
+          </div>
+          {tabIdx === '0' ? (
+            <textarea disabled value={jsonContent} className={styles.textarea} />
+          ) : (
+            <div className={styles.conciseData}>
+              <div className={styles.inputWrap}>
+                <h2>Inputs</h2>
+                {offlineSignJson.transaction?.inputs?.map((v: any) => (
```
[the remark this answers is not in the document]
JeffreyMa597: The type still `any`.

packages/neuron-ui/src/services/remote/updater.ts
```diff
@@ -1,6 +1,6 @@
 import { remoteApi } from './remoteApiWrapper'
 
 export const checkForUpdates = remoteApi<void>('check-for-updates')
-export const downloadUpdate = remoteApi<void>('download-update')
-export const installUpdate = remoteApi<void>('quit-and-install-update')
+export const downloadUpdate = remoteApi<any>('download-update')
```
JeffreyMa597, line 4: Can we defined this type?

JeffreyMa597 (approved):

packages/neuron-ui/src/components/ApproveMultisigTxDialog/approveMultisigTx.module.scss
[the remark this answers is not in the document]
WhiteMinds: > It is necessary

As long as the existing value of left is not `!important` or an inline style, it is unnecessary.
The reason why the style did not take effect is simply because the selector's weight is not high enough.
There are various ways to change the weight, such as in this case, the original built-in selector is `.tipWithNode[data-has-trigger=true][data-placement=bottom]:not([data-trigger-next-to-child=true]) .tip`, which has a weight of `5 * 10 = 50`, while the current selector only has a weight of 30. Here are some examples of how to increase weight:

1. `.tip.tip.tip` - adding multiple class selectors

2. `.tip:not([a][b])` - adding invalid attribute selectors

Alternatively, you can decrease the weight of the previous selector:

Move `.tip` from `.tipWithNode` to the top level in `Tooltip/tooltip.module.scss`, so the weight will only be 10.

I would recommend using the second solution, as common components should be designed to make their styles easy to override.

I hope to never use `!important`, as it can break the maintainability of CSS.

packages/neuron-ui/src/components/ApproveMultisigTxDialog/approveMultisigTx.module.scss
```diff
@@ -0,0 +1,167 @@
...
+    padding: 0 16px;
+    font-size: 14px;
+    display: flex;
+    align-items: center;
+    background: var(--table-head-background-color);
+    color: var(--secondary-text-color);
+    border-bottom: 1px solid var(--table-head-border-color);
+
+    .tip {
+      &:not([data-tooltip='']) {
```
WhiteMinds, line 21:
```suggestion
      // This selector is only used to increase specificity and override the default styles in the Tooltip component.
      &:not([data-tooltip='']) {
```

WhiteMinds (changes requested):

WhiteMinds (approved):

PR #2669, opened 2023-05-15
Refactor Tabs

WhiteMinds:
Refactor Tabs

### DIFF
--- a/packages/neuron-ui/src/widgets/Tabs/index.tsx
+++ b/packages/neuron-ui/src/widgets/Tabs/index.tsx
@@ -1,32 +1,37 @@
-/* eslint-disable no-console */
-import React, { FC, useState } from 'react'
+import React, { ReactElement, ReactNode, useCallback, useState } from 'react'
+import { clsx } from 'utils'
 import { onEnter } from 'utils/inputDevice'
 
-export type Tab<T extends any = {}> = {
-  id: string | number
-  label?: string | React.ReactNode
-  render?: (tab: Tab<T>) => React.ReactNode
+type TabId = string | number
+
+export type Tab<T = {}> = {
+  id: TabId
+  label?: ReactNode
+  render?: (tab: Tab<T>) => ReactNode
 } & T
 
 export interface VariantProps<T> {
   tabs: Tab<T>[]
-  onTabChange: (id: string | number) => void
+  onTabChange: (tabId: TabId) => void
   selectedTab: Tab<T>
 }
 
-export type TabsProps<T extends any = {}> = Omit<VariantProps<T>, 'onTabChange' | 'selectedTab'> & {
-  Variant?: FC<VariantProps<T>>
-  onChange?: (changedTabValue: string | number) => void
-} & Partial<{ tabsClassName?: string; tabsWrapClassName?: string; tabsColumnClassName?: string }>
+type DefaultVariantProps = {
+  tabsClassName?: string
+  tabsWrapClassName?: string
+  tabsColumnClassName?: string
+  activeColumnClassName?: string
+}
 
-const DefaultVariant = <T extends any = {}>({
+export const DefaultVariant = <T extends {} = {}>({
   tabs,
   selectedTab,
   onTabChange,
   tabsClassName,
   tabsWrapClassName,
   tabsColumnClassName,
-}: VariantProps<T> & { tabsClassName?: string; tabsWrapClassName?: string; tabsColumnClassName?: string }) => {
+  activeColumnClassName,
+}: VariantProps<T> & DefaultVariantProps) => {
   return (
     <div className={tabsClassName}>
       <div className={tabsWrapClassName} role="tablist">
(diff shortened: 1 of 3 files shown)


packages/neuron-ui/src/widgets/Tabs/index.tsx
```diff
@@ -49,23 +53,41 @@ const DefaultVariant = <T extends any = {}>({
...
   const selectedTab = tabs.find(tab => tab.id === selectedTabId) ?? tabs[0]
 
-  const handleTabChange = (id: string | number) => {
-    setSelectedTabId(id)
-    onChange?.(id)
-  }
+  const combinedOnTabChange = useCallback<VariantProps<T>['onTabChange']>(
+    id => {
+      setSelectedTabId(id)
+      const selectedTab = tabs.find(tab => tab.id === id) ?? tabs[0]
```
JeffreyMa597, line 77: Same as above.

packages/neuron-ui/src/widgets/Tabs/index.tsx
```diff
@@ -49,23 +53,41 @@ const DefaultVariant = <T extends any = {}>({
...
+      onTabChange?.(selectedTab)
```
JeffreyMa597, line 78: Would it be better to just pass the id? Or we can pass the entire object through the second argument.

packages/neuron-ui/src/widgets/Tabs/index.tsx
```diff
@@ -49,23 +53,41 @@ const DefaultVariant = <T extends any = {}>({
...
+  const [selectedTabId, setSelectedTabId] = useState<TabId>(tabs[0].id)
   const selectedTab = tabs.find(tab => tab.id === selectedTabId) ?? tabs[0]
 
-  const handleTabChange = (id: string | number) => {
-    setSelectedTabId(id)
-    onChange?.(id)
-  }
+  const combinedOnTabChange = useCallback<VariantProps<T>['onTabChange']>(
+    id => {
+      setSelectedTabId(id)
```
yanguoyu, line 76: I think the `id` should be `selectedTab.id`, currently logic the `id` may not match `selectedTab`.

packages/neuron-ui/src/widgets/Tabs/index.tsx
```diff
@@ -37,8 +42,7 @@ const DefaultVariant = <T extends any = {}>({
             tabIndex={0}
             onKeyDown={onEnter(() => onTabChange(tab.id))}
             onClick={() => onTabChange(tab.id)}
-            className={tabsColumnClassName}
-            data-active={selectedTab.id === tab.id}
+            className={clsx(tabsColumnClassName, selectedTab.id === tab.id && activeColumnClassName)}
```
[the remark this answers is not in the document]
WhiteMinds: This requires confirming what the default style looks like, and currently, I'm not entirely sure about the default style of the tabs.
Replacing `data-active` with `className` was done to avoid implicit dependencies from external components on the internal implementation details of this component, thus preventing coupling.

packages/neuron-ui/src/widgets/Tabs/index.tsx
WhiteMinds: In the callback, it is possible to destructure `id`, so I think it should be fine.

packages/neuron-ui/src/widgets/Tabs/index.tsx
WhiteMinds: That makes sense, updated.

yanguoyu (approved):

packages/neuron-ui/src/widgets/Tabs/index.tsx
JeffreyMa597: > In the callback, it is possible to destructure `id`, so I think it should be fine.

I mean, in most situations, users only care about the id they get, but the entire object.

packages/neuron-ui/src/widgets/Tabs/index.tsx
```diff
@@ -49,23 +53,41 @@ const DefaultVariant = <T extends any = {}>({
...
+  onTabChange?: (currentTab: Tab<T>) => void
+  // The finer-grained type of E is lost during destructuring, so a manual exclusion operation is performed to address this.
+} & Omit<E, keyof VariantProps<T>>) => {
   if (tabs.length === 0) {
     throw new Error('Tabs must have at least one tab')
   }
 
-  const [selectedTabId, setSelectedTabId] = useState<string | number>(tabs[0].id)
+  const [selectedTabId, setSelectedTabId] = useState<TabId>(tabs[0].id)
   const selectedTab = tabs.find(tab => tab.id === selectedTabId) ?? tabs[0]
```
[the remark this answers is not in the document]
JeffreyMa597: Yes

packages/neuron-ui/src/widgets/Tabs/index.tsx
WhiteMinds: I'm not entirely sure if that's the case because regular users usually don't need to be concerned about tab changes.
When they focus on a specific tab, they may often require additional information from the tab. This would require writing an additional computation. 
I feel that the current implementation doesn't incur significant additional cost, which is why it was implemented this way.

packages/neuron-ui/src/widgets/Tabs/index.tsx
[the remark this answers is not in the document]
WhiteMinds: `find` may potentially find an empty value because users can pass in a dynamic tabs array.
Therefore, using `tabs[0]` ensures that `selectedTab` is always non-empty.

JeffreyMa597 (approved):

PR #2672, opened 2023-05-18
refactor: Optimize input and output storage

yanguoyu:
To search others' addresses at the history page, I create a table `tx_lock` to save `tx_hash` map `lockHash`

1. Only save local wallets' lock into `input` and `output`, and save others' cell lockHash to `tx_lock`
2. When searching by address, search addresses from `input`, `output` and `tx_lock`.
3. When getting transaction detail, get the transaction detail from RPC.
4. Record the last block number when caching tx to `indexer_tx_hash_cache`

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
@@ -4,6 +4,7 @@ import AddressMeta from '../../database/address/meta'
 import IndexerTxHashCache from '../../database/chain/entities/indexer-tx-hash-cache'
 import RpcService from '../../services/rpc-service'
 import TransactionWithStatus from '../../models/chain/transaction-with-status'
+import SyncInfoEntity from '../../database/chain/entities/sync-info'
 import { TransactionCollector, CellCollector, Indexer as CkbIndexer } from '@ckb-lumos/ckb-indexer'
 
 export default class IndexerCacheService {
@@ -11,6 +12,7 @@ export default class IndexerCacheService {
   private rpcService: RpcService
   private walletId: string
   private indexer: CkbIndexer
+  #cacheBlockNumberEntity?: SyncInfoEntity
 
   constructor(walletId: string, addressMetas: AddressMeta[], rpcService: RpcService, indexer: CkbIndexer) {
     for (const addressMeta of addressMetas) {
@@ -23,16 +25,6 @@ export default class IndexerCacheService {
     this.addressMetas = addressMetas
     this.rpcService = rpcService
     this.indexer = indexer
-  }
-
-  private async countTxHashes(): Promise<number> {
-    return getConnection()
-      .getRepository(IndexerTxHashCache)
-      .createQueryBuilder()
-      .where({
-        walletId: this.walletId,
-      })
-      .getCount()
   }
 
   private async getTxHashes(): Promise<IndexerTxHashCache[]> {
@@ -79,6 +71,8 @@ export default class IndexerCacheService {
   }
 
   private async fetchTxMapping(): Promise<Map<string, Array<{ address: string; lockHash: string }>>> {
+    const lastCacheBlockNumber = await this.getCachedBlockNumber()
+    const currentHeaderBlockNumber = await this.rpcService.getTipBlockNumber()
     const mappingsByTxHash = new Map()
     for (const addressMeta of this.addressMetas) {
       const lockScripts = [
@@ -88,9 +82,18 @@ export default class IndexerCacheService {
       ]
 
       for (const lockScript of lockScripts) {
-        const transactionCollector = new TransactionCollector(this.indexer, { lock: lockScript }, this.rpcService.url, {
-          includeStatus: false,
-        })
+        const transactionCollector = new TransactionCollector(
+          this.indexer,
+          {
+            lock: lockScript,
+            fromBlock: lastCacheBlockNumber.value,
+            toBlock: currentHeaderBlockNumber,
+          },
+          this.rpcService.url,
+          {
+            includeStatus: false,
+          }
+        )
 
         const fetchedTxHashes = await transactionCollector.getTransactionHashes()
         if (!fetchedTxHashes.length) {
@@ -126,6 +129,8 @@ export default class IndexerCacheService {
             args: lockScript.args.slice(0, 42),
           },
           argsLen,
+          fromBlock: lastCacheBlockNumber.value,
+          toBlock: currentHeaderBlockNumber,
         })
 
         for await (const cell of cellCollector.collect()) {
@@ -142,17 +147,33 @@ export default class IndexerCacheService {
     return mappingsByTxHash
   }
 
+  private async getCachedBlockNumber() {
+    if (!this.#cacheBlockNumberEntity) {
+      this.#cacheBlockNumberEntity = (await getConnection().getRepository(SyncInfoEntity).findOne({ name: SyncInfoEntity.getLastCachedKey(this.walletId) })) ?? 
+        SyncInfoEntity.fromObject({
+          name: SyncInfoEntity.getLastCachedKey(this.walletId),
+          value: '0x0'
+        })
+    }
+
+    return this.#cacheBlockNumberEntity
+  }
+
+  private async saveCacheBlockNumber(cacheBlockNumber: string) {
+    let cacheBlockNumberEntity = await this.getCachedBlockNumber()
+    cacheBlockNumberEntity.value = cacheBlockNumber
+    await getConnection().manager.save(cacheBlockNumberEntity)
+  }
+
   public async upsertTxHashes(): Promise<string[]> {
+    const tipBlockNumber = await this.rpcService.getTipBlockNumber()
     const mappingsByTxHash = await this.fetchTxMapping()
 
     const fetchedTxHashes = [...mappingsByTxHash.keys()]
-    const fetchedTxHashCount = fetchedTxHashes.reduce((sum, txHash) => sum + mappingsByTxHash.get(txHash)!.length, 0)
-
-    const txCount = await this.countTxHashes()
-    if (fetchedTxHashCount === txCount) {
+    if (!fetchedTxHashes.length) {
+      await this.saveCacheBlockNumber(tipBlockNumber)
       return []
     }
-
     const txMetasCaches = await this.getTxHashes()
     const cachedTxHashes = txMetasCaches.map(meta => meta.txHash.toString())
 
@@ -161,6 +182,7 @@ export default class IndexerCacheService {
     const newTxHashes = fetchedTxHashes.filter(hash => !cachedTxHashesSet.has(hash))
 
     if (!newTxHashes.length) {
+      await this.saveCacheBlockNumber(tipBlockNumber)
       return []
     }
 
@@ -208,6 +230,7 @@ export default class IndexerCacheService {
       }
     }
 
+    await this.saveCacheBlockNumber(tipBlockNumber)
     return newTxHashes
   }
 
--- a/packages/neuron-wallet/src/database/chain/entities/tx-lock.ts
+++ b/packages/neuron-wallet/src/database/chain/entities/tx-lock.ts
@@ -1,0 +1,24 @@
+import { BaseEntity, Entity, PrimaryColumn } from 'typeorm'
+
+@Entity()
+export default class TxLock extends BaseEntity {
+  @PrimaryColumn({
+    type: 'varchar'
+  })
+  transactionHash!: string
+
+  @PrimaryColumn({
+    type: 'varchar',
+  })
+  lockHash!: string
+  
+  static fromObject(obj: {
+    txHash: string
+    lockHash: string
+  }) {
+    const res = new TxLock()
+    res.transactionHash = obj.txHash
+    res.lockHash = obj.lockHash
+    return res
+  }
+}
--- a/packages/neuron-wallet/src/services/tx/transaction-persistor.ts
+++ b/packages/neuron-wallet/src/services/tx/transaction-persistor.ts
@@ -9,6 +9,8 @@ import OutPoint from '../../models/chain/out-point'
 import Output, { OutputStatus } from '../../models/chain/output'
 import Transaction, { TransactionStatus } from '../../models/chain/transaction'
 import Input from '../../models/chain/input'
+import TxLockEntity from '../../database/chain/entities/tx-lock'
+import { CHEQUE_ARGS_LENGTH, DEFAULT_ARGS_LENGTH } from '../../utils/const'
 
 export enum TxSaveType {
   Sent = 'sent',
@@ -42,7 +44,10 @@ export class TransactionPersistor {
   // After the tx is fetched:
   // 1. If the tx is not persisted before fetching, output = live, input = dead
   // 2. If the tx is already persisted before fetching, output = live, input = dead
-  private static saveWithFetch = async (transaction: Transaction): Promise<TransactionEntity> => {
+  private static saveWithFetch = async (
+    transaction: Transaction,
+    lockArgsSetNeedsDetail?: Set<string>
+  ): Promise<TransactionEntity> => {
     const connection = getConnection()
     const txEntity: TransactionEntity | undefined = await connection
       .getRepository(TransactionEntity)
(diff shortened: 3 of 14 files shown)


Keith-CY:
I'm not sure if I get the point.

Seems that all `inputs` and `outputs` are saved into `inputs`, `outputs`, `others_input`, and `others_output`.

It speeds searching up, but the storage is not optimized as mentioned at https://github.com/Magickbase/neuron-public-issues/issues/77#issue-1460921092

If a transaction includes many cells, they are still all saved into the database. The disk usage is not ameliorated.

yanguoyu:
> I'm not sure if I get the point.
> 
> Seems that all `inputs` and `outputs` are saved into `inputs`, `outputs`, `others_input`, and `others_output`.
> 
> It speeds searching up, but the storage is not optimized as mentioned at [Magickbase/neuron-public-issues#77 (comment)](https://github.com/Magickbase/neuron-public-issues/issues/77#issue-1460921092)
> 
> If a transaction includes many cells, they are still all saved into the database. The disk usage is not ameliorated.

Currently yes, I have not paid attention to the SQLite file's size, but only care about the search speed and SQL execution time. I want to use the data that has been got.

You remind me that we also pay attention to the SQLite file's size. Actually, I have tried another way to finish it, create a table that only saves `tx hash` and `lockHash` for searching addresses, and also works. Maybe it will reduce the disk cost by saving only two fields. I think saving two fields for searching addresses is cost-effective. But finally, for fewer RPC requests so I use currently schema.

For smaller SQLite'file size, there exist two schemas:
1. Create a new table to save `tx hash` and `lockHash` for searching addresses.
2. When searching others' addresses, get all transactions' detail in the current wallet and then filter with the address. I'm worry about the rpc's result is too big.

So, for reduce SQLite's file size, I suggestion for first solution. What do you think? @Keith-CY

Keith-CY:
> > I'm not sure if I get the point.
> > Seems that all `inputs` and `outputs` are saved into `inputs`, `outputs`, `others_input`, and `others_output`.
> > It speeds searching up, but the storage is not optimized as mentioned at [Magickbase/neuron-public-issues#77 (comment)](https://github.com/Magickbase/neuron-public-issues/issues/77#issue-1460921092)
> > If a transaction includes many cells, they are still all saved into the database. The disk usage is not ameliorated.
> 
> Currently yes, I have not paid attention to the SQLite file's size, but only care about the search speed and SQL execution time. I want to use the data that has been got.
> 
> You remind me that we also pay attention to the SQLite file's size. Actually, I have tried another way to finish it, create a table that only saves `tx hash` and `lockHash` for searching addresses, and also works. Maybe it will reduce the disk cost by saving only two fields. I think saving two fields for searching addresses is cost-effective. But finally, for fewer RPC requests so I use currently schema.
> 
> For smaller SQLite'file size, there exist two schemas:
> 
> 1. Create a new table to save `tx hash` and `lockHash` for searching addresses.
> 2. When searching others' addresses, get all transactions' detail in the current wallet and then filter with the address. I'm worry about the rpc's result is too big.
> 
> So, for reduce SQLite's file size, I suggestion for first solution. What do you think? @Keith-CY

Solution 2 has an alternative as mentioned at https://github.com/Magickbase/neuron-public-issues/issues/77#issuecomment-1480693839

yanguoyu:
> Solution 2 has an alternative as mentioned at [Magickbase/neuron-public-issues#77 (comment)](https://github.com/Magickbase/neuron-public-issues/issues/77#issuecomment-1480693839)

I'm worried about search one address that may belong to a miner.

Keith-CY:
> > Solution 2 has an alternative as mentioned at [Magickbase/neuron-public-issues#77 (comment)](https://github.com/Magickbase/neuron-public-issues/issues/77#issuecomment-1480693839)
> 
> I'm worried about search one address that may belong to a miner.

Indexer supports pagination so the count of transactions won't bother. But the pagination of indexer is cursor-style, while neuron uses offset-style it, it's a problem if we use indexer to list transactions.

So I think solution 1 is good for our case now

yanguoyu:
I will reopen the PR after I changed it to Solution 1

yanguoyu:
> I will reopen the PR after I changed it to Solution 1

It's ready for review now.

packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
```diff
@@ -153,17 +151,36 @@ export default class IndexerCacheService {
     return mappingsByTxHash
   }
 
+  private async getCacheBlockNumber() {
```
Keith-CY, line 154: How about naming it `getCachedBlockNumber` or `getBlockNumberCache`

packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
```diff
@@ -153,17 +151,36 @@ export default class IndexerCacheService {
...
+    if (!this.#cacheBlockNumberEntity) {
+      let lastCacheBlockNumber = await getConnection().getRepository(SyncInfoEntity).findOne({ name: `${SyncInfoEntity.LAST_CACHE_BLOCK_NUMBER}_${this.walletId}` })
+
+      if (!lastCacheBlockNumber) {
+        lastCacheBlockNumber = new SyncInfoEntity()
+        lastCacheBlockNumber.name = `${SyncInfoEntity.LAST_CACHE_BLOCK_NUMBER}_${this.walletId}`
```
Keith-CY, line 160: A method of getting the key can be defined in `SyncInfoEntity` so it won't be constructed repeatedly.

And the initialization can be defined in `SyncInfoEntity` too.

packages/neuron-wallet/src/block-sync-renderer/sync/indexer-cache-service.ts
```diff
@@ -153,17 +151,36 @@ export default class IndexerCacheService {
...
+  }
+
+  private async saveCacheBlockNumber(cacheBlockNumber: string) {
+    let cacheBlockNumberEntity = await this.getCacheBlockNumber()
+    cacheBlockNumberEntity.value = cacheBlockNumber
+    await getConnection().manager.save(cacheBlockNumberEntity)
+  }
+
   public async upsertTxHashes(): Promise<string[]> {
+    const currentHeaderBlockNumber = await this.rpcService.getTipBlockNumber()
```
Keith-CY, line 176: Why not use `tipBlockNumber` which is defined clearly

packages/neuron-wallet/src/database/chain/entities/tx-lock.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseEntity, Entity, PrimaryColumn } from 'typeorm'
+
+@Entity()
+export default class TxLock extends BaseEntity {
```
Keith-CY, line 4: `Lock` is mentioned in the name but `lock hash` is its field, which is confusing.

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -373,6 +387,16 @@ export class TransactionPersistor {
     const txEntity: TransactionEntity = await TransactionPersistor.convertTransactionAndSave(tx, TxSaveType.Sent)
     return txEntity
   }
+
+  private static isCurrentWalletCell(cell: InputEntity | OutputEntity, lockArgsSet: Set<string>) {
+    return (cell.lockArgs && (
+      lockArgsSet.has(cell.lockArgs)
+      || (
+        cell.lockArgs.length === 82
```
Keith-CY, line 395: `82` and `42` are a magic numbers

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -76,32 +78,27 @@ export class TransactionsService {
...
+        .query(
+          `
             SELECT transactionHash from (
-              SELECT output.transactionHash FROM output WHERE output.lockHash in (:...lockHashes)
+              SELECT output.transactionHash FROM output WHERE output.lockHash = @0
               UNION
-              SELECT input.transactionHash FROM input WHERE input.lockHash in (:...lockHashes)
+              SELECT input.transactionHash FROM input WHERE input.lockHash = @0
+              UNION
+              SELECT tx_lock.transactionHash FROM tx_lock WHERE tx_lock.lockHash = @0
```
Keith-CY, line 90: How about separating this query into 2 types

```sql
SELECT output.transactionHash from output WHERE output.lockHash = @0
UNION
SELECT input.transactionHash from input WHERE input.lockHash = @0
```
and
```sql
SELECT tx_lock.transactionHash from tx_lock WHERE tx_lock.lockHash = @0
```

And add a condition of `lockHash in current wallet` before this query to boost it.

`tx_lock` could be huge to introduce performance impact while it should not be queried so often as `current wallet`

By doing this, the `intersect` could be avoided too.

yanguoyu: `lockHash` is an index of `tx_lock` and `SELECT tx_lock.transactionHash from tx_lock WHERE tx_lock.lockHash = @` is filtered by whether `lockHash` is equal, so it should execute fast.

packages/neuron-wallet/src/database/chain/entities/tx-lock.ts
yanguoyu: I have no good idea for the table name, do you have any suggestions?
I thought if the table name includes the field name, it will change to `txhash_lockhash`, but it looks strange. So I simply name it `tx_lock`.

Keith-CY: Maybe `tx_lock` is good enough if `tx hash` and `lock hash` are treated as `id`

Keith-CY (approved):

Keith-CY:
Please have a review @WhiteMinds @devchenyan @JeffreyMa597

JeffreyMa597 (approved):

JeffreyMa597:
please resolve the conflicts.

yanguoyu:
@FrederLu Please test it with the package.

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -297,22 +300,32 @@ export class TransactionPersistor {
...
+    let txLocks: TxLockEntity[] = []
+    if (lockArgsSet?.size) {
+      currentWalletInputs = inputs.filter(v => this.isCurrentWalletCell(v, lockArgsSet))
+      currentWalletOutputs = outputs.filter(v => this.isCurrentWalletCell(v, lockArgsSet))
+      txLocks = [...new Set(
+        [
+          ...inputs.filter(v => v.lockHash && !this.isCurrentWalletCell(v, lockArgsSet))
+                  .map(v => v.lockHash!),
+          ...outputs.filter(v => !this.isCurrentWalletCell(v, lockArgsSet))
+                  .map(v => v.lockHash),
```
WhiteMinds, line 314: Perhaps using `!currentWalletInputs.includes(v)` to replace `!this.isCurrentWalletCell(v, lockArgsSet)` might be more readable and easier to maintain.
Just my personal thoughts, it's optional.

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -485,29 +482,45 @@ export class TransactionsService {
...
+    const inputTxs = await ckb.rpc
+      .createBatchRequest<'getTransaction', string[], CKBComponents.TransactionWithStatus[]>(
+        inputTxHashes.map(v => ['getTransaction', v])
+      )
+      .exec()
+    const inputTxMap = new Map<string, CKBComponents.Transaction>()
+    inputTxs.forEach((v, idx) => {
+      inputTxMap.set(inputTxHashes[idx], v.transaction)
+    })
+    txWithStatus.transaction.inputs.forEach(v => {
```
WhiteMinds, line 511: The operation performed by this for loop leaves me a bit confused, so it might lack comments, or it should be extracted into a separate function.

WhiteMinds (changes requested):

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
```diff
@@ -342,25 +355,27 @@ export class TransactionPersistor {
...
-      txEntity = await TransactionPersistor.saveWithFetch(tx)
+      txEntity = await TransactionPersistor.saveWithFetch(tx, lockArgsSet)
     } else {
       throw new Error('Error TxSaveType!')
     }
     return txEntity
   }
 
-  public static saveFetchTx = async (transaction: Transaction): Promise<TransactionEntity> => {
+  public static saveFetchTx = async (transaction: Transaction, lockArgsSet?: Set<string>): Promise<TransactionEntity> => {
```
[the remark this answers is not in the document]
yanguoyu: For this function, it does not care whether the `lockArgsSet` is owned by itself.

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
yanguoyu: I can't get the point. Where is the `currentWalletInputs ` from?

WhiteMinds: It is in the lines of code not shown above:
```currentWalletInputs = inputs.filter(v => this.isCurrentWalletCell(v, lockArgsSet))```

Comment (yanguoyu) [reply packages/neuron-wallet/src/services/tx/transaction-persistor.ts:314]: `currentWalletInputs` is the filter result, add a filter `!currentWalletInputs.includes(v)` to filter which?

File: packages/neuron-wallet/src/services/tx/transaction-persistor.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (WhiteMinds) [reply packages/neuron-wallet/src/services/tx/transaction-persistor.ts:374]: The call chain is as follows:
```
saveFetchTx -> TransactionPersistor.convertTransactionAndSave ->
saveWithFetch -> TransactionPersistor.create(..., lockArgsSet)
```
The parameter is used in `TransactionPersistor.create` as follows:
```
currentWalletInputs = inputs.filter(v => this.isCurrentWalletCell(v, lockArgsSet))
```
For the `isCurrentWalletCell(cell: InputEntity | OutputEntity, lockArgsSet: Set<string>)` function, the parameter should actually be `currentWalletLockArgsSet` (also known as `selfLockArgsSet`).

Therefore, the parameter required by `TransactionPersistor.create` should be `selfLockArgsSet` or an abstraction that implements `isSelfInputOrOutput`.

Following the call chain upwards, if the upper layers do not have branch checks such as `isSelf(lockArgsSet)`, then the meaning of `lockArgsSet` must be `selfLockArgsSet`.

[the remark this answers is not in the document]
Comment (yanguoyu) [reply packages/neuron-wallet/src/services/tx/transaction-persistor.ts:374]: Maybe I need to rename `currentWalletInputs`? The `lockArgsSet` is from `queue`'s constructor's `addresses`, but the `addresses` is not named `selfAddresses`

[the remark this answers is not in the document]
Comment (WhiteMinds) [reply packages/neuron-wallet/src/services/tx/transaction-persistor.ts:374]: I understand now.

In this call chain (at least in `TransactionPersistor.create`), its meaning could be something like "lockArgsSet for which details need to be saved".

It is a broader abstraction than what `selfLockArgsSet` represents, so perhaps it could be named `saveDetailLockArgsSet`.

The names `isCurrentWalletCell` and `currentWalletInputs` do indeed need adjustment.

File: packages/neuron-wallet/src/services/tx/transaction-persistor.ts (hunk 7 above)
Comment (WhiteMinds) [reply packages/neuron-wallet/src/services/tx/transaction-persistor.ts:314]: like this:
```
txLocks = [...new Set(
  [
    ...inputs.filter(v => v.lockHash && !currentWalletInputs.includes(v))
            .map(v => v.lockHash!),
    ...outputs.filter(v => !currentWalletOutputs.includes(v))
            .map(v => v.lockHash),
  ]
```

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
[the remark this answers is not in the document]
yanguoyu: Sounds good.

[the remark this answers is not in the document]
yanguoyu: How about `lockArgsSetNeedsDetail`, and `saveDetailLockArgsSet` like a verb that we will save detail?

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
yanguoyu: In my opinion, `...inputs.filter(v => v.lockHash && !this.isCurrentWalletCell(v, lockArgsSet))` is more clearly and fast.
And it does not need to depend on `currentWalletInputs`

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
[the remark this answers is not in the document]
WhiteMinds: > How about `lockArgsSetNeedsDetail`

LGTM

packages/neuron-wallet/src/services/tx/transaction-persistor.ts
WhiteMinds: Ok no problem, it just seems to me that it might improve readability

Keith-CY:
Please resolve the conflicts and squash merge @yanguoyu

PR #2674, opened 2023-05-19
feat: improve the interaction design of the DatetimePicker component

WhiteMinds:
feat: improve the interaction design of the DatetimePicker component

Ref: https://github.com/Magickbase/neuron-public-issues/issues/151 

https://github.com/nervosnetwork/neuron/assets/9160743/a6775127-a1df-437a-b5f9-d6770e466d97

### DIFF
--- a/packages/neuron-ui/src/widgets/Calendar/index.tsx
+++ b/packages/neuron-ui/src/widgets/Calendar/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useMemo, useCallback } from 'react'
+import React, { useState, useMemo, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import {
   getMonthCalendar,
@@ -71,11 +71,6 @@ const Calendar: React.FC<CalendarProps> = ({
   const [month, setMonth] = useState(new Date().getMonth() + 1)
   const [status, setStatus] = useState<'year' | 'month' | 'date'>('date')
 
-  useEffect(() => {
-    setYear(value?.getFullYear() ?? new Date().getFullYear())
-    setMonth((value?.getMonth() ?? new Date().getMonth()) + 1)
-  }, [value?.toDateString()])
-
   const [uId] = useState(() => (+new Date()).toString(16).slice(-4))
 
   const [t, { language }] = useTranslation()
@@ -87,12 +82,10 @@ const Calendar: React.FC<CalendarProps> = ({
   const monthName = monthNames[month - 1]
   const monthShortName = monthShortNames[month - 1]
 
-  const calendar = useMemo(() => getMonthCalendar(year, month, firstDayOfWeek, language), [
-    year,
-    month,
-    firstDayOfWeek,
-    language,
-  ])
+  const calendar = useMemo(
+    () => getMonthCalendar(year, month, firstDayOfWeek, language),
+    [year, month, firstDayOfWeek, language]
+  )
   function isDisabledTime(date: Date): boolean {
     return !isDayInRange(date, { minDate, maxDate })
   }
--- a/packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
+++ b/packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
@@ -4,7 +4,6 @@ import Button from 'widgets/Button'
 import { useTranslation } from 'react-i18next'
 import styles from './datetimePicker.module.scss'
 
-const SECONDS_PER_DAY = 24 * 3600 * 1000
 let UTC: string | number = -new Date().getTimezoneOffset() / 60
 if (UTC > 0) {
   UTC = `UTC+${UTC}`
@@ -14,10 +13,7 @@ if (UTC > 0) {
 
 export const formatDate = (datetime: Date) => {
   const month = (datetime.getMonth() + 1).toString().padStart(2, '0')
-  const date = datetime
-    .getDate()
-    .toString()
-    .padStart(2, '0')
+  const date = datetime.getDate().toString().padStart(2, '0')
   const year = datetime.getFullYear()
   return `${month}/${date}/${year}`
 }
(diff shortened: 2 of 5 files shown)


Keith-CY:
Please review the interaction @Danie0918, @Sven-TBD

packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
```diff
@@ -49,7 +40,9 @@ const DatetimePicker = ({
   const tomorrow = new Date()
   tomorrow.setDate(tomorrow.getDate() + 1)
   const isSinceTomorrow = display ? new Date(display).getTime() >= new Date(tomorrow.toDateString()).getTime() : false
-  const disabled = !selected || !isSinceTomorrow
+  const isDayEqualsForPreset =
+    preset != null && selected != null && new Date(+preset).toDateString() === selected.toDateString()
+  const disabled = !selected || !isSinceTomorrow || isDayEqualsForPreset
```
Keith-CY, line 45: Why does `isDayEqualsForPreset` disable `save`

Danie0918:
Note that when no date is selected, the save button is unavailable instead of hidden. In this case, it is recommended to unify the copywriting as "locked to".

packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
WhiteMinds: Oh, this is to implement `We should have the 'Save' button initially disabled until the user manually makes a change.`. I'm not sure if we still need this after displaying the locked date on the save button. @Danie0918 , what do you think?

Danie0918: It depends on whether our default date is a time commonly used by users, and if so, we can provide options. Conversely, if the user does not have a universal lock time, it is best not to allow saving without manual selection to avoid misuse.

yanguoyu (approved):

packages/neuron-ui/src/widgets/DatetimePicker/index.tsx
WhiteMinds: After discussing with @Danie0918 , it was decided to remove this condition to avoid confusion when the user selects a different date and then reverts it, resulting in the button being disabled.

WhiteMinds:
> Note that when no date is selected, the save button is unavailable instead of hidden. In this case, it is recommended to unify the copywriting as "locked to".

Resolved

WhiteMinds:
This is the result after making changes based on the comments in the PR:

https://github.com/nervosnetwork/neuron/assets/9160743/3f403130-668a-46fe-a83d-58fa50860161


Additionally, for the translation key, I chose to use `release-on`, but it may not align with the initial `locked to` mentioned in the product:

image

@Danie0918 Should I specify a new key, or should we continue using "release-on"?

packages/neuron-ui/src/widgets/Calendar/index.tsx
```diff
@@ -67,8 +67,8 @@ const Calendar: React.FC<CalendarProps> = ({
   firstDayOfWeek = 0,
   className = '',
 }) => {
-  const [year, setYear] = useState(new Date().getFullYear())
-  const [month, setMonth] = useState(new Date().getMonth() + 1)
+  const [year, setYear] = useState(value?.getFullYear() ?? new Date().getFullYear())
+  const [month, setMonth] = useState((value?.getMonth() ?? new Date().getMonth()) + 1)
```
Keith-CY, line 71: Should be `value?.getFullYear` and `value?.getMonth`?

WhiteMinds: This is to address a previously existing bug where the panel did not automatically navigate to the correct position when a date attribute was passed.
After giving it some thought, I have come up with a more appropriate fix.
PR updated.

JeffreyMa597 (approved):

Keith-CY (approved):

Keith-CY:
Please have a review @FrederLu @silySuper @littleLip520

FrederLu:
@WhiteMinds 
Testers need to verify the PR through the test package. This PR is initiated from your own Repo, so the packaging process cannot be triggered.
It needs to be initiated from the Repo under nervosnetwork to trigger the packaging process because a certificate is required.
Please re-operate the pr submission process.
Status quo
image

---
Desired state
image
image
image

WhiteMinds:
@Keith-CY I believe it's quite common for the source branch of a PR to not be on the main repository. Considering that, we should modify the workflow. Here are a few options:

1. use `fail-fast: false`

1. Check if it is the main repository before executing `Package for MacOS`, and automatically skip if it is not

1. Check for the presence of necessary secrets before executing `Package for MacOS`, and output a warning message if they are missing

1. Automatically switch to non-signing mode and output a warning when the required signing secrets are missing

@FrederLu I have just pushed the branch to the main repository. You can now access the packaging result through this link: https://github.com/nervosnetwork/neuron/commit/a9eea2ee6b299ea272c7f6c8009df52971c906d3#commitcomment-114552831

Keith-CY:
> @Keith-CY I believe it's quite common for the source branch of a PR to not be on the main repository. Considering that, we should modify the workflow. Here are a few options:
> 
> 1. use `fail-fast: false`
> 2. Check if it is the main repository before executing `Package for MacOS`, and automatically skip if it is not
> 3. Check for the presence of necessary secrets before executing `Package for MacOS`, and output a warning message if they are missing
> 4. Automatically switch to non-signing mode and output a warning when the required signing secrets are missing
> 
> @FrederLu I have just pushed the branch to the main repository. You can now access the packaging result through this link: [a9eea2e#commitcomment-114552831](https://github.com/nervosnetwork/neuron/commit/a9eea2ee6b299ea272c7f6c8009df52971c906d3#commitcomment-114552831)

I'll open an issue to improve the workflow of **packaging for test**

FrederLu:
> @Keith-CY I believe it's quite common for the source branch of a PR to not be on the main repository. Considering that, we should modify the workflow. Here are a few options:
> 
> 1. use `fail-fast: false`
> 2. Check if it is the main repository before executing `Package for MacOS`, and automatically skip if it is not
> 3. Check for the presence of necessary secrets before executing `Package for MacOS`, and output a warning message if they are missing
> 4. Automatically switch to non-signing mode and output a warning when the required signing secrets are missing
> 
> @FrederLu I have just pushed the branch to the main repository. You can now access the packaging result through this link: [a9eea2e#commitcomment-114552831](https://github.com/nervosnetwork/neuron/commit/a9eea2ee6b299ea272c7f6c8009df52971c906d3#commitcomment-114552831)

Thanks, the test package has been downloaded through the portal.

PR #2675, opened 2023-05-19
fix: Fix some bugs.

yanguoyu:
1. When the light client just started, some dep cells should wait for a moment to get. So if the current light client can not get the dep cells, wait for 10 seconds and try it again.
2. If process. stdout is `pipe`, it should add event listener, otherwise when call process. kill, we need to destroy the stdout
3. If we clean data without cleaning the indexer data, we should not restart the ckb node.
4. When we clean data, restart the sync task to avoid the check node skipping.

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
@@ -9,7 +9,7 @@ import { scheduler } from 'timers/promises'
 import SyncProgressService from '../../services/sync-progress'
 import { BlockTips, LumosCellQuery, Connector, AppendScript } from './connector'
 import { scriptToHash } from '@nervosnetwork/ckb-sdk-utils'
-import { LightRPC, LightScriptFilter } from '../../utils/ckb-rpc'
+import { FetchTransactionReturnType, LightRPC, LightScriptFilter } from '../../utils/ckb-rpc'
 import HexUtils from '../../utils/hex'
 import Multisig from '../../services/multisig'
 import { SyncAddressType } from '../../database/chain/entities/sync-progress'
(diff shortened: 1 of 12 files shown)


packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -75,30 +75,36 @@ export default class LightConnector extends Connector<CKBComponents.Hash> {
...
-      .createBatchRequest<any, string[], (CKBComponents.TransactionWithStatus | null)[]>(fetchTxHashes)
+    const txs = await this.lightRpc
+      .createBatchRequest<any, string[], (CKBComponents.TransactionWithStatus | { status: 'fetching' | 'added' | 'not_found' })[]>(fetchTxHashes)
       .exec()
-    if (txs.some(v => !v)) {
-      // if some txs fetch to added, then fetch the actural txs
-      txs = await this.lightRpc
-        .createBatchRequest<any, string[], (CKBComponents.TransactionWithStatus | null)[]>(fetchTxHashes)
-        .exec()
+    if (txs.some(v => 'status' in v && !!v.status)) {
```
Keith-CY, line 81: In which case, the `v.status` is `undefined` while the field exists

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -75,30 +75,36 @@ export default class LightConnector extends Connector<CKBComponents.Hash> {
...
+      // wait for light client sync the dep cell
+      await scheduler.wait(10000)
+      return await this.getDepTxs()
     }
-    const depGroupOutputsData: string[] = fetchCellDeps
+    return fetchCellDeps
       .map((v, idx) => {
         if (v.depType === DepType.DepGroup) {
```
Keith-CY, line 88: `fetchCellDeps` can be filtered by `v.depType === DepType.DepGroup` before fetching transactions to reduce the payload of batch PRC and avoid this `if statement` if `map`

yanguoyu: It can not reduce the payload of batch RPC. If the cellDep's `depType` is not equal `DepType.DepGroup`, it also needs to fetch the cellDep's outPoint tx. And only if the cellDep's `depType` is equal `DepType.DepGroup`, it needs to fetch its dep group cell.

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
yanguoyu: There is no case for that. I change the return type of `fetchTransaction` by https://github.com/nervosnetwork/neuron/pull/2675/commits/25a753daf8e9a5a76e2eeb0aebaacc628c72eab2 and I think I should keep all the status of `fetch_transaction`'s results. Then it will be more clear.

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
Keith-CY: I see

Keith-CY (approved):

devchenyan (approved):

PR #2679, opened 2023-05-23
fix: Fix dao progress bar and light client should not verify version and indexer

yanguoyu:


packages/neuron-wallet/src/services/node.ts
```diff
@@ -143,8 +143,11 @@ class NodeService {
     } else {
       logger.info('CKB:\texternal RPC on default uri detected, skip starting bundled CKB node.')
       this._isCkbNodeExternal = true
-      await this.verifyNodeVersion()
-      await this.verifyStartWithIndexer()
+      let network = NetworksService.getInstance().getCurrent()
```
Keith-CY, line 146:
```suggestion
      const network = NetworksService.getInstance().getCurrent()
```

packages/neuron-ui/src/components/NervosDAO/hooks.ts
```diff
@@ -565,16 +565,19 @@ export const useUpdateDepositEpochList = ({
         depositBlockHashes => {
           const recordKeyIdxMap = new Map<string, number>()
           const batchParams: ['getHeader', string][] = []
-          records.forEach((record, idx) => {
+          let index = 0
+          records.forEach((record) => {
             if (!record.depositOutPoint && record.blockHash) {
               batchParams.push(['getHeader', record.blockHash])
-              recordKeyIdxMap.set(record.outPoint.txHash, idx)
+              recordKeyIdxMap.set(record.outPoint.txHash, index)
```
Keith-CY, line 572: If the elements of `recordKeyIdxMap` comes from 2 arrays and the index grows 1 by 1, it could be an array.

yanguoyu: It is used to find the index of the record. Because the records key's order is different with batchRequest.

Keith-CY: Personal feeling is that

```javascript
let index = 0
const recordKeyIdxMap = new Map()
items.forEach(item => {
  if (randomTrue()) {
    recordKeyIdxMap.set(item, index)
    index += 1
  }
}
```

is hard to understand its purpose, but

```javascript
const recordKeyIdx = []
items.forEach(item => {
  if (randomTrue()) {
    recordKeyIdx.push(item)
  }
}
```

is clear

yanguoyu: It needs a loop to get the item's index when I know the item's value.
But for Map, it can easily get the index by `Map.get`.

Keith-CY: > It needs a loop to get the item's index when I know the item's value. But for Map, it can easily get the index by `Map.get`.

Does `Array.prototype.indexOf` help?

yanguoyu: I think `Array.prototype.indexOf` is also using a loop to find the index of `array`. Actually, because the length of array will not be large, I will change it to this for more clarity.

Keith-CY (approved):

devchenyan (approved):

PR #2682, opened 2023-05-25
fix: Fix multisig with light client

yanguoyu:
1. Any multisig script should sync from 0.
2. When adding multisig address after deletion, it should start syncing again.
3. Every 10s to refresh multisig sync height, not when multisig address changes.

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
@@ -156,8 +156,8 @@ export default class LightConnector extends Connector<CKBComponents.Hash> {
     })
   }
 
-  private async initSyncProgress(appendScripts?: AppendScript[]) {
-    if (!this.addressMetas.length && !appendScripts?.length) {
+  private async initSyncProgress(appendScripts: AppendScript[] = []) {
+    if (!this.addressMetas.length && !appendScripts.length) {
       return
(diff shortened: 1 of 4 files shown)


packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -181,27 +181,35 @@ export default class LightConnector extends Connector<CKBComponents.Hash> {
...
+        blockNumber:
+          existSyncscripts[scriptToHash(v.script)]?.blockNumber ??
+          `0x${(otherTypeSyncProgress[scriptToHash(v.script)] ?? 0).toString(16)}`
+      }))
+    ]
     await this.lightRpc.setScripts(setScriptsParams)
-    await SyncProgressService.resetSyncProgress(allScripts)
     const walletIds = [...new Set(this.addressMetas.map(v => v.walletId))]
-    await SyncProgressService.removeWalletsByExists(walletIds)
+    await SyncProgressService.resetSyncProgress([...allScripts, ...appendScripts])
```
Keith-CY, line 208: Combination of arrays with spreading is used here and somewhere, the readability could be improved by `[allScripts, appendScripts].flat()`

Just a suggestion for code style

Keith-CY (approved):

JeffreyMa597 (approved):

PR #2686, opened 2023-05-26
fix: download x64 portable and fix running neuron locally

yanguoyu:
1. Some arm64 can not run not portable x64 ckb
2. When starting neuron to develop, the ckb file should add arch.


Keith-CY:
Maybe aarch64 should be portable too.
Ref: https://github.com/nervosnetwork/neuron/pull/2686/files#diff-cc3c5c8da7c89e637147feebb27beae48a7a55591e85f833ddea9395ff1592d8R26

Keith-CY (approved):

devchenyan (approved):

PR #2683, opened 2023-05-25
refactor: upgrade lumos to the 0.20.x

homura:
This PR resolve https://github.com/Magickbase/neuron-public-issues/issues/140

- upgrading lumos to 0.20.x
- replace `@nervina-labs/ckb-indexer` with `@ckb-lumos/ckb-indexer`

PR #2692, opened 2023-05-30
chore: enable prettier check in ci

homura:
Keeping the same code style will make the project more predictable, especially when finding some code in the whole project.

To keep the code in the same style, we can 

- A. add the `"prettier/prettier": "error"` in the eslint config, and run eslint in the CI
- B. run prettier when we commit, and run `prettier && git diff --exit-code` in the CI
- C. or something others

This PR using the option A as the solution, but this solution may not fit the habits of most Neuron contributors, so I need some feedback to continue this PR

## TODO

- [ ] https://github.com/nervosnetwork/neuron/pull/2683

### DIFF
--- a/.github/workflows/check-code-style.yml
+++ b/.github/workflows/check-code-style.yml
@@ -1,0 +1,38 @@
+name: Check Code Style
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  default:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          cache: "yarn"
+
+      - name: Restore
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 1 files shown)


homura:
If you prefer adding the `"prettier/prettier": "error"` in the eslint config, and run eslint in the CI, vote up for me

homura:
If you prefer running prettier when we commit, and run `prettier && git diff --exit-code` in the CI, vote up for me

Keith-CY:
How is it going?

homura:
According to the votes, I changed to check the code style in CI with `prettier --check`, and here is a [demo](https://github.com/homura/neuron/actions/runs/5409379433/jobs/9829446894#step:7:8) to show how the workflow works

yanguoyu (approved):

Keith-CY (approved):

JeffreyMa597 (approved):

WhiteMinds (approved):

PR #2693, opened 2023-05-31
feat: Add package for pull request and comment

yanguoyu:
Because the trigger on `workflow_run` needs the action's yml file to exist on the default branch.
https://github.com/yanguoyu/neuron/pull/38#issuecomment-1569914593 is an example of a comment package for committing to PR.

1. Add pull_request trigger for the `Package Neuron for Test` action
2. Skip the code sign if there not exist certificate
3. Because for pull_request action, there is no access to write comments. Add a workflow to add comments on the PR. https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#restrictions-on-repository-forks. Then when the trigger on push event, add a comment on the commit, else add a comment on the PR.

### DIFF
--- a/.github/workflows/comment_for_pr_package.yml
+++ b/.github/workflows/comment_for_pr_package.yml
@@ -1,0 +1,43 @@
+name: Comment on the pull request
+
+on:
+  workflow_run:
+    workflows: ["Package Neuron for Test"]
+    types:
+      - completed
+
+jobs:
+  comment:
+    name: Append links to the Pull Request
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - id: download_artifact
+        uses: dawidd6/action-download-artifact@v2
+        continue-on-error: true
+        with:
+          name: package_info
+          run_id: ${{ github.event.workflow_run.id }}
+
+      - name: if non exist, skip comment
+        if: ${{ steps.download_artifact.outcome == 'failure' }}
+        run: exit 0
+
+      - id: pr_number
+        run: |
+          export PR_NUM=$(cat ./pr_number)
+          echo "pr_num=$PR_NUM" >> $GITHUB_OUTPUT
+      - id: run_id
+        run: |
+          export RUN_ID=$(cat ./run_id)
+          echo "run_id=$RUN_ID" >> $GITHUB_OUTPUT
+      - id: sha
+        run: |
+          export SHA=$(cat ./sha)
+          echo "sha=$SHA" >> $GITHUB_OUTPUT
+      - uses: thollander/actions-comment-pull-request@v2
+        if: ${{ steps.pr_number.outputs.pr_num != '' && steps.run_id.outputs.run_id != '' }}
+        with:
+          message: |
+            Packaging for test is done in [${{ steps.run_id.outputs.run_id }}](https://github.com/${{github.repository}}/actions/runs/${{steps.run_id.outputs.run_id}}) for commit https://github.com/${{github.repository}}/pull/${{steps.pr_number.outputs.pr_num}}/commits/${{steps.sha.outputs.sha}} .
+          pr_number: ${{ steps.pr_number.outputs.pr_num }}
--- a/.github/workflows/package_for_test.yml
+++ b/.github/workflows/package_for_test.yml
@@ -1,6 +1,8 @@
 name: Package Neuron for Test
 
-on: [push]
+on: 
+  pull_request:
+  push:
 
 jobs:
   packaging:
@@ -16,6 +18,9 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     name: ${{ matrix.os }}(Node.js ${{ matrix.node }})
+    env:
+      APPLE_ID: ${{ secrets.APPLE_ID }}
+      WIN_CERTIFICATE_BASE64: ${{ secrets.WIN_CERTIFICATE_BASE64 }}
 
     steps:
       - name: Set git to use LF
@@ -62,8 +67,9 @@ jobs:
           CI: false
 
       - name: Package for MacOS
-        if: matrix.os == 'macos-latest'
+        if: ${{ matrix.os == 'macos-latest' && env.APPLE_ID != '' }}
         run: |
+          export CSC_FOR_PULL_REQUEST=true
           ./scripts/download-ckb.sh mac
           yarn package:test mac
         env:
@@ -73,8 +79,19 @@ jobs:
           CSC_LINK: ${{ secrets.MAC_CERTIFICATE_BASE64 }}
           CSC_KEY_PASSWORD: ${{ secrets.MAC_CERTIFICATE_PASSWORD }}
 
+      - name: Package for MacOS for skip code sign
+        if: ${{ matrix.os == 'macos-latest' && env.APPLE_ID == '' }}
+        run: |
+          export CSC_IDENTITY_AUTO_DISCOVERY=false
+          unset APPLE_ID
+          ./scripts/download-ckb.sh mac
+          yarn package:test mac
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SKIP_NOTARIZE: true
+
       - name: Package for Windows
-        if: matrix.os == 'windows-2019'
+        if: ${{ matrix.os == 'windows-2019' && env.WIN_CERTIFICATE_BASE64 != '' }}
         run: |
           bash ./scripts/download-ckb.sh win
           yarn build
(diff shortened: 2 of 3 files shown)


Keith-CY:
> Because the trigger on `workflow_run` needs the action's yml file to exist on the default branch. [yanguoyu#38 (comment)](https://github.com/yanguoyu/neuron/pull/38#issuecomment-1569914593) is an example of a comment package for committing to PR.
> 
> 1. Add pull_request trigger for the `Package Neuron for Test` action
> 2. When the action is triggered by `pull_request` event, packaging without code sign.
> 3. After package success, comment on the PR with the package result.

It requires an extra action to sign a package, what if sign packages when the PR author is permitted, as we did in **add-replied-label**(https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/add-replied-label.yml#L13-L29)

Keith-CY:
> > Because the trigger on `workflow_run` needs the action's yml file to exist on the default branch. [yanguoyu#38 (comment)](https://github.com/yanguoyu/neuron/pull/38#issuecomment-1569914593) is an example of a comment package for committing to PR.
> > 
> > 1. Add pull_request trigger for the `Package Neuron for Test` action
> > 2. When the action is triggered by `pull_request` event, packaging without code sign.
> > 3. After package success, comment on the PR with the package result.
> 
> It requires an extra action to sign a package, what if sign packages when the PR author is permitted, as we did in **add-replied-label**([`develop`/.github/workflows/add-replied-label.yml#L13-L29](https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/add-replied-label.yml?rgh-link-date=2023-06-01T07%3A13%3A15Z#L13-L29))

The original intention is to skip `code sign` when `certificates` are missing. So `sign-when-permitted` is not good enough too because `code sign` will be skipped even the contributor has his/her own certificates.

Ref: https://github.com/Magickbase/neuron-public-issues/issues/154

yanguoyu:
> > Because the trigger on `workflow_run` needs the action's yml file to exist on the default branch. [yanguoyu#38 (comment)](https://github.com/yanguoyu/neuron/pull/38#issuecomment-1569914593) is an example of a comment package for committing to PR.
> > 
> > 1. Add pull_request trigger for the `Package Neuron for Test` action
> > 2. When the action is triggered by `pull_request` event, packaging without code sign.
> > 3. After package success, comment on the PR with the package result.
> 
> It requires an extra action to sign a package, what if sign packages when the PR author is permitted, as we did in **add-replied-label**(https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/add-replied-label.yml#L13-L29)

> [Workflows in forked repositories](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories)
Workflows don't run in forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.
With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The GITHUB_TOKEN has read-only permissions in pull requests from forked repositories. For more information, see "[Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)."

From the GitHub action document, because most of the PR is opened by the forked repository, the action cannot read other secrets. But for every push event, it will sign with certificates. it's enough to sign on `push` event.

Developers can get the signed package when they push codes to their repository if they have certificates.

Keith-CY:
> every package in the PR comment will be unsigned, and every package in the commit message will be signed.

If I forked Neuron to `Keith-CY/neuron` and I don't have certificates configured in `Keith-CY/neuron`, will my commits be signed?

---

If `sign code` depends on if there's a certificate, I can fork `nervosnetwork/neuron` to `Keith-CY/neuron`, and sign every commit. There's no need to have an unsigned neuron if I have a certificate.

yanguoyu:
> If I forked Neuron to Keith-CY/neuron and I don't have certificates configured in Keith-CY/neuron, will my commits be signed?

It will try to sign but will fail.

I want to keep uniform rules, then every pr comment package will be unsigned, and if a commit comment exists it will be a signed package.
For skip code sign when certificates are missing. Then Sometimes the package is signed, and sometimes the package is unsigned. 

For the original intention, I will change the GitHub action, to skip code sign when certificates are missing.

Keith-CY:
CI failed at https://github.com/nervosnetwork/neuron/actions/runs/5150607517/jobs/9274938116?pr=2693#step:13:91

.github/workflows/comment_for_pr_package.yml
Keith-CY, line 1: Should we keep the `comment to package without a code sign` action?

Keith-CY: I've misunderstood this action

.github/workflows/package_for_test.yml
```diff
@@ -135,8 +162,29 @@ jobs:
...
+          echo $RUN_ID > ./package_info/run_id
+          echo $SHA > ./package_info/sha
+      - uses: actions/upload-artifact@v3
+        with:
+          name: package_info
+          path: package_info/
+          retention-days: 3
   comment:
     needs: [packaging]
+    if: ${{ github.event_name == 'push' }}
```
WhiteMinds, line 187: I think we should add comments to these places where the background reasons are not easily understood from the code, to avoid them becoming increasingly difficult to maintain.

WhiteMinds (changes requested):

packages/neuron-wallet/scripts/notarize.js
```diff
@@ -6,6 +6,11 @@ exports.default = async function notarizing(context) {
     return
   }
 
+  if (process.env.SKIP_NOTARIZE === 'true') {
+    console.warn('Skip notarizing when package on pull_request')
```
Keith-CY, line 10: The condition doesn't cover if the package is triggered by a pull_request, so the message is misleading.

.github/workflows/package_for_test.yml
```diff
@@ -62,8 +67,9 @@ jobs:
           CI: false
 
       - name: Package for MacOS
-        if: matrix.os == 'macos-latest'
+        if: ${{ matrix.os == 'macos-latest' && env.APPLE_ID != '' }}
         run: |
+          export CSC_FOR_PULL_REQUEST=true
```
Keith-CY, line 72: what does this `CSC_FOR_PULL_REQUEST` do, this keyword is not found in https://www.electron.build/, and there's a warning about it at https://github.com/electron-userland/electron-builder/blob/e3fc9b544cc8c6728ffd77a45408d6e0e87dbb46/packages/app-builder-lib/src/codeSign/macCodeSign.ts#L39

.github/workflows/package_for_test.yml
```diff
@@ -73,8 +79,19 @@ jobs:
           CSC_LINK: ${{ secrets.MAC_CERTIFICATE_BASE64 }}
           CSC_KEY_PASSWORD: ${{ secrets.MAC_CERTIFICATE_PASSWORD }}
 
+      - name: Package for MacOS for skip code sign
+        if: ${{ matrix.os == 'macos-latest' && env.APPLE_ID == '' }}
+        run: |
+          export CSC_IDENTITY_AUTO_DISCOVERY=false
```
Keith-CY, line 85: `CSC_IDENTITY_AUTO_DISCOVERY` can be set in `env` along with `SKIP_NOTARIZE`.
Another question is that is `CSC_IDENTITY_AUTO_DISCOVERY=false` necessary when `SKIP_NOTARIZE=true`?

.github/workflows/package_for_test.yml
yanguoyu: >  Current build is a part of pull request, code signing will be skipped.
Set env CSC_FOR_PULL_REQUEST to true to force code signing.

For electron builder, it will always skip code sign if the action is triggered by `pull_request` unless set CSC_FOR_PULL_REQUEST true.

.github/workflows/package_for_test.yml
yanguoyu: `CSC_IDENTITY_AUTO_DISCOVERY ` uses for code sign. And `SKIP_NOTARIZE` uses for notarizing.
https://developer.apple.com/documentation/security/code_signing_services
https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution

.github/workflows/package_for_test.yml
Keith-CY: I'm not sure if we should take the risk mentioned at https://github.com/electron-userland/electron-builder/blob/e3fc9b544cc8c6728ffd77a45408d6e0e87dbb46/packages/app-builder-lib/src/codeSign/macCodeSign.ts#L39 to force PR triggering for packaging.

> "There are serious security concerns with CSC_FOR_PULL_REQUEST=true (see the  CircleCI documentation (https://circleci.com/docs/1.0/fork-pr-builds/) for details)" +
    "\nIf you have SSH keys, sensitive env vars or AWS credentials stored in your project settings and untrusted forks can make pull requests against your repo, then this option isn't for you."

It would be better to add this annotation next to the configuration.

Besides, what dost it cost if the package of a PR is not signed

yanguoyu: Maybe I can remove `CSC_FOR_PULL_REQUEST`, because for `pull_request` event, it cannot access other secret variables, so if `APPLE_ID ` is not null, it must be triggered by the `pull` event.

Keith-CY (approved):

devchenyan (approved):

PR #2702, opened 2023-06-06
feat: Custom dev ui port(load env from .env)

FrankFang:
OS: Windows 10
When I run `yarn start`, I get `Something is already running on port 3000.`. 
It's good if I can customize the dev port.

P.S. I am sure there is no server running on port 3000 because I just restarted my computer. Maybe something is broken but I no time to figure it out, so custimzation of port is needed for me.


FrankFang:
Add a shebang for start.mjs

scripts/start.mjs
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env zx
+
+import dotenv from 'dotenv'
+import path from 'path'
+dotenv.config({ path:  path.join(__dirname, '..', 'packages', 'neuron-wallet', '.env.development.local')})
```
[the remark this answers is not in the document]
FrankFang: > NODE_ENV !== 'test' && `${paths.dotenv}.local`,

I understand all the code except for this line.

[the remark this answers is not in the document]
homura: My bad, I have updated it, and the code is just a variant from the [CRA](https://github.com/facebook/create-react-app/blob/0f5e990b8a04f53861d64ff53751517bbf73d867/packages/react-scripts/config/env.js#L25-L49)

FrankFang:
I update a lot of code:
1. Create neuron-shared as a shared package
2. Add load-env.ts into neuron-shared and test it
3. Upgrade dotenv from 8 to 16.1.4 
4. Use loadEnv to get env vars anywhere

packages/neuron-shared/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@nervina-labs/neuron-shared",
```
Keith-CY, line 2:
```suggestion
  "name": "@nervosnetwork/neuron-shared",
```

packages/neuron-shared/tests/load-env.test.ts
```diff
@@ -0,0 +1,46 @@
+import { loadEnv } from "../src/load-env"
+import fs from 'fs'
+
+
+jest.mock('fs')
```
Keith-CY, line 5:
```suggestion
jest.mock('node:fs')
```

packages/neuron-shared/tests/load-env.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    fs.readFileSync = jest.fn().mockImplementation((filepath) => {
+      if (filepath === '.env.test.local') {
+        return 'TEST_ENV="from .env.test.local"'
+      } else if (filepath === '.env.test') {
+        return 'TEST_ENV="from .env.test"'
+      } else if (filepath === '.env') {
+        return 'TEST_ENV="from .env"'
+      } else {
+        return ''
+      }
```
Keith-CY, line 17: Multiple `if-else statements` could be replace by `switch-case statement` for readability

packages/neuron-shared/tests/load-env.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    })
+  })
+  afterAll(() => {
+    jest.resetAllMocks()
+  })
+  describe("read one by one", () => {
```
Keith-CY, line 23: `it("no matched .env files", () => {})` should be covered to ensure no unexpected files will be loaded when there aren't any `.env` files

packages/neuron-wallet/package.json
```diff
@@ -42,13 +42,13 @@
     "@iarna/toml": "2.2.5",
     "@ledgerhq/hw-transport-node-hid": "6.27.13",
     "@nervina-labs/ckb-indexer": "0.1.1",
+    "@nervina-labs/neuron-shared": "0.110.1",
```
Keith-CY, line 46: Should be updated along with https://github.com/nervosnetwork/neuron/pull/2702/files#diff-059a0f39aaa673028ecc51dfab39fbdf84452a7b5a830c6820008f82331576acR2

packages/neuron-wallet/src/env.ts
```diff
@@ -1,7 +1,7 @@
 import os from 'os'
 import path from 'path'
-import dotenv from 'dotenv'
 import { app as electronApp } from 'electron'
+import {loadEnv} from '@nervina-labs/neuron-shared'
```
Keith-CY, line 4:
```suggestion
import { loadEnv } from '@nervina-labs/neuron-shared'
```

package.json
```diff
@@ -56,7 +57,8 @@
     "prettier": "2.8.7",
     "ts-jest": "27.1.5",
     "typescript": "5.0.4",
-    "wait-on": "6.0.1"
+    "wait-on": "6.0.1",
+    "zx": "^7.2.2"
```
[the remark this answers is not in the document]
FrankFang: get it.

FrankFang:
It's late now, I'm going to sleep.

packages/neuron-shared/tests/load-env.test.ts
FrankFang: I suggest use fs instead of node:fs for now, reasons:
1. [dotenv use 'fs'](https://github.com/search?q=repo%3Amotdotla%2Fdotenv+fs&type=code), so even if we use node:fs, we have to mock fs too. 
2. I see a lof of usages of 'fs' in the codebase, it's a little confusing to use 'fs' and 'node:fs' at the same time.

FrankFang:
Most are fixed but not 'node:fs’

packages/neuron-shared/tests/load-env.test.ts
Keith-CY: > I suggest use fs instead of node:fs for now, reasons:
> 
> 1. [dotenv use 'fs'](https://github.com/search?q=repo%3Amotdotla%2Fdotenv+fs&type=code), so even if we use node:fs, we have to mock fs too.
> 2. I see a lof of usages of 'fs' in the codebase, it's a little confusing to use 'fs' and 'node:fs' at the same time.

It's reasonable

packages/neuron-wallet/package.json
```diff
@@ -42,13 +42,13 @@
     "@iarna/toml": "2.2.5",
     "@ledgerhq/hw-transport-node-hid": "6.27.13",
     "@nervina-labs/ckb-indexer": "0.1.1",
+    "@nervosnetwork/neuron-shared": "0.110.1",
     "@nervosnetwork/ckb-sdk-core": "0.107.0",
     "@nervosnetwork/ckb-sdk-utils": "0.107.0",
```
Keith-CY, line 47:
```suggestion
    "@nervosnetwork/ckb-sdk-core": "0.107.0",
    "@nervosnetwork/ckb-sdk-utils": "0.107.0",
    "@nervosnetwork/neuron-shared": "0.110.1",
```

Dependencies managed by `lerna` are ordered alphanumerically.

packages/neuron-shared/tests/load-env.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      expect(process.env.TEST_VAR).toEqual('from .env.test')
+    })
+    it(".env.test.local > .env.test > .env", () => {
+      fs.existsSync = jest.fn().mockImplementation((filepath) => {
+        return ['.env.test.local', '.env.test', '.env'].includes(filepath)
+      })
+      loadEnv()
+      expect(process.env.TEST_VAR).toEqual('from .env.test.local')
+    })
+    it("does not throw error when no env file exists", () => {
```
Keith-CY, line 57:
```suggestion
    it("neither import files nor throw error when no env file exists", () => {
```

packages/neuron-shared/tests/load-env.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      expect(process.env.TEST_VAR).toEqual('from .env.development')
+    })
+    it(".env.development.local > .env.development > .env", () => {
+      fs.existsSync = jest.fn().mockImplementation((filepath) => {
+        return ['.env.development.local', '.env.development', '.env'].includes(filepath)
+      })
+      loadEnv()
+      expect(process.env.TEST_VAR).toEqual('from .env.development.local')
+    })
+    it("does not throw error when no env file exists", () => {
```
Keith-CY, line 91:
```suggestion
    it("neither import files nor throw error when no env file exists", () => {
```

packages/neuron-shared/tests/load-env.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      expect(process.env.TEST_VAR).toEqual('from .env.production')
+    })
+    it(".env.production.local > .env.production > .env", () => {
+      fs.existsSync = jest.fn().mockImplementation((filepath) => {
+        return ['.env.production.local', '.env.production', '.env'].includes(filepath)
+      })
+      loadEnv()
+      expect(process.env.TEST_VAR).toEqual('from .env.production.local')
+    })
+    it("does not throw error when no env file exists", () => {
```
Keith-CY, line 126:
```suggestion
    it("neither import files nor throw error when no env file exists", () => {
```

packages/neuron-ui/package.json
```diff
@@ -48,6 +48,7 @@
     "i18next": "21.10.0",
     "immer": "9.0.21",
     "jsqr": "1.4.0",
+    "@nervosnetwork/neuron-shared": "0.110.1",
```
Keith-CY, line 51: Github suggestion is not available here, so it should be updated manually.

This dependency should be placed beneath `"@nervosnetwork/ckb-sdk-utils": "0.107.0",` because dependencies imported by `lerna` are ordered alphanumerically.

Keith-CY (approved):

FrankFang:
I initiated a discussion https://github.com/nervosnetwork/neuron/discussions/2704

packages/neuron-shared/src/load-env.ts
```diff
@@ -17,8 +18,7 @@ export const loadEnv = () => {
   for (let i = 0; i < dotenvFiles.length; i++) {
     const dotenvFile = dotenvFiles[i]
     if (fs.existsSync(dotenvFile)) {
-      dotenv.config({ path: dotenvFile })
-      break
+      dotenvExpand.expand(dotenv.config({ path: dotenvFile, override: false }))
```
FrankFang, line 21: dotenvExpand make `ENV_2=${ENV_1}` possible.

yanguoyu:
With the PR, we can not auto-start ui with another port when run the command `yarn start:ui` if the default port has been used. And if the 3000 port has been used, we cannot get any help from the output. The `neuron-wallet` only waits for `http://127.0.0.1:3000` because the `neuron-ui` does not throw any exceptions with the used port.

FrankFang:
> The `neuron-wallet` only waits for `http://127.0.0.1:3000`

I've change the start task for `neuron-wallet`
```
"start": "node scripts/wait-ui.mjs && yarn run start:debug
```

So `neuron-wallet` could wait for `http://127.0.0.1:${PORT}` now.

WhiteMinds:
My main concern currently is to avoid introducing the complexity of `neuron-shared` too early. At the moment, it only provides a unified rule for loading `envFiles`, which we can achieve using libraries like `@next/env`.

@Keith-CY @yanguoyu @devchenyan @JeffreyMa597 @homura , what do you think?

If we all agree to introduce `neuron-shared`, there are a few tasks that need to be done:
1. Automatically compile on the first yarn, similar to this: https://github.com/nervosnetwork/neuron/blob/e7adec4b89ce9976ba9d598960a50f7dce176cff/package.json#LL38C38-L38C87
2. Like `neuron-wallet` and `neuron-ui`, include `Prettier` and `ESLint`.
3. Optionally, rename `main.ts` to a more conventional `index.ts`

Keith-CY:
> My main concern currently is to avoid introducing the complexity of `neuron-shared` too early. At the moment, it only provides a unified rule for loading `envFiles`. Which we can achieve using libraries like `@next/env`. This way, we can directly implement the same logic in both packages through simple code duplication, without the need for a new package implementation.
> 
> @Keith-CY @yanguoyu @devchenyan @JeffreyMa597 @homura , what do you think?
> 
> If we all agree to introduce `neuron-shared`, there are a few tasks that need to be done:
> 
> 1. Automatically compile on the first yarn, similar to this: [`e7adec4`/package.json#LL38C38-L38C87](https://github.com/nervosnetwork/neuron/blob/e7adec4b89ce9976ba9d598960a50f7dce176cff/package.json#LL38C38-L38C87)
> 2. Like `neuron-wallet` and `neuron-ui`, include `Prettier` and `ESLint`.
> 3. Optionally, rename `main.ts` to a more conventional `index.ts`

It may be redundant to share an `env` through a package, it can be done simply by extending `env`s from the project root.

WhiteMinds:
> It may be cumbersome to share an `env` through a package, it can be done simply by extending `env`s from the project root.

Currently, we are not sharing the environment variables themselves, but rather the `loadEnv` function. In reality, the environment configuration for each project is still independent. Therefore, I believe introducing this complexity (`neuron-shared`) is unnecessary.

yanguoyu:
> > It may be cumbersome to share an `env` through a package, it can be done simply by extending `env`s from the project root.
> 
> Currently, we are not sharing the environment variables themselves, but rather the `loadEnv` function. In reality, the environment configuration for each project is still independent. Therefore, I believe introducing this complexity (`neuron-shared`) is unnecessary.

I strongly agree, and `lerna` can auto-load environment files when running tasks with `lerna`.

Keith-CY:
Will this PR be updated according to the comments above?

PR #2709, opened 2023-06-09
fix: New UI menu bar issues

JeffreyMa597:
------
As title.

Ref:https://github.com/Magickbase/neuron-public-issues/issues/158

Check List
------

### Test
e2e Test


packages/neuron-ui/src/styles/mixin.scss
```diff
@@ -175,8 +175,12 @@
...
       g,
       path {
-        fill: $hoverColor
+        fill: $hoverColor;
       }
     }
   }
 }
+
+@mixin cursor-mode {
```
yanguoyu, line 184: Is it necessary to define one property with `mixin`? I think `cursor: pointer` is clear to replace `@include cursor-mode;`

yanguoyu (approved):

WhiteMinds:
Scroll bars and unexpected overflows when shrinking windows

image

The language option is missing in the settings, I can't switch languages.

WhiteMinds (changes requested):
Scroll bars and unexpected overflows when shrinking windows

image

The language option is missing in the settings, I can't switch languages.

JeffreyMa597:
> Scroll bars and unexpected overflows when shrinking windows

updated. 
image


> The language option is missing in the settings, I can't switch languages.

You can select them over here.
image

WhiteMinds (approved):

WhiteMinds:
> You can select them over here.

I can't believe I looked at it 3 times and didn't notice it 🫠. Is it my problem or a UI issue?

Keith-CY:
> > You can select them over here.
> 
> I can't believe I looked at it 3 times and didn't notice it 🫠. Is it my problem or a UI issue?

@Danie0918 @Sven-TBD please check if it is user friendly

Danie0918:
> > > You can select them over here.
> > 
> > 
> > I can't believe I looked at it 3 times and didn't notice it 🫠. Is it my problem or a UI issue?
> 
> @Danie0918 @Sven-TBD please check if it is user friendly

Language setting is not a high-frequency operation, so it is reasonable to put it in the settings and continue the previous user habits. The problem here is that the language of the current settings is weakened, so users can't find it quickly, and this needs to be optimized by adding an icon or deepening the color.  @Kirl

Kirl70:
> > > > You can select them over here.
> > > 
> > > 
> > > I can't believe I looked at it 3 times and didn't notice it 🫠. Is it my problem or a UI issue?
> > 
> > 
> > @Danie0918 @Sven-TBD please check if it is user friendly
> 
> Language setting is not a high-frequency operation, so it is reasonable to put it in the settings and continue the previous user habits. The problem here is that the language of the current settings is weakened, so users can't find it quickly, and this needs to be optimized by adding an icon or deepening the color. @Kirl70

Icons have been added in front of the options as an indication:
https://www.figma.com/file/6XNoimRDbFTTNm016rbIdU/Magickbase?type=design&node-id=1165%3A12477&t=JhtdSRXHKrCkhAGF-1
image

JeffreyMa597:
> Icons have been added in front of the options as an indication:
https://www.figma.com/file/6XNoimRDbFTTNm016rbIdU/Magickbase?type=design&node-id=1165%3A12477&t=JhtdSRXHKrCkhAGF-1

Updated.

FrederLu (approved):

devchenyan (approved):

PR #2711, opened 2023-06-09
fix: Fix sync stop with exception and multisig balance.

yanguoyu:


devchenyan (approved):

packages/neuron-wallet/src/services/asset-account-service.ts
WhiteMinds, line 1: Strange, this file appears differently formatted on my end. Could it be because the base branch used when creating this commit is older?

image

packages/neuron-wallet/src/services/asset-account-service.ts
```diff
@@ -256,14 +256,23 @@ export default class AssetAccountService {
...
+      })
+      .getCount()
+    // check whether the entity exists before insert. Reason: https://github.com/Magickbase/neuron-public-issues/issues/184#issue-1749746997
+    if (!existAccountAcount) {
+      await getConnection()
+        .createQueryBuilder()
+        .insert()
+        .into(AssetAccountEntity)
+        .values(assetAccountEntity)
+        .execute()
```
WhiteMinds, line 274: I'm not very familiar with TypeORM, but can we directly use `await getConnection().getRepository(SudtTokenInfoEntity).upsert` here instead of making two separate queries?
In theory, the database should support such an operation, and it might be a matter of how to achieve this through TypeORM's API calls.

packages/neuron-wallet/src/database/chain/entities/asset-account.ts
```diff
@@ -60,7 +60,7 @@ export default class AssetAccount {
       this.sudtTokenInfo.decimal,
       this.balance,
       this.blake160,
-      this.id
+      this.id ?? undefined
```
WhiteMinds, line 63: Why could the `id` be null? We need a comment here. otherwise, this line of code might be removed by someone else in the future.

WhiteMinds (changes requested):

packages/neuron-wallet/src/services/asset-account-service.ts
yanguoyu: More detail can be found https://github.com/Magickbase/neuron-public-issues/issues/184#issue-1749746997, it may be a bug of `typeorm`

packages/neuron-wallet/src/services/asset-account-service.ts
yanguoyu: The base branch is the latest when I create the PR.

packages/neuron-wallet/src/database/chain/entities/asset-account.ts
yanguoyu: I forget to remove it, I will remove it later.

JeffreyMa597 (approved):

packages/neuron-wallet/src/services/asset-account-service.ts
WhiteMinds: fixed on https://github.com/nervosnetwork/neuron/pull/2720

WhiteMinds (approved):

Keith-CY:
Please have a test @FrederLu @littleLip520 @silySuper

silySuper:
Verified

silySuper (approved):

PR #2714, opened 2023-06-12
feat: Add package for pull request and comment (#2693)

yanguoyu:
Merge develop to new-ui


Keith-CY:
We need to use `npm/yearn install` in stead of `lerna` first

yanguoyu:
The `yarn.lock` has a conflict with default merge, I create a new PR https://github.com/nervosnetwork/neuron/pull/2717

PR #2715, opened 2023-06-12
feat: Update lerna to v7

yanguoyu:
For `lerna` v7 run, it seems will auto-read the .env file in the current directory. So I need to change the test file.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: Boostrap
         run: |
-          yarn bootstrap
+          yarn
         env:
           CI: false
 
--- a/.github/workflows/package_for_test.yml
+++ b/.github/workflows/package_for_test.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: Boostrap
         run: |
-          yarn bootstrap
+          yarn
         env:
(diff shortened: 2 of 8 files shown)


packages/neuron-wallet/tests/controllers/sync-api.test.ts
```diff
@@ -172,6 +172,7 @@ describe('SyncApiController', () => {
             cacheRate: undefined,
             estimate: undefined,
             status: 3,
+            validTarget: process.env.CKB_NODE_ASSUME_VALID_TARGET,
```
homura, line 175: I think we should avoid the side effects in unit tests when we can, and the `process.env` is a case of side effect, maybe we could launch the unit test with [--load-env-files=false](https://lerna.js.org/docs/features/run-tasks#automatic-loading-of-env-files) to disable it

Keith-CY: And the `env.CKB_NODE_ASSUMED_VALID_TARGET` can be mocked

homura (approved):
LGTM

JeffreyMa597 (approved):

Keith-CY (approved):

PR #2721, opened 2023-06-14
fix: issues for history page

JeffreyMa597:
------
As title.

Ref:https://github.com/Magickbase/neuron-public-issues/issues/169

Check List
------

### Test
e2e Test

### DIFF
--- a/packages/neuron-ui/src/utils/formatters.ts
+++ b/packages/neuron-ui/src/utils/formatters.ts
@@ -310,3 +310,14 @@ export const padFractionDigitsIfDecimal = (num: string | number, minimumFraction
   const isDecimal = numText.includes('.')
   return isDecimal ? numText.padEnd(numText.indexOf('.') + 1 + minimumFractionDigits, '0') : numText
 }
+
+/**
+ * Convert complex numbers for millennial counting to pure numbers.
+ * ex:
+ * -100.00002362 CKB ---> -100.00002362
+ * +10,000 CKB ---> 10000
+ *  */
+export const complexNumberToPureNumber = (value: string) => {
+  const num = Number(value.replace(/[^-0-9.]/g, ''))
+  return num
+}
--- a/packages/neuron-ui/src/components/History/RowExtend.tsx
+++ b/packages/neuron-ui/src/components/History/RowExtend.tsx
@@ -2,7 +2,7 @@ import React, { useCallback } from 'react'
 import { useDispatch } from 'states'
 import { openExternal } from 'services/remote'
 
-import { getExplorerUrl, localNumberFormatter, RoutePath, useLocalDescription } from 'utils'
+import { clsx, getExplorerUrl, localNumberFormatter, RoutePath, useLocalDescription } from 'utils'
 import { TableProps } from 'widgets/Table'
 import CopyZone from 'widgets/CopyZone'
 import { useNavigate } from 'react-router-dom'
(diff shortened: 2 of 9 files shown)


packages/neuron-ui/src/utils/formatters.ts
```diff
@@ -304,3 +304,14 @@ export const errorFormatter = (error: string | FailureFromController['message'],
...
   return error.content || unknownError
 }
+
+/**
+ * Convert complex numbers for millennial counting to pure numbers.
+ * ex:
+ * -100.00002362 CKB ---> -100.00002362
+ * +10,000 CKB ---> 10000
+ *  */
+export const complexNumberToPureNumber = (value: string) => {
```
yanguoyu, line 314: Should add test case

yanguoyu (approved):

packages/neuron-ui/src/widgets/CopyZone/index.tsx
```diff
@@ -15,14 +15,18 @@ const CopyZone = ({ children, content, name, style, className = '' }: CopyZonePr
...
-    setCopied(true)
-    window.navigator.clipboard.writeText(content)
-    clearTimeout(timer.current!)
-    timer.current = setTimeout(() => {
-      setCopied(false)
-    }, 1000)
-  }, [setCopied, content])
+  const onCopy = useCallback(
+    (event: React.MouseEvent<HTMLDivElement>) => {
+      event.stopPropagation()
```
WhiteMinds, line 20: Try to avoid using stopPropagation as much as possible, as it is not a good pattern. Related articles:
https://stackoverflow.com/a/54817255/21858805

WhiteMinds (changes requested):

littleLip520:
@JeffreyMa597 please help to push this review, tks

devchenyan (approved):

packages/neuron-ui/src/widgets/CopyZone/index.tsx
JeffreyMa597: @WhiteMinds 
I tried to separate the 'copy' operation from the click event by using the type of event.target to determine whether the user clicked 'copy' or 'expand'. However, in order to do this, the information about which row was clicked needs to be passed into the Table component through props for it to determine whether to expand. This would greatly increase the complexity of the Table component, making it messy. It would also add a burden to the render function, and I believe that determining which row was clicked should indeed be maintained by the Table component itself. Therefore, I choose to continue using stopPropagation to prevent its bubbling behavior to achieve our needs.

littleLip520 (approved):

Keith-CY:
Please resolve the conflict and squash merge

packages/neuron-ui/src/widgets/CopyZone/index.tsx
WhiteMinds: The drawbacks of using stopPropagation are discussed in the link provided above, but to summarize, it can cause issues such as disabling onClickOutside.

If this decision is made, it is likely that the bad pattern will spread further when other clickable elements are added to the rows in the future, and it will be difficult to reverse.

I can suggest some alternative solutions, but they will all require event checking in onRowClick of the table to determine whether to execute the expand function:

1. The most common solution is to check event.target. If this is deemed too complex or interferes with readability, we can use other solutions.
If we want to use this approach, we can further consider how to design it.

3. Add a new property to the event, such as preventTableExtend (of type symbol, to prevent interference from other components that may add this property).
Then, the Table component can export two functions: setPreventTableExtend(event, bool) and isPreventTableExtend(event).
This will limit this special pattern to the Table component and all places that use this function will know exactly what they are doing, unlike event.stopPropagation(), which makes it difficult to remove because developers are unsure which component it is serving.

4. Use the built-in [cancelable](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelable) property, but this is not common and can reduce readability, so it is not recommended.

JeffreyMa597: @WhiteMinds update

WhiteMinds (approved):

PR #2731, opened 2023-06-20
fix: If event_name is equal pull_request, we should skip code sign

yanguoyu:


WhiteMinds (approved):

Keith-CY:
I would suggest adding a step of the job to output if the package should be signed because the condition is complex

yanguoyu:
> I would suggest adding a step of the job to output if the package should be signed because the condition is complex

https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/package_for_test.yml#L122-L148 is the job to upload packages for `macos-lateste`. It does not care whether the code sign will happen.

yanguoyu:
> > > I would suggest adding a step of the job to output if the package should be signed because the condition is complex
> > 
> > 
> > [`develop`/.github/workflows/package_for_test.yml#L122-L148](https://github.com/nervosnetwork/neuron/blob/develop/.github/workflows/package_for_test.yml?rgh-link-date=2023-06-20T06%3A52%3A46Z#L122-L148) is the job to upload packages for `macos-lateste`. It does not care whether the code sign will happen.
> 
> The `output` in my context is to set a flag in github actions, the flag makes condition more clear
> 
> Current conditions are slightly different and prone to mix.
> 
> ```
>         if: ${{ matrix.os == 'macos-latest' && env.APPLE_ID != '' && github.event_name != 'pull_request' }}
>         if: ${{ matrix.os == 'macos-latest' && (env.APPLE_ID == '' || github.event_name == 'pull_request') }}
> ```

I use `MAC_SHOULD_CODE_SIGN` to choose the step to package MacOS.

PR #2734, opened 2023-06-20
Refactor: remove ckb-sdk-utils and repalce with lumos

zhangyouxin:
## Description

Due to merge of `ckb-sdk-x` to `lumos`, we need to gradually deprecate `ckb-sdk-x`, this PR removes `ckb-sdk-utils` in `packages/neuron-wallet`, It's not a small PR and may be not that easy to review, so I plan to seperate this task to two PRs, will create another PR to remove it in `packages/neuron-ui`

This PR is not not compilable, because I have updated `lumos` to `0.20`, some snake_case fields has changed to camelCase, the task is in another PR: https://github.com/nervosnetwork/neuron/pull/2683, only when after that PR is merged, this PR can be ready for review.

Resolve https://github.com/Magickbase/neuron-public-issues/issues/201

## changed

- `scriptToHash(script)` should be replaced with `utils.computeScriptHash(script)`.
- `scriptToAddress(script, isMainet)` should be replaced with `helpers.encodeToAddress(script, lumosOptions)`. We should translate `isMainet` to `lumosOptions` here.
- `addressToScript(address)` should be replaced with `helpers.addressToScript(address, lumosOptions)`. We should detect if `isMainet` by checking if the address starts with 'ckb' or not. Alternatively, we could use `NetworksService.getInstance().isMainnet()`.
- `src/models/blake2b.ts` has been deleted and replaced with `@ckb-lumos/hd` and `lumos helpers`.
- `null` values are not compatible with `@ckb-lumos/base` types.
- Some test cases have been changed:
    - `args` should be a valid hex string. Values like "0xuuid" have been deleted.
    - Optional hex values should be `0x` or `undefined`. Empty strings have been deleted.
    - Hex strings are not hex numbers. Values like `0x1` have been changed to `0x01`.


homura:
@zhangyouxin the conflicts is needed to resolve

packages/neuron-wallet/tests/controllers/sudt.test.ts
```diff
@@ -63,8 +63,8 @@ describe('SUDTController', () => {
       'txHash',
       '0',
       '10000',
-      Script.fromObject({ codeHash: '', args: '', hashType: ScriptHashType.Type }),
-      Script.fromObject({ codeHash: '', args: '', hashType: ScriptHashType.Type }),
+      Script.fromObject({ codeHash: `0x${'00'.repeat(32)}`, args: '0x', hashType: ScriptHashType.Type }),
```
homura, line 66: it is better to add some comment to describe why the empty `codeHash` is required here

packages/neuron-wallet/src/models/transaction-size.ts
```diff
@@ -28,7 +29,15 @@ export default class TransactionSize {
...
-    const bytes = serializeOutput(output.toSDK())
+    // const bytes = serializeOutput(output.toSDK())
+    const cellOutput = output.toSDK()
+    const bytes = bytesUtils.hexify(
+      blockchain.CellOutput.pack({
+        capacity: cellOutput.capacity,
+        lock: cellOutput.lock,
+        type: cellOutput.type || undefined, // TODO: null is not a packable type
+      })
+    )
```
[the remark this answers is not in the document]
zhangyouxin: `cellOutput.type` could be `null`, so maybe it's better to provide a translator to convert `CKBCompont` types with `null`s to `lumos` compatible types with `undefined`s so that the code may look nicer.

packages/neuron-wallet/src/models/transaction-size.ts
```diff
@@ -65,10 +77,19 @@ export default class TransactionSize {
...
+  /**
+   * @param witness a witness object or a hex string
+   * @returns witness wraped in a fixed Uint32Le vector (+4 bytes for header) and add another +4 bytes for the offset
+   */
   public static witness(witness: WitnessArgs | string): number {
-    const wit: string = typeof witness === 'string' ? witness : serializeWitnessArgs(witness.toSDK())
-    const bytes = serializeFixVec(wit)
-    return HexUtils.byteLength(bytes) + TransactionSize.SERIALIZED_OFFSET_BYTESIZE
+    if (typeof witness === 'string') {
+      const fixvecCodec = fixvec(number.Uint8)
```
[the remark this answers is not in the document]
zhangyouxin: Yes this part of code is hard to understand. So I just try to keep the logic as it is.

The original code just wraps `witness` in `Uint8Array` by using `serializeFixVec`.

The idea you provided here is changing original logic, which may cause unintended bugs, please refer to : https://github.com/nervosnetwork/neuron/blob/55e0db3c4d7f1b1d4d0770b06056d4dd78f0514c/packages/neuron-wallet/src/models/transaction-size.ts#L68-L72

packages/neuron-wallet/tests/controllers/sudt.test.ts
zhangyouxin: I have described the reason in the PR comment, please refer to:https://github.com/nervosnetwork/neuron/pull/2734#issue-1765073549

packages/neuron-wallet/tests/services/address.test.ts
```diff
@@ -310,7 +310,8 @@ describe('integration tests for AddressService', () => {
 
     describe('#generateAndSaveForPublicKey', () => {
       describe('with public key info exist for the public key', () => {
-        const publicKey = 'public key'
+        // const publicKey = 'public key' // not valid public key
+        const publicKey = '0x' + '00'.repeat(65) // 65 bytes pub key
```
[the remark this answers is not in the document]
zhangyouxin: It can't pass test after refactor, because `publicKey` is used to generate secp256k1 `args`, in `lumos` we can call `utils.ckbHash(pubkey).slice(0, 42)`, which requires `publicKey` to be a valid `BytesLike` value.

packages/neuron-wallet/tests/controllers/multisig.test.ts
```diff
@@ -77,13 +77,13 @@ const multisigConfig = {
     params: {
       multisig_configs: {
         sighash_addresses: multisigBlake160s.map(args =>
-          scriptToAddress(
+          helpers.encodeToAddress(
             {
               args,
-              codeHash: systemScripts.SECP256K1_BLAKE160.codeHash,
-              hashType: systemScripts.SECP256K1_BLAKE160.hashType,
+              codeHash: config.predefined.AGGRON4.SCRIPTS.SECP256K1_BLAKE160.CODE_HASH,
```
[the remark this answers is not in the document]
zhangyouxin: For this part, we are just using `SECP256K1_BLAKE160` code hash, which is the same in mainnet/testnet/devnet. 

Because `lumos` utils like `helpers.encodeToAddress`, `parseAddress` needs to receive a config, so I plan to create a static method to provide config like `NeuronInstance.getLumosConfig()`

packages/neuron-wallet/src/models/system-script-info.ts
```diff
@@ -13,11 +14,18 @@ export default class SystemScriptInfo {
...
+  // FIXME: this is werid that `blockchain.Script.pack` will throw error here
+  // static DAO_SCRIPT_HASH = new Script(
+  //   SystemScriptInfo.DAO_CODE_HASH,
+  //   '0x',
+  //   SystemScriptInfo.DAO_HASH_TYPE
+  // ).computeHash()
+
+  static DAO_SCRIPT_HASH = utils.computeScriptHash({
+    codeHash: config.predefined.LINA.SCRIPTS.DAO.CODE_HASH,
+    hashType: config.predefined.LINA.SCRIPTS.DAO.HASH_TYPE,
```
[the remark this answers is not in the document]
zhangyouxin: The codeHashes of systemScripts are the same in mainnet/testnet/devnet

packages/neuron-wallet/src/controllers/multisig.ts
```diff
@@ -12,6 +11,7 @@ import OfflineSignService from '../services/offline-sign'
 import Multisig from '../models/multisig'
 import SystemScriptInfo from '../models/system-script-info'
 import NetworksService from '../services/networks'
+import { config as lumosConig, helpers, utils, config } from '@ckb-lumos/lumos'
```
Keith-CY, line 14: What's the difference between `config as lumosConfig` and `config`

zhangyouxin: setting `lumosConfig` as alias for `config`, because there is another viarable name `config` in the scope

zhangyouxin:
I find that totally removal of `ckb-sdk-utils` may be not as easy as I thougt of, and some changes needs to be discussed, so is it a better idea to seperate the task to more small tasks? @homura 

After creating this PR, I can see some sub-tasks.

So I would like to close this PR and create each PR for:

- [ ] config management 
  - (discussion) how to manage a lumos config, or we don't need a full lumos config, just part of it
  - convert `Script` to `Address`
  - convert `Address` to `Script`
- [ ]  refactor `Script` class: `packages/neuron-wallet/src/models/chain/script.ts`
  - plan to remove this class, the methods of this class can be non-state, so it can be replaced by util functions
  - computeHash -> `utils.computeScriptHash()`
  - calculateBytesize -> `blockchain.Script.pack(script).bytelength`
- [ ] Serialization/Deserialization for `Script`, `RawTransaction`, `WitnessArgs` 
- [ ] refactor: `src/models/blake2b.ts`
  - replace with `CKBHasher` in `@ckb-lumos/base`
- [ ] Test cases
  - (discussion) some old conventions are not sutable for `lumos` style , such as script with `args: "0xnot-hex"` is not allowed when we use `blockChain.Script.pack(script)`

With the tasks seperated we can push the diliveration of this PR more smoothly

packages/neuron-wallet/tests/models/transaction-size.test.ts
```diff
@@ -25,12 +25,26 @@ describe('TransactionSize', () => {
...
+    //  input_type:"0x" // more 4 bytes representin '0' in Uint32 format to indicate the length of input_type
+    //  output_type:undefined
+    // }  
+    // it has the same for 0x1000000010000000100000 (in ckb-sdk-js format)
+    // {
+    //  lock: undefined
+    //  input_type: undefined
+    //  output_type: undefined
+    // }  
+    expect(result).toEqual(HexUtils.byteLength('0x1400000010000000100000001400000000000000') + 4 + 4)
```
yanguoyu, line 43: It may cause here https://github.com/ckb-js/lumos/blob/develop/packages/codec/src/blockchain.ts#L64-L66 Lumos set all fields as `option`. But `ckb-sdk-js` does not set fields as `option`

yanguoyu: I feel confused with https://github.com/ckb-js/lumos/blob/develop/packages/codec/src/blockchain.ts#L64-L66.
When I call `bytes.hexify(base.blockchain.WitnessArgs.pack({ lock: '0x', input_type: '0x', output_type: '0x'}))`, it will output `0x1400000010000000140000001400000000000000`, why `input_type` and `output_type` set empty, but `lock` set '0x00000000'.

packages/neuron-wallet/tests/services/tx/transaction-generator.test.ts
```diff
@@ -1158,7 +1161,8 @@ describe('TransactionGenerator', () => {
             feeRate,
             '0'
           )
-          tx.witnesses[0] = serializeWitnessArgs(WitnessArgs.emptyLock().toSDK())
+          // tx.witnesses[0] = bytes.hexify(blockchain.WitnessArgs.pack(WitnessArgs.emptyLock().toSDK())
```
yanguoyu, line 1164: Please remove annotation

packages/neuron-wallet/src/models/address-parser.ts
```diff
@@ -1,11 +1,14 @@
...
+import { config, helpers } from '@ckb-lumos/lumos'
 import SystemScriptInfo from './system-script-info'
 
 export default class AddressParser {
   public static parse(address: string): Script {
     try {
-      const script = addressToScript(address)
+      const isMainnet = address.startsWith('ckb')
+      const script = isMainnet
+        ? helpers.addressToScript(address, { config: config.predefined.LINA })
```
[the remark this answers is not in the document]
zhangyouxin: Good idea, it's common way to determine if `isMainnet` by address prefix, in the mean time, when we need to translate `Script` to `Address`, we still need to get a `isMainnet` env from somewhere else, so I try to solve the two convertion problem together, maybe more common way is to call `config.initializeConfig(isMainnet ? LINA : AGGRON4)` when network changes? @homura

packages/neuron-wallet/tests/models/transaction-size.test.ts
zhangyouxin: here '0x00000000' is `args`, check the test case setup: https://github.com/nervosnetwork/neuron/pull/2734/files/dd6a4deaea055f8627e5e3362e7c87ef52ef9865#diff-67c688a92fedba703d332bc20e7149a30b2439001df088a9b6038b0b06d63470R19

packages/neuron-wallet/tests/services/tx/transaction-generator.test.ts
zhangyouxin: ok

packages/neuron-wallet/src/models/address-parser.ts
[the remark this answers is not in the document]
homura: > Can we move the address.startsWith('ckb') logic to helpers.addressToScript function? 

It's ok, we'll implement it in the Lumos

packages/neuron-wallet/src/controllers/multisig.ts
Keith-CY: I didn't get it, `lumosConfig` and `config` are both from `@ckb-lumos/lumos.config`, pointing to the same one.

BTW, `lumosConig` is a typo

zhangyouxin: https://github.com/nervosnetwork/neuron/blob/b4f7de26155e09a4ce7cab0ad5364664c765ac1f/packages/neuron-wallet/src/controllers/multisig.ts#L110-L117

`config` is delared here in line 110, if `import { config } from '@ckb-lumos/lumos'`, then there are two `config`s in the mapFunction, I'll fix the typo

yanguoyu:
```suggestion
import { config as lumosConig, helpers, utils } from '@ckb-lumos/lumos'
```

PR #2736, opened 2023-06-20
version 0.106.0 for QA

devchenyan:
Set the version to facilitate the verification of the UI modification of the update interface.

https://github.com/Magickbase/neuron-public-issues/issues/179

ps:
This branch does not need to be merged.


Keith-CY:
image
It can be converted to a draft so it won't be merged by accident

PR #2739, opened 2023-06-21
refactor: fix typos in neuron-ui

Keith-CY:
This PR fixes typos in packages/neuron-ui module


WhiteMinds:
I still receive a lot of spelling error prompts when using typos locally.
Is this all the fixes, or are only a portion of them being done first?

JeffreyMa597 (approved):

Keith-CY:
> I still receive a lot of spelling error prompts when using typos locally. Is this all the fixes, or are only a portion of them being done first?

This fix is for neuron-ui module, an other PR for neuron-wallet will be proposed later

_typos.toml
```diff
@@ -0,0 +1,3 @@
+[default.extend-words]
+thur = "thur"
```
homura, line 2: where is the "thur" used in this PR?

Keith-CY: It's the whitelist to inform `typos` that `thur` is a valid spell, for `Thur.`

homura: thx for the explanation, I didn't see the word "thur" in the diff view in this PR, so this word was appended to whitelist because it is used in `neuron-ui`, right?

homura: Ok, I've found it in i18n file

WhiteMinds (approved):

PR #2744, opened 2023-06-27
fix: the .prettierignore file did not take effect during precommit

WhiteMinds:
When Lerna executes package scripts, it uses the package directory as the execution directory.
However, Prettier only reads the `.prettierignore` file in the execution directory, so the root-level `.prettierignore` file is not effective when running `npx lerna run --no-bail --stream precommit`.

Since VSCode only reads the `.prettierignore` file in the opened folder, the file must exist in the root directory. Therefore, the current solution is adapted from https://github.com/prettier/prettier/issues/4081#issuecomment-1157180927.

### DIFF
--- a/packages/neuron-ui/package.json
+++ b/packages/neuron-ui/package.json
@@ -29,12 +29,12 @@
   },
   "lint-staged": {
     "src/**/*.{js,cjs,mjs,jsx,ts,tsx}": [
-      "prettier --write",
+      "prettier --ignore-path ../../.prettierignore --write",
       "eslint --fix",
       "git add"
     ],
     "src/**/*.{css,scss}": [
-      "prettier --write",
+      "prettier --ignore-path ../../.prettierignore --write",
       "git add"
     ]
   },
--- a/packages/neuron-wallet/package.json
+++ b/packages/neuron-wallet/package.json
@@ -31,7 +31,7 @@
   },
   "lint-staged": {
     "src/**/*.{js,cjs,mjs,jsx,ts,tsx}": [
-      "prettier --write",
+      "prettier --ignore-path ../../.prettierignore --write",
       "eslint --fix",
       "git add"
     ]


yanguoyu (approved):

Keith-CY:
> When Lerna executes package scripts, it uses the package directory as the execution directory. However, Prettier only reads the `.prettierignore` file in the execution directory, so the root-level `.prettierignore` file is not effective when running `npx lerna run --no-bail --stream precommit`.
> 
> Since VSCode only reads the `.prettierignore` file in the opened folder, the file must exist in the root directory. Therefore, the current solution is adapted from [prettier/prettier#4081 (comment)](https://github.com/prettier/prettier/issues/4081#issuecomment-1157180927).

Is it feasible to `run prettier in the root directory` instead of `run prettier by lerna in the package directories`

WhiteMinds:
> Is it feasible to `run prettier in the root directory` instead of `run prettier by lerna in the package directories`

It can be a bit complex, because the current process is:
```
git commit -> .git/hooks/pre-commit -> .husky/pre-commit -> npx lerna run --no-bail --stream precommit ->
[multiple package parallel] package:precommit -> package:lint-staged -> prettier & eslint.
```

To achieve the goal, you need to extract prettier from lint-staged and move it to .husky/pre-commit.

Keith-CY:
> > Is it feasible to `run prettier in the root directory` instead of `run prettier by lerna in the package directories`
> 
> It can be a bit complex, because the current process is:
> 
> ```
> git commit -> .git/hooks/pre-commit -> .husky/pre-commit -> npx lerna run --no-bail --stream precommit ->
> [multiple package parallel] package:precommit -> package:lint-staged -> prettier & eslint.
> ```
> 
> To achieve the goal, you need to extract prettier from lint-staged and move it to .husky/pre-commit. This will make the management of prettier isolated.

I can understand all except 
> This will make the management of prettier isolated.
What does `isolated` mean if `prettier` is executed in the root directory

WhiteMinds:
> What does `isolated` mean if `prettier` is executed in the root directory

What I mean is that it will no longer be a part of lint-staged that is based on each package's independent configuration, but rather something configured relative to the root level.

I hope that this kind of individual devops for each package should be maintained as much as possible within the package itself.

Keith-CY:
> > What does `isolated` mean if `prettier` is executed in the root directory
> 
> What I mean is that it will no longer be a part of lint-staged that is based on each package's independent configuration, but rather something configured relative to the root level.
> 
> I hope that this kind of individual devops for each package should be maintained as much as possible within the package itself.

Reasonable and let's keep prettier independent in packages

Keith-CY (approved):

homura (approved):
LGTM

PR #2755, opened 2023-07-02
ci: fix api of github-scripts

Keith-CY:
A breaking change was introduced in github-scripts@v5 that all rest apis are moved from top-level to top-level.rest.

The breaking change in v5 was not noticed when the script was upgraded from v4 to v6

Ref: https://github.com/octokit/plugin-rest-endpoint-methods.js/releases/tag/v5.0.0

---

Before:
image

After:
image

### DIFF
--- a/.github/workflows/add-replied-label.yml
+++ b/.github/workflows/add-replied-label.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           script: |
             try{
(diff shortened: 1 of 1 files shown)


devchenyan (approved):

homura (approved):
LGTM

JeffreyMa597 (approved):

PR #2765, opened 2023-07-08
Fix asset accounts

devchenyan:
fix issue: 

- https://github.com/Magickbase/neuron-public-issues/issues/208

### DIFF
--- a/packages/neuron-ui/src/components/PricePanel/index.tsx
+++ b/packages/neuron-ui/src/components/PricePanel/index.tsx
@@ -3,13 +3,13 @@ import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { localNumberFormatter, shannonToCKBFormatter } from 'utils'
 import RingProgressBar from 'widgets/RingProgressBar'
 import TextField from 'widgets/TextField'
-import { Transfer } from 'widgets/Icons/icon'
+import { ReactComponent as Change } from 'widgets/Icons/Change.svg'
 import DropdownWithCustomRender from 'widgets/DropdownWithCustomRender'
 import useGetCountDownAndFeeRateStats from 'utils/hooks/useGetCountDownAndFeeRateStats'
 import { useTranslation } from 'react-i18next'
 import { appState, useState as useGlobalState } from 'states'
-
 import { FeeRateValueArrayItemType, useGetBatchGeneratedTx } from 'components/Send/hooks'
+import { batchGenerateExperimental } from 'components/SUDTSend/hooks'
 
 import styles from './pricePanel.module.scss'
 
@@ -17,6 +17,7 @@ interface PricePanelProps {
   field: string
   price: string
   onPriceChange: (value: string) => void
+  isExperimental?: boolean
 }
 enum PriceTypeEnum {
   Custom = 'custom',
@@ -25,7 +26,12 @@ enum PriceTypeEnum {
 const DEFAULT_PRICE_ARRAY = ['1000', '2000', '3000']
 const DEFAULT_COUNT_DOWN = 30
 
-const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, onPriceChange }: PricePanelProps) => {
+const PricePanel: React.FunctionComponent<PricePanelProps> = ({
+  price,
+  field,
+  onPriceChange,
+  isExperimental,
+}: PricePanelProps) => {
   const [t] = useTranslation()
   const [type, setType] = useState<PriceTypeEnum>(PriceTypeEnum.Standard)
   const [feeRateValueArray, setFeeRateValueArray] = useState<FeeRateValueArrayItemType[]>([])
@@ -35,7 +41,9 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
   const {
     app: { send = appState.send },
     wallet: { id: walletID = '' },
+    experimental,
   } = useGlobalState()
+
   const { countDown, suggestFeeRate } = useGetCountDownAndFeeRateStats({ seconds: DEFAULT_COUNT_DOWN })
 
   const isStandard = type === PriceTypeEnum.Standard
@@ -108,10 +116,16 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
   )
 
   useEffect(() => {
-    useGetBatchGeneratedTx({ walletID, items: send.outputs, priceArray }).then(res => {
-      setFeeRateValueArray(res)
-    })
-  }, [send.outputs, priceArray])
+    if (isExperimental && experimental) {
+      batchGenerateExperimental(experimental, priceArray).then(res => {
+        setFeeRateValueArray(res)
+      })
+    } else {
+      useGetBatchGeneratedTx({ walletID, items: send.outputs, priceArray }).then(res => {
+        setFeeRateValueArray(res)
+      })
+    }
+  }, [send.outputs, priceArray, isExperimental, experimental, batchGenerateExperimental, setFeeRateValueArray])
 
   useEffect(() => {
     if (suggestFeeRate === 0) {
@@ -141,7 +155,7 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
             onKeyDown={() => {}}
             type="button"
           >
-            <Transfer />
+            <Change />
           </button>
         </div>
         {isStandard ? (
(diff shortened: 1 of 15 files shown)


packages/neuron-ui/src/components/SUDTSend/index.tsx
```diff
@@ -386,35 +399,42 @@ const SUDTSend = () => {
...
-                                  showTriangle
-                                >
-                                  <AttentionOutline className={styles.attention} />
-                                </Tooltip>
-                              </div>
-                            ) : null,
-                            tip:
-                              item.key === SendType.secp256Cheque && !isMainnet ? (
-                                <div className={styles.selectError}>{t('messages.light-client-cheque-warning')}</div>
+                        {options.length > 1 ? (
```
yanguoyu, line 407: IMO, if `options`.length is 1, we can also use `RadioGroup`.

packages/neuron-ui/src/components/TransactionFeePanel/index.tsx
```diff
@@ -7,19 +7,21 @@ interface TransactionFeeProps {
...
   onPriceChange,
+  isExperimental,
 }: TransactionFeeProps) => {
   const [t] = useTranslation()
 
   return (
     <div>
       <TextField label={t('send.fee')} field="fee" value={`${fee} CKB`} readOnly disabled width="100%" />
-      <PricePanel field="price" price={price} onPriceChange={onPriceChange} />
+      <PricePanel field="price" price={price} onPriceChange={onPriceChange} isExperimental={isExperimental} />
```
yanguoyu, line 24: IMO, the experimental will not affect `PricePanel`

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -107,11 +115,35 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
...
+    if (experimental && experimental?.params) {
+      const generator = getGenerator(experimental.params?.sendType)
+      try {
+        const requestArray = priceArray.map(itemPrice => generator({ ...experimental.params, feeRate: itemPrice }))
+        const allPromiseResult = await Promise.allSettled(requestArray)
+
+        const resList = allPromiseResult.map((batchItem, index: number) => ({
+          feeRateValue: priceArray[index],
+          // @ts-ignore
+          feeValue: batchItem.value.result?.fee,
```
WhiteMinds, line 128:
```typescript
interface PromiseFulfilledResult<T> {
    status: "fulfilled";
    value: T;
}

interface PromiseRejectedResult {
    status: "rejected";
    reason: any;
}

type PromiseSettledResult<T> = PromiseFulfilledResult<T> | PromiseRejectedResult;
```

The error here is because there is indeed no value in `PromiseRejectedResult`, and `ts-ignore` should not be used to break the type system.
If you think it is impossible to have an error, then you should use `Promise.all` instead of `Promise.allSettled`.
If you think there may be errors, but you want to ignore them, then create an `isFulfilled` utility function and do a `.filter(isFulfilled)` before `.map`.

```typescript
const isFulfilled = <T extends any>(input: PromiseSettledResult<T>): input is PromiseFulfilledResult<T> =>
          input.status === 'fulfilled'
```

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -107,11 +115,35 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
...
+        }))
+
+        setFeeRateValueArray(resList)
+      } catch (error) {
+        // @ts-ignore
```
WhiteMinds, line 133: I think there should be an explanation for any place that swallows errors, even if it's just saying "this error is not important".

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -107,11 +115,35 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
...
-    })
-  }, [send.outputs, priceArray])
+    if (isExperimental) {
+      batchGenerateExperimental()
+    } else {
+      useGetBatchGeneratedTx({ walletID, items: send.outputs, priceArray }).then(res => {
+        setFeeRateValueArray(res)
+      })
+    }
+  }, [send.outputs, priceArray, isExperimental, experimental])
```
WhiteMinds, line 146:
```suggestion
  }, [send.outputs, priceArray, isExperimental, batchGenerateExperimental])
```

WhiteMinds (changes requested):

packages/neuron-ui/src/components/PricePanel/index.tsx
```diff
@@ -107,11 +115,35 @@ const PricePanel: React.FunctionComponent<PricePanelProps> = ({ price, field, on
     [priceArray, feeValuesArray]
   )
 
+  const batchGenerateExperimental = useCallback(async () => {
+    if (experimental && experimental?.params) {
```
JeffreyMa597, line 119:
```suggestion
    if (experimental?.params) {
```

packages/neuron-ui/src/components/SUDTSend/hooks.ts
```diff
@@ -152,6 +160,26 @@ export function getGenerator(sendType?: SendType) {
...
 
+export async function batchGenerateExperimental(
+  experimental: { tx: any; assetAccount?: any; params?: any },
+  priceArray: string[]
+) {
+  if (experimental?.params) {
+    const generator = getGenerator(experimental?.params?.sendType)
+    const requestArray = priceArray.map(itemPrice => generator({ ...experimental.params, feeRate: itemPrice }))
+    const allPromiseResult = await Promise.allSettled(requestArray)
+    const resList = (allPromiseResult as GenerateResult[]).map((batchItem, index: number) => ({
```
WhiteMinds, line 171: `as` is like `ts-ignore` and `any`, all of which break the type system.
This is not better than `ts-ignore`, because at least `ts-ignore` clearly indicates that the type system is being broken.
Please refer to this approach: https://github.com/nervosnetwork/neuron/pull/2765#discussion_r1257826045 .

WhiteMinds (dismissed):

packages/neuron-ui/src/components/SUDTSend/index.tsx
devchenyan: The single option needs to support unchecking, so it need to use the checkbox

JeffreyMa597 (approved):

packages/neuron-ui/src/states/stateProvider/reducer.ts
```diff
@@ -93,7 +93,7 @@ export type StateAction =
   | { type: AppActions.ToggleAllNotificationVisibility; payload?: boolean }
   | { type: AppActions.ToggleIsAllowedToFetchList; payload?: boolean }
   | { type: AppActions.Ignore; payload?: any }
-  | { type: AppActions.UpdateExperimentalParams; payload: { tx: any; assetAccount?: any } | null }
+  | { type: AppActions.UpdateExperimentalParams; payload: { tx: any; assetAccount?: any; params?: any } | null }
```
yanguoyu, line 96: We should try our best to avoid using any.

packages/neuron-ui/src/types/App/index.d.ts
```diff
@@ -289,7 +289,7 @@ declare namespace State {
     nervosDAO: NervosDAO
     updater: AppUpdater
     sUDTAccounts: SUDTAccount[]
-    experimental: { tx: any; assetAccount?: any } | null
+    experimental: { tx: any; assetAccount?: any; params?: any } | null
```
yanguoyu, line 292: We should try our best to avoid using any.

packages/neuron-ui/src/types/App/index.d.ts
```diff
@@ -281,6 +281,29 @@ declare namespace State {
     errorMsg: string
   }
 
+  enum SendType {
```
yanguoyu, line 284: After `SendType` define here, then we can remove https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/SUDTSend/hooks.ts#L8-L15. And use `State.SendType` to replace `SendType`.

silySuper:
please help to push this review @devchenyan @yanguoyu

yanguoyu:
> please help to push this review @devchenyan @yanguoyu

@devchenyan  There are still some unresolved conversations.

packages/neuron-ui/src/types/App/index.d.ts
devchenyan: `enum` is not a pure type, `SendType` cannot be used directly through `State.SendType`

yanguoyu: Gotcha

yanguoyu: Gotcha

silySuper (approved):

WhiteMinds (approved):

PR #2769, opened 2023-07-10
feat:Detect new upgrade on launch

JeffreyMa597:
------
As title.

https://github.com/nervosnetwork/neuron/assets/22511289/9d6b27d7-b047-43e9-a90a-9d88a814091a

Ref: https://github.com/Magickbase/neuron-public-issues/issues/139

Check List
------

### Test
e2e Test

### DIFF
--- a/packages/neuron-ui/src/components/GeneralSetting/generalSetting.module.scss
+++ b/packages/neuron-ui/src/components/GeneralSetting/generalSetting.module.scss
@@ -12,6 +12,7 @@ $action-button-width: 11.25rem;
     background: var(--input-disabled-color);
     border-radius: 8px;
     margin-right: 16px;
+
     p {
       font-size: 14px;
       line-height: 20px;
@@ -31,6 +32,25 @@ $action-button-width: 11.25rem;
       svg {
         margin-right: 4px;
       }
+    }
+  }
+  .showVersion {
+    position: relative;
+
+    &::after {
+      content: attr(data-new-version-tip);
+      background-color: #ff1e1e;
+      border-top-right-radius: 8px;
+      border-bottom-left-radius: 8px;
+      position: absolute;
+      top: 0;
+      right: 0;
+      font-size: 12px;
+      font-style: normal;
+      font-weight: 500;
+      line-height: normal;
+      padding: 2px 4px;
+      color: #fff;
     }
   }
 }
--- a/packages/neuron-ui/src/components/GeneralSetting/index.tsx
+++ b/packages/neuron-ui/src/components/GeneralSetting/index.tsx
@@ -14,13 +14,13 @@ import {
   installUpdate,
   getVersion,
 } from 'services/remote'
-import { uniformTimeFormatter, bytesFormatter } from 'utils'
+import { uniformTimeFormatter, bytesFormatter, clsx } from 'utils'
 import { LanguageSelect } from 'widgets/Icons/icon'
 import styles from './generalSetting.module.scss'
 
 interface UpdateDownloadStatusProps {
   show: boolean
-  onCancel: () => void
+  onCancel: (status?: string) => void
   progress: number
   newVersion: string
   releaseDate: string
@@ -69,7 +69,7 @@ const UpdateDownloadStatus = ({
         onConfirm={handleConfirm}
         disabled={!available}
         confirmText={t('updates.install-update')}
-        onCancel={onCancel}
+        onCancel={() => onCancel('checked')}
         title={t('updates.update-available')}
         confirmProps={{
           'data-method': 'download',
@@ -135,7 +135,9 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
   const [showLangDialog, setShowLangDialog] = useState(false)
   const [searchParams] = useSearchParams()
   const [errorMsg, setErrorMsg] = useState('')
-  const [dialogType, setDialogType] = useState<'' | 'checking' | 'updating' | 'updated'>('')
+  const [dialogType, setDialogType] = useState<'' | 'checking' | 'checked' | 'updating' | 'updated'>('')
+  const [isFetchUpdateByClick, setIsFetchUpdateByClick] = useState<boolean>(false)
+  const { version: newVersion } = updater
 
   const version = useMemo(() => {
     return getVersion()
@@ -145,7 +147,9 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
     const checkUpdate = searchParams.get('checkUpdate')
     if (checkUpdate === '1') {
       checkForUpdates()
+      setIsFetchUpdateByClick(true)
     }
+    return () => setIsFetchUpdateByClick(false)
   }, [searchParams, checkForUpdates])
 
   useEffect(() => {
@@ -162,7 +166,11 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
       setDialogType('checking')
       return
     }
-    if (updater.version || updater.downloadProgress > 0) {
+    if (newVersion) {
+      setDialogType('checked')
+      return
+    }
+    if (updater.downloadProgress > 0) {
       setDialogType('updating')
       return
     }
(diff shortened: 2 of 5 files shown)


yanguoyu:
It will open a dialog like that if there is no update when I open the setting page.

image

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -189,12 +198,12 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
 
   return (
     <div className={styles.container}>
-      <div className={styles.content}>
+      <div className={clsx(styles.content, `${newVersion ? styles.showVersion : ''}`)} data-new-version-tip="New">
         <p>
-          {t('settings.general.version')} v{version}
+          {t('settings.general.version')} v{newVersion || version}
```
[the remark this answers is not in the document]
JeffreyMa597: @yanguoyu It mentioned that we should mark it as the new version.
https://github.com/Magickbase/neuron-public-issues/issues/139#issuecomment-1617464081

packages/neuron-ui/src/components/GeneralSetting/generalSetting.module.scss
```diff
@@ -33,6 +34,26 @@ $action-button-width: 11.25rem;
...
+
+    &::after {
+      content: attr(data-new-version-tip);
+      background-color: #ff1e1e;
+      border-top-right-radius: 8px;
+      border-bottom-left-radius: 8px;
+      position: absolute;
+      top: 0;
+      right: 0;
+      font-family: PingFang SC;
```
Keith-CY, line 48: This font is Apple-specific

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -162,7 +164,11 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
       setDialogType('checking')
       return
     }
-    if (updater.version || updater.downloadProgress > 0) {
+    if (updater.version) {
```
Keith-CY, line 167: `newVersion` could be used because it's assigned as `updater.version`

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -135,7 +135,9 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
   const [showLangDialog, setShowLangDialog] = useState(false)
   const [searchParams] = useSearchParams()
   const [errorMsg, setErrorMsg] = useState('')
-  const [dialogType, setDialogType] = useState<'' | 'checking' | 'updating' | 'updated'>('')
+  const [dialogType, setDialogType] = useState<'' | 'checking' | 'checked' | 'updating' | 'updated'>('')
```
Keith-CY, line 138: The new type `checked` can be avoided by `const isChecked = !!updater.version`

packages/neuron-ui/src/widgets/Dot/dot.module.scss
```diff
@@ -0,0 +1,16 @@
+@import '../../styles/theme.scss';
+
+.dot {
```
Keith-CY, line 3: Does `dot` mean a simple badge as https://m3.material.io/components/badges/overview
image

Keith-CY:
> It will open a dialog like that if there is no update when I open the setting page. @JeffreyMa597
> 
> image
> And if there is no update when I open Neuron, but after a while, there exists an update version, when I click `check for update` on the setting page, the dialog will close after checking the update. @JeffreyMa597
> 
>  exception.mov 
> On the other hand, do we only check once when we open Neuron or should we check for updates every some intervals? @Danie0918

I would suggest checking new updates daily because some may keep Neuron open because it synchronizes slowly. The case may be rare but the cost is low

packages/neuron-ui/src/widgets/Dot/dot.module.scss
JeffreyMa597: > Does `dot` mean a simple badge as https://m3.material.io/components/badges/overview

I think it's the same as a badge. And I try to avoid using external UI component libraries.

packages/neuron-ui/src/components/GeneralSetting/generalSetting.module.scss
JeffreyMa597: > This font is Apple-specific

image

I'll ask our designer for help to handle it.

packages/neuron-ui/src/widgets/Dot/dot.module.scss
Keith-CY: Ah I'm not saying the standalone implementation is not good, the point is that the name is not straight, `badge` or something is more clear than `dot`

Danie0918:
> > It will open a dialog like that if there is no update when I open the setting page. @JeffreyMa597
> > image
> > And if there is no update when I open Neuron, but after a while, there exists an update version, when I click `check for update` on the setting page, the dialog will close after checking the update. @JeffreyMa597
> > exception.mov
> > On the other hand, do we only check once when we open Neuron or should we check for updates every some intervals? @Danie0918
> 
> I would suggest checking new updates daily because some may keep Neuron open if it synchronizes slowly. The case may be rare but the cost is low

Yes, it is true that there may be users who keep Neuron open. So here we need to add a daily check in addition to the open check, how about we set it to UTC+0 check?

packages/neuron-ui/src/widgets/Dot/dot.module.scss
JeffreyMa597: > Ah I'm not saying the standalone implementation is not good, the point is that the name is not straight, `badge` or something is more clear than `dot`

Oh, got it.

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -237,9 +246,12 @@ const GeneralSetting = ({ updater }: GeneralSettingProps) => {
       </Dialog>
 
       <UpdateDownloadStatus
-        show={dialogType === 'updating'}
-        onCancel={() => {
-          cancelDownloadUpdate()
+        show={dialogType === 'updating' || (dialogType === 'checked' && isClickedFetchUpdate)}
+        onCancel={status => {
+          if (status !== 'checked') {
```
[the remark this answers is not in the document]
JeffreyMa597: > Reference to https://github.com/nervosnetwork/neuron/pull/2769/files#diff-8a655605bf3e06e1e9ca7a9f3eba06c222f722b732059db07848e7fc66551c56R72,
> the status is always `checked`, I'm confused about it.

As you see, this method is a common method to handle `cancel` action. So I gave it a property to distinguish between different cases.
image

packages/neuron-ui/src/components/GeneralSetting/index.tsx
JeffreyMa597: > The new type `checked` can be avoided by `const isChecked = !!updater.version`

In the logic of the function `handleUpdate`, there needs to a key to open our installation dialog, and on the other hand, we don't want the installation dialog to open as soon as we enter the page.

packages/neuron-ui/src/components/GeneralSetting/index.tsx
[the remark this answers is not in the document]
JeffreyMa597: This function will reset the state of the version when we just cancel installing the new version in the `pre-install` dialog. IMO, it's unnecessary for us to fetch the new data again. Also, the display on our page is not expected.

image

yanguoyu:
> how about we set it to UTC+0 check?

If we check it at UTC+0, we should check whether the current is UTC+0 every period of time. How about checking it every period of time?

Danie0918:
> > how about we set it to UTC+0 check?
> 
> If we check it at UTC+0, we should check whether the current is UTC+0 every period of time. How about checking it every period of time?

OK, set up a 24-hour check.

yanguoyu:
Close with https://github.com/nervosnetwork/neuron/pull/2783

PR #2773, opened 2023-07-17
chore: Add action to update assume valid target

yanguoyu:
Reference to https://github.com/Magickbase/neuron-public-issues/issues/211.

https://github.com/yanguoyu/neuron/pull/82 is the PR create by action triggered by https://github.com/yanguoyu/neuron/pull/72 ready for review.

### DIFF
--- a/.github/workflows/update_valid_target.yml
+++ b/.github/workflows/update_valid_target.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Write env file
         run: |
-          npm run update:valid:target
+          npm run update:valid-target
 
       - name: Commit env file
         uses: actions/github-script@v6
--- a/.github/workflows/update_valid_target.yml
+++ b/.github/workflows/update_valid_target.yml
@@ -1,0 +1,86 @@
+name: Update ckb node assume valid target
+
+on:
+  pull_request:
+    types: [ready_for_review]
+    branches:
+      - master
+
+jobs:
+  ready-for-release:
+    name: Update ckb node assume valid target
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create Branch
+        uses: peterjgrainger/action-create-branch@v2.2.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          branch: 'chore-update-valid-target/${{github.head_ref}}'
+          sha: '${{ github.event.pull_request.head.sha }}'
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: 'chore-update-valid-target/${{github.head_ref}}'
+      
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.12.0
+
+      - name: Write env file
+        run: |
+          npm run update:valid:target
+
+      - name: Commit env file
+        uses: actions/github-script@v6
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BASE: ${{ github.head_ref }}
+        with:
+          script: |
+            const fs = require('node:fs')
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -35,7 +35,8 @@
     "test:ci": "yarn build:main && yarn test",
     "lint": "lerna run --stream lint",
     "postinstall": "husky install",
-    "db:chain": "node ./node_modules/.bin/typeorm"
+    "db:chain": "node ./node_modules/.bin/typeorm",
+    "update:valid:target": "node ./scripts/update-valid-target.js"
```
Keith-CY, line 39: What if naming it `update:valid-target` to keep `valid target` in the save level

scripts/update-valid-target.js
```diff
@@ -0,0 +1,64 @@
+const https = require('node:https')
```
Keith-CY, line 1: `fetch` is supported natively since Node.js@18

Ref:
- node.js engine specification: https://github.com/nervosnetwork/neuron/blob/develop/package.json#L18
- compatibility: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility

.github/workflows/update_valid_target.yml
```diff
@@ -0,0 +1,86 @@
+name: Update ckb node assume valid target
+
+on:
+  pull_request:
+    types: [ready_for_review]
+    branches:
+      - master
```
homura, line 7: Would it be better to trigger by a cronjob instead of every PR? If every PR triggers this action, it will make a lot of bot PRs and clutter the PR list.

homura: I misread it, I found the development is at `develop` branch and only push to `master` when release, so it makes sense to have this trigger strategy like this

.github/workflows/update_valid_target.yml
```diff
@@ -0,0 +1,86 @@
...
+            if (pulls.some(pull => pull.head.ref === HEAD)) {
+              return
+            }
+            github.rest.pulls.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: HEAD,
+              base: BASE,
+              title: 'chore: Update ckb node assume valid target',
+              body: `This PR uses to update ckb node assume valid target for PR https://github.com/yanguoyu/neuron/pull/${context.issue.number}`,
```
Keith-CY, line 85: The message should refer to the `refs` of this action instead of a hard-coded one

Keith-CY (approved):

WhiteMinds (approved):

scripts/update-valid-target.js
```diff
@@ -0,0 +1,34 @@
...
+const envFilePath = path.resolve(__dirname, '../packages/neuron-wallet/.env')
+const validTargetReg = /(CKB_NODE_ASSUME_VALID_TARGET=)[\S]*/
+
+;(async function () {
+  const tipBlockNumber = (await rpcRequest('get_tip_block_number')).result
+  const validTargetBlockNumber = `0x${(BigInt(tipBlockNumber) - BigInt(ESTIMATE_BLOCK_COUNT_PER_DAY)).toString(16)}`
+  const blockHash = (await rpcRequest('get_block_hash', [validTargetBlockNumber])).result
+  const originEnvContent = fs.readFileSync(envFilePath).toString('utf-8')
+  fs.writeFileSync(envFilePath, originEnvContent.replace(validTargetReg, `CKB_NODE_ASSUME_VALID_TARGET='${blockHash}'`))
+})()
```
WhiteMinds, line 34: node support `top level await` since v14

yanguoyu: `top level await` should run with `esm`.

WhiteMinds: Oh, I forgot about this.

devchenyan (approved):

PR #2774, opened 2023-07-18
chore(CI): Use pr branch for checkout.

yanguoyu:
https://github.com/yanguoyu/neuron/pull/85#issuecomment-1639160771 is the result that runs by checking out the PR branch with the PR number.

### DIFF
--- a/.github/workflows/package_for_test.yml
+++ b/.github/workflows/package_for_test.yml
@@ -53,19 +53,11 @@ jobs:
         uses: actions/checkout@v3
         if: ${{ github.event_name == 'push' }}
 
-      - name: Get PR Branch
-        if: github.event_name == 'issue_comment'
-        id: comment-branch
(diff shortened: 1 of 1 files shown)


Keith-CY:
Does it resolve the problem that `package should be built on the head branch instead of the base branch`?

yanguoyu:
> Does it resolve the problem that `package should be built on the head branch instead of the base branch`?

Sure, here is the checkout action log for checking out pr number.
```
Fetching the repository
/usr/local/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/pull/85/merge:refs/remotes/pull/85/merge

Checking out the ref

/usr/local/bin/git checkout --progress --force refs/remotes/pull/85/merge
```

Keith-CY (approved):

devchenyan (approved):

PR #2787, opened 2023-07-27
Just for test update neuron

yanguoyu:


silySuper:
If I cancel update in midway ,what the UI will show ?If show old version ,how to operate to show new icon again? @Danie0918 @Kirl170

Danie0918:
> If I cancel update in midway ,what the UI will show ?If show old version ,how to operate to show new icon again? @Danie0918 @Kirl170

Canceling an update should revert to the previous state, which is to show the new version.

PR #2791, opened 2023-07-29
fix: issues - 231、224、244、225、223、228、229

devchenyan:
fix:
- https://github.com/Magickbase/neuron-public-issues/issues/224
- https://github.com/Magickbase/neuron-public-issues/issues/231
- https://github.com/Magickbase/neuron-public-issues/issues/244 
- https://github.com/Magickbase/neuron-public-issues/issues/225 
- https://github.com/Magickbase/neuron-public-issues/issues/223
- https://github.com/Magickbase/neuron-public-issues/issues/228
- https://github.com/Magickbase/neuron-public-issues/issues/229


packages/neuron-ui/src/components/SUDTAccountPile/index.tsx
```diff
@@ -35,6 +36,16 @@ const SUDTAccountPile = ({
...
+  const balanceText = sudtValueToAmount(balance, decimal)
+  let overBalanceText = ''
+
+  if (Number(decimal) > 8) {
+    const [integer, radix] = balanceText.split('.')
+    if (radix && radix.length > 8) {
+      overBalanceText = `${integer}.${radix.slice(0, 8)}`
+    }
+  }
+
```
yanguoyu, line 48: Here can use `useMemo` to cache the calculate.

packages/neuron-ui/src/states/stateProvider/actionCreators/wallets.ts
```diff
@@ -43,6 +42,9 @@ export const updateWalletList = () => (dispatch: StateDispatch) => {
     if (isSuccessResponse(res)) {
       const payload = res.result || []
       dispatch({ type: NeuronWalletActions.UpdateWalletList, payload })
+      if (walletsCache.load().length < payload.length) {
```
yanguoyu, line 45: It's better to move this logic to https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/utils/hooks/index.ts#L489-L497.

packages/neuron-ui/src/components/SpecialAssetList/index.tsx
```diff
@@ -217,16 +219,17 @@ const SpecialAssetList = () => {
     (item: SpecialAssetCell) => {
       const { timestamp, customizedAssetInfo, capacity, lock, type, data } = item
 
-      return useGetSpecialAssetColumnInfo({
+      return getSpecialAssetColumnInfo({
```
yanguoyu, line 222: `getSpecialAssetColumnInfo` is not a hook function, could we move 
```typescript
useCallback(
    (item: SpecialAssetCell) => {
      const { timestamp, customizedAssetInfo, capacity, lock, type, data } = item

      return getSpecialAssetColumnInfo({
        cell: { capacity, lock, type, data },
        datetime: +timestamp,
        epoch,
        assetInfo: customizedAssetInfo,
        bestKnownBlockTimestamp,
        tokenInfoList,
      })
    },
    [epoch, bestKnownBlockTimestamp, tokenInfoList, useGetSpecialAssetColumnInfo]
```

to hooks file?

packages/neuron-ui/src/states/stateProvider/actionCreators/wallets.ts
devchenyan: Because the message is also displayed when a new wallet is added, it can't just be judged in `useFirstLoadApp`.

silySuper:
@devchenyan help to push this review and confirm whether the latest package can be tested.

packages/neuron-ui/src/states/stateProvider/actionCreators/wallets.ts
yanguoyu: May be we can rename `useFirstLoadApp` to `useFirstLoadWallet`

packages/neuron-ui/src/components/SpecialAssetList/hooks.ts
```diff
@@ -15,9 +14,33 @@ import {
   useDialogWrapper,
   useDidMount,
 } from 'utils'
+import { TFunction } from 'i18next'
 import { MILLISECONDS, MILLISECONDS_PER_DAY } from 'utils/const'
 import { AssetInfo, ChequeAssetInfo, NFTType } from '.'
 
+interface SpecialAssetCell {
```
WhiteMinds, line 21: I see there is also a definition of this type in `SpecialAssetList/index.tsx`.
Should we consider having `index.tsx` import this type from `hooks.ts`?

WhiteMinds (approved):

FrederLu (approved):

packages/neuron-ui/src/utils/hooks/index.ts
```diff
@@ -486,14 +486,13 @@ export const useOutputErrors = (
   )
 }
 
-export const useFirstLoadApp = (dispatch: StateDispatch) => {
+export const useFirstLoadWallet = (dispatch: StateDispatch, num: number) => {
   useEffect(() => {
-    const isFirstLoad = firstLoadApp.load()
-    if (isFirstLoad) {
+    if (num > Number(walletNumber.load())) {
```
yanguoyu, line 491: How about saving `walletID` to record wallet has loaded for every wallet?

yanguoyu (approved):

PR #2795, opened 2023-08-01
fix: Fix DAO styles.

yanguoyu:
The fixed styles refer to https://github.com/Magickbase/neuron-public-issues/issues/246


yanguoyu:
@devchenyan @WhiteMinds @Keith-CY @homura Please help me review it.

packages/neuron-ui/src/components/NervosDAODetail/CellsCard/cellsCard.module.scss
```diff
@@ -92,7 +89,32 @@
...
+      .scriptTag:hover {
+        color: var(--primary-color);
+      }
+    }
+  }
+
+  .copyTableAddress {
+    word-break: break-all;
+    text-align: center;
+    font-family: 'JetBrains Mono';
```
Keith-CY, line 110: The font family appears multiple times with its literal value, maybe using a mixin helps.

packages/neuron-ui/src/components/HistoryDetailPage/index.tsx
```diff
@@ -202,9 +240,14 @@ const HistoryDetailPage = () => {
       dataIndex: 'amount',
       align: 'left',
       isBalance: true,
+      className: styles.amount,
       render(_, __, item, show: boolean) {
         const { capacity } = handleListData(item)
-        return show ? <CopyZone content={capacity.replace(/,/g, '')}>{`${capacity} CKB`}</CopyZone> : HIDE_BALANCE
+        return show ? (
+          <CopyZone maskRadius={8} content={capacity.replace(/,/g, '')}>{`${capacity} CKB`}</CopyZone>
```
Keith-CY, line 247: `capacity.replaceAll(',', '')` can be used now

Ref:
  - Compatibility: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility
  - Chrome in Electron: https://www.electronjs.org/blog/electron-24-0#notable-changes

packages/neuron-ui/src/components/LockInfoDialog/index.tsx
```diff
@@ -14,7 +15,38 @@ interface LockInfoDialogProps {
   onDismiss: () => void
 }
 
-const ShortAddr = ({ lockScript, isMainnet }: { lockScript: CKBComponents.Script | null; isMainnet: boolean }) => {
+const useCopy = () => {
+  const timer = useRef<ReturnType<typeof setTimeout>>()
+  const [copied, setCopied] = useState(false)
+  const [refreshKey, setRefreshKey] = useState(1)
```
Keith-CY, line 21: What does `refresh key` do, only for `key` of duplicated components?

packages/neuron-ui/src/components/NervosDAODetail/CellsCard/index.tsx
```diff
@@ -71,12 +73,36 @@ const TabsVariantWithCellsCard: FC<
...
+                      tip={
+                        <CopyZone content={address} className={styles.copyTableAddress}>
+                          {address}
+                        </CopyZone>
+                      }
+                      className={styles.addressTips}
+                      showTriangle
+                      isTriggerNextToChild
+                    >
+                      <span className={styles.address}>{`${address.slice(0, 20)}...${address.slice(-20)}`} </span>
```
homura, line 86: I recommend taking a peek at the [react-middle-ellipsis](https://github.com/bluepeter/react-middle-ellipsis) if we should truncate something in the middle in the future

packages/neuron-ui/src/components/NervosDAO/index.tsx
```diff
@@ -263,28 +272,32 @@ const NervosDAO = () => {
...
-                content={shannonToCKBFormatter(`${free}`, false, '')}
-                name={t('nervos-dao.copy-balance')}
-                className={styles.balance}
-              >
-                <span className={styles.number}>{isPrivacyMode ? HIDE_BALANCE : shannonToCKBFormatter(`${free}`)}</span>{' '}
-                CKB
-              </CopyZone>
+              {isPrivacyMode ? (
+                <>
+                  <span className={styles.number}>{HIDE_BALANCE}</span> CKB
```
homura, line 283: I find there were many `HIDE_BALANCE` in Neuron, maybe we can extract them as a component in the future

```tsx
interface Balance {
  value?: Numberish;
  hidden?: boolean;
  copyable?: boolean;
  formatter?: string; // '0,0.[0000]'
  suffix?: ReactElement; // CKB or something else
}
```

packages/neuron-ui/src/components/LockInfoDialog/index.tsx
yanguoyu: Add `key` to force to re-render the `Alert` component. Because the `Alert` has animation. Otherwise, if there is no change in the component when the user clicks copy more times in an animation execute time, the animation will not normally. I can also use `debounce` to show the last animation.

packages/neuron-ui/src/components/NervosDAODetail/CellsCard/index.tsx
yanguoyu: For the safety wallet, we will use fewer third-party libs because they may be out of control.

homura: > For the safety wallet, we will use fewer third-party libs because they may be out of control.

exactly, I forget that Neuron is a crypto wallet

packages/neuron-ui/src/components/NervosDAO/index.tsx
yanguoyu: A good suggestion, it can refactor later.

Keith-CY (approved):

devchenyan (approved):

homura (approved):

FrederLu (approved):

PR #2800, opened 2023-08-09
fix: Use alert dialog to replace default dialog

yanguoyu:


FrederLu (approved):

yanguoyu:
@devchenyan  @WhiteMinds @Keith-CY Please help me have a review.

packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -294,13 +296,17 @@ export const useSubscription = ({
         }
       }
     })
+    const showGlobalDialogSubject = ShowGlobalDialogSubject.subscribe(params => {
```
Keith-CY, line 299: I'l like to know why `show global dialog` and `show alert dialog` are paired

Keith-CY (approved):

packages/neuron-ui/src/containers/Main/hooks.ts
yanguoyu: Maybe renaming `showAlertDialog` to `showGlobalAlertDialog` is clear?

Keith-CY: It would be more clear and won't mislead other developers in the future

WhiteMinds (approved):

PR #2802, opened 2023-08-10
refactor: use lumos to calculate script hash

zhangyouxin:
## Description 

remove `scriptToHash` usage of `ckb-sdk-js`

Resolve part of: https://github.com/Magickbase/neuron-public-issues/issues/201


packages/neuron-wallet/src/models/chain/script.ts
```diff
@@ -34,7 +34,17 @@ export default class Script {
   }
 
   public computeHash(): string {
-    return scriptToHash(this.toSDK())
+    const script = this.toSDK()
+    // TODO: should throw error, skip calculate hash if parameter not enough (only in test environment)
+    if (!script || !script.codeHash || !script.hashType) {
+      return ''
+    }
```
homura, line 41: Why return an empty string here instead of throwing an error?

zhangyouxin: You are right here, I intended to throwing an error (see the TODO), but meet some obstacles:

In test environment, the `.env` file variables can't be read, for example the [SystemScriptInfo.DAO_CODE_HASH](https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/models/system-script-info.ts#L16-L20) here, causing test cases fail to run... 

Anybody can help me to get around with this problem here ? @yanguoyu @devchenyan @Keith-CY

WhiteMinds (approved):

packages/neuron-wallet/src/models/chain/script.ts
zhangyouxin: updated by [563ef4c](https://github.com/nervosnetwork/neuron/pull/2802/commits/563ef4cb4037d56721d6b1dc2954e427d40518d8)

packages/neuron-wallet/src/models/system-script-info.ts
```diff
@@ -14,9 +15,9 @@ export default class SystemScriptInfo {
   static MULTI_SIGN_HASH_TYPE = ScriptHashType.Type
 
   static DAO_SCRIPT_HASH = new Script(
-    SystemScriptInfo.DAO_CODE_HASH,
+    systemScripts.NERVOS_DAO.codeHash,
```
yanguoyu, line 18: To remove `@nervosnetwork/ckb-sdk-utils` finally, I propose we'd better use the process.env to create the DAO script.

zhangyouxin: will replace `systemScripts` with lumos in another PR

packages/neuron-wallet/src/models/chain/script.ts
yanguoyu: The env for the test case is init at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/tests/setup.ts. And the local env file will not load because load-env-file is false https://github.com/nervosnetwork/neuron/blob/develop/package.json#L34.

homura (approved):

silySuper (approved):

PR #2803, opened 2023-08-10
refactor: remove usage of serialization methods from ckb-sdk-js

zhangyouxin:
## Description 

remove `serializeXXX` usage of `ckb-sdk-js`

Resolve part of: https://github.com/Magickbase/neuron-public-issues/issues/201


homura (approved):

yanguoyu (approved):

packages/neuron-wallet/tests/services/tx/transaction-generator.test.ts
```diff
@@ -1196,7 +1197,7 @@ describe('TransactionGenerator', () => {
           const targetOutput: Output = Output.fromObject({
             capacity: toShannon('61'),
             lock: aliceAnyoneCanPayLockScript,
-            type: assetAccountInfo.generateSudtScript('0xuuid'),
+            type: assetAccountInfo.generateSudtScript('0x1234'),
```
Keith-CY, line 1200: Does this change matter in this refactor?

packages/neuron-wallet/tests/models/transaction-size.test.ts
```diff
@@ -16,7 +16,7 @@ describe('TransactionSize', () => {
     type: new Script('0xece45e0979030e2f8909f76258631c42333b1e906fd9701ec3600a464a90b8f6', '0x', ScriptHashType.Data),
   })
 
-  const witnessArgs = new WitnessArgs('', '0x', '')
+  const witnessArgs = new WitnessArgs(undefined, '0x', undefined)
```
Keith-CY, line 19: `''` should be handled too because we don't know if there are `''`s passed to this method.

packages/neuron-wallet/tests/services/tx/transaction-generator.test.ts
zhangyouxin: Yes, the `0xuuid` is not a valid hex string, `lumos` serilization will check it and throw `invalid hex string` error.

Keith-CY: Got it.

packages/neuron-wallet/tests/models/transaction-size.test.ts
zhangyouxin: updated: https://github.com/nervosnetwork/neuron/pull/2803/commits/b50a7edaccf69264bffb28295bcf47a466822cb3

Keith-CY (approved):

WhiteMinds (approved):

silySuper (approved):

PR #2808, opened 2023-08-16
feat: Add check ckb node update or compatibility

yanguoyu:
All of the scenes should start ckb external.
1. When Neuron has an updated version, and the running external ckb version is bigger than Neuron's internal ckb version
This scene can test with https://github.com/nervosnetwork/neuron/pull/2807, the external ckb version should be bigger than v0.109.0
image

2. When the running external ckb version is not compatible with Neuron.
This scene can be tested with the current pr package, and the external ckb version should not be 0.109 ~ 0.110
image

### DIFF
--- a/compatible.csv
+++ b/compatible.csv
@@ -1,5 +1,5 @@
 CKB,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
-Neuron
+Neuron,,,,,,,,
 0.110,yes,yes,no,no,no,no,no,no
 0.106,no,no,yes,yes,yes,yes,no,no
 0.103,no,no,no,no,no,no,yes,yes
--- a/compatible.csv
+++ b/compatible.csv
@@ -1,0 +1,5 @@
+CKB,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
+Neuron
+0.110,yes,yes,no,no,no,no,no,no
+0.106,no,no,yes,yes,yes,yes,no,no
+0.103,no,no,no,no,no,no,yes,yes
--- a/packages/neuron-ui/src/services/remote/app.ts
+++ b/packages/neuron-ui/src/services/remote/app.ts
@@ -24,6 +24,22 @@ export const stopProcessMonitor = remoteApi<'ckb'>('stop-process-monitor')
 export const startProcessMonitor = remoteApi<'ckb'>('start-process-monitor')
 export const getIsCkbRunExternal = remoteApi<void, boolean>('is-ckb-run-external')
 
+export enum VerifyCkbVersionResult {
+  Same,
+  Compatible,
+  ShouldUpdate,
+  Incompatible,
+}
+
+export type VerifyExternalCkbNodeRes =
+  | {
+      ckb: VerifyCkbVersionResult
+      withIndexer: boolean
+    }
+  | undefined
+
+export const verifyExternalCkbNode = remoteApi<void, VerifyExternalCkbNodeRes>('verify-external-ckb-node')
+
 export const clearCellCache = remoteApi<Controller.ClearCache.Params>('clear-cache')
 
 export const invokeShowErrorMessage = remoteApi<{ title: string; content: string }>('show-error-message')
--- a/packages/neuron-ui/src/services/remote/app.ts
+++ b/packages/neuron-ui/src/services/remote/app.ts
@@ -24,20 +24,13 @@ export const stopProcessMonitor = remoteApi<'ckb'>('stop-process-monitor')
 export const startProcessMonitor = remoteApi<'ckb'>('start-process-monitor')
 export const getIsCkbRunExternal = remoteApi<void, boolean>('is-ckb-run-external')
 
-export enum VerifyCkbVersionResult {
-  Same,
-  Compatible,
-  ShouldUpdate,
-  Incompatible,
-}
-
 export type VerifyExternalCkbNodeRes =
   | {
-      ckb: VerifyCkbVersionResult
+      ckbIsCompatible: boolean | undefined
       withIndexer: boolean
+      shouldUpdate: boolean
     }
   | undefined
-
 export const verifyExternalCkbNode = remoteApi<void, VerifyExternalCkbNodeRes>('verify-external-ckb-node')
 
 export const clearCellCache = remoteApi<Controller.ClearCache.Params>('clear-cache')
--- a/packages/neuron-ui/src/widgets/AlertDialog/index.tsx
+++ b/packages/neuron-ui/src/widgets/AlertDialog/index.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react'
+import React, { useMemo, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useDialog } from 'utils'
 import Button from 'widgets/Button'
@@ -8,6 +8,7 @@ import Tips from 'widgets/Icons/Tips.png'
 import styles from './alertDialog.module.scss'
 
 type AlertType = 'success' | 'failed' | 'warning'
+type Action = 'ok' | 'cancel' | 'all'
 
 const AlertDialog = ({
   show,
@@ -17,6 +18,7 @@ const AlertDialog = ({
   onClose,
   onOk,
   onCancel,
+  action,
 }: {
   show?: boolean
   title?: string
@@ -25,10 +27,17 @@ const AlertDialog = ({
   onClose?: () => void
   onOk?: () => void
   onCancel?: () => void
+  action?: Action
 }) => {
   const [t] = useTranslation()
   const dialogRef = useRef<HTMLDialogElement | null>(null)
   useDialog({ show, dialogRef, onClose: onClose || (() => {}) })
+  const actions = useMemo<('cancel' | 'ok')[]>(() => {
+    if (action) {
+      return action === 'all' ? ['cancel', 'ok'] : [action]
+    }
+    return type === 'warning' ? ['cancel', 'ok'] : ['ok']
+  }, [action, type])
 
   return (
     <dialog ref={dialogRef} className={styles.alertDialog}>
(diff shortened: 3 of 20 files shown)


packages/neuron-ui/src/services/remote/app.ts
```diff
@@ -24,6 +24,22 @@ export const stopProcessMonitor = remoteApi<'ckb'>('stop-process-monitor')
 export const startProcessMonitor = remoteApi<'ckb'>('start-process-monitor')
 export const getIsCkbRunExternal = remoteApi<void, boolean>('is-ckb-run-external')
 
+export enum VerifyCkbVersionResult {
+  Same,
+  Compatibility,
```
Keith-CY, line 29: `Compatibility` and `NotCompatibility` could be `Compatible` and `Incompatible`

packages/neuron-ui/src/widgets/AlertDialog/index.tsx
```diff
@@ -25,10 +27,17 @@ const AlertDialog = ({
...
   onCancel?: () => void
+  action?: Action
 }) => {
   const [t] = useTranslation()
   const dialogRef = useRef<HTMLDialogElement | null>(null)
   useDialog({ show, dialogRef, onClose: onClose || (() => {}) })
+  const actions = useMemo<('cancel' | 'ok')[]>(() => {
+    if (action) {
+      return action === 'all' ? ['cancel', 'ok'] : [action]
+    }
```
Keith-CY, line 38: Why is `all` mapped to `[cancel, ok]`

packages/neuron-wallet/src/services/node.ts
```diff
@@ -21,6 +21,15 @@ import { generateRPC } from '../utils/ckb-rpc'
 import startMonitor from './monitor'
 import { CKBLightRunner } from './light-runner'
 
+export enum VerifyCkbVersionResult {
```
Keith-CY, line 24: `Compatibility` and `NotCompatibility` could be `Compatible` and `Incompatible`

packages/neuron-wallet/src/services/node.ts
```diff
@@ -219,38 +234,35 @@ class NodeService {
...
-        type: 'warning',
-        message: t('messageBox.node-version-different.message', { version: internalNodeVersion }),
-      })
+    if (internalMajor === externalMajor && internalMinor === externalMinor) return VerifyCkbVersionResult.Same
+    if (internalMajor < externalMajor || (internalMajor === externalMajor && internalMinor < externalMinor)) {
+      return VerifyCkbVersionResult.ShouldUpdate
+    }
+    if (supportCkbVersions.every(v => !localNodeInfo.version.startsWith(v))) {
+      return VerifyCkbVersionResult.NotCompatibility
     }
```
Keith-CY, line 250: The comparison should be an independent function for testing.

packages/neuron-wallet/src/services/node.ts
```diff
@@ -21,6 +21,15 @@ import { generateRPC } from '../utils/ckb-rpc'
...
 import { CKBLightRunner } from './light-runner'
 
+export enum VerifyCkbVersionResult {
+  Same,
+  Compatibility,
+  ShouldUpdate,
+  NotCompatibility,
+}
+
+const supportCkbVersions = ['0.110', '0.109']
```
Keith-CY, line 31: How about defining it in an independent file so the compatibility table can be shared elsewhere.

Keith-CY: Especially to be synced to the troubleshooting page

packages/neuron-ui/src/widgets/AlertDialog/index.tsx
yanguoyu: `all` means show `cancel` and `ok` action buttons.

packages/neuron-wallet/src/services/node.ts
yanguoyu: Do you mean to add test cases for the function? https://github.com/yanguoyu/neuron/blob/feat-ckb-check/packages/neuron-wallet/tests/services/node.test.ts#L455-L504 is the test case.

packages/neuron-wallet/src/services/node.ts
yanguoyu: The troubleshooting page will show all Neuron versions, but here it will only show the current Neuron version.
And how to share it to the troubleshooting page? They are different projects.

Keith-CY: It won't be considerable to include a complete compatibility table but it helps users to check compatibility without visiting another project.

The troubleshooting page fetches contents, e.g. issues and discussions from this project, so it can fetch the compatibility table similarly, to render the compatibility table on page.

---

CSV is preferable to be used for the compatibility table, e.g. https://raw.githubusercontent.com/citizenlab/test-lists/master/lists/00-LEGEND-category_codes.csv

packages/neuron-wallet/src/services/node.ts
yanguoyu: IMO, `verifyCkbNodeVersion` is small, should I split this function into two functions `getCKbVersion` and `veifyCkbVersion`?

packages/neuron-wallet/src/services/node.ts
yanguoyu: |  Neuron  | CKB |
|  ----  | ----  |
| 0.110  | 0.109;0.110; |
| 0.106  | 0.105;0.106;0.107;0.108; |
| 0.103 | 0.103;0.104;|

How about defining it like this?

Keith-CY: > Neuron	CKB
> 0.110	0.109;0.110;
> 0.106	0.105;0.106;0.107;0.108;
> 0.103	0.103;0.104;
> How about defining it like this?

It's hard to read and parse, I would suggest https://gist.github.com/Keith-CY/f76ebd3ae0f89e5a5f6dc85e29096144

packages/neuron-wallet/src/services/node.ts
Keith-CY: I would define a utility function `getCompatibility(neuronVersion, nodeVersion)` to return the compatibility and tested by a set of test cases.

Then call it in `verifyExternalCkbNode()` with `getExternalCkbNodeVersion()` and `getNeuronVersion()`

```js
const nodeVersion = getExternalCkbNodeVersion()
const neuronVersion = getNeuronVersion()
return getCompatibility(neuronVersion, nodeVersion)
```

By doing this, only `getExternalCkbNodeVersion` should be tested with mock. Or not, because the external ckb node version can be fetched by RPC

devchenyan (approved):

silySuper (approved):

compatible.csv
```diff
@@ -0,0 +1,5 @@
+CKB,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
+Neuron
```
Keith-CY, line 2:
```suggestion
Neuron,,,,,,,,
```

packages/neuron-wallet/src/services/node.ts
```diff
@@ -219,38 +235,58 @@ class NodeService {
...
+
+    if (internalMajor === externalMajor && internalMinor === externalMinor) return VerifyCkbVersionResult.Same
+    if (internalMajor < externalMajor || (internalMajor === externalMajor && internalMinor < externalMinor)) {
+      return VerifyCkbVersionResult.ShouldUpdate
     }
+    const supportCkbVersions = this.getNeuronCompatibilityCKB(neuronCKBVersion)
+    if (supportCkbVersions?.every(v => !externalCKBVersion.startsWith(v))) {
+      return VerifyCkbVersionResult.Incompatible
+    }
+    return VerifyCkbVersionResult.Compatible
```
Keith-CY, line 275: What if loading the compatibility result from the compatibility table

yanguoyu: `getNeuronCompatibilityCKB` loading the compatibility result from the compatibility table

Keith-CY: My bad, but why not `get compatibility result` first, and check `should update` when it returns `incompatible`. The logic would be quite simple.

Keith-CY: The logic of locating the table is a bit weird. Usually we locate the value in a table by `row` and `column`, e.g. `result = table[row][column]`. In this case, the logic is 

```
1. columns = table[row]
2. return yes if column in columns
```

A bit hard to read.

yanguoyu: > My bad, but why not `get compatibility result` first, and check `should update` when it returns `incompatible`. The logic would be quite simple.

I found that `VerifyCkbVersionResult.ShouldUpdate` should return by another field, because `ShouldUpdate` may intersect with `VerifyCkbVersionResult.Incompatible` and `VerifyCkbVersionResult.Compatible`.
So I will change the return type of `verifyExternalCkbNode`. And the `getCompatibility` function will only return true or false.

```typescript
// verifyExternalCkbNode return type
Promise<{
ckbIsCompatible: boolean
withIndexer: boolean
shouldUpdate: boolean
}>
```

compatible.csv
Keith-CY: It's blamed on GitHub if `,`s don't exist.

image

I think CSV has its own spec

packages/neuron-wallet/src/services/node.ts
Keith-CY: > > The logic of locating the table is a bit weird. Usually we locate the value in a table by `row` and `column`, e.g. `result = table[row][column]`. In this case, the logic is
> > ```
> > 1. columns = table[row]
> > 2. return yes if column in columns
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > A bit hard to read.
> 
> Sorry, I can not get the point. Finally, we need to verify whether the value is `yes` to `no`

An intuitive way to get compatibility would be `table[neuronVersion][ckbVersion] === yes`. The current implementation is

```
const supportedCkbVersion = table[neuronVersion].filter(compatible => compatible === yes)
const compatible = supportedCkbVersion.includes(ckbVersion)
```

Keith-CY (approved):

PR #2813, opened 2023-08-25
feat: Change default network name to Internal Node

yanguoyu:
1. Download and install Neuron for the first time. We can mock this to delete the `network.json` file at `/Users/{username}/Library/Application Support/Neuron/networks/index.json`

before:

image

after migrate:

image

2. Has added a new network but not edit the default network

before:

image

after migrate:

image

3. Has added a new network and edited the default network

before:

image

after migrate:

image


packages/neuron-ui/src/widgets/AlertDialog/index.tsx
```diff
@@ -51,7 +51,7 @@ const AlertDialog = ({
           v === 'cancel' ? (
             <Button type="cancel" onClick={onCancel} label={t('common.cancel')} />
           ) : (
-            <Button type="confirm" onClick={onCancel || onOk} label={t('common.confirm')} />
+            <Button type="confirm" onClick={onOk || onCancel} label={t('common.confirm')} />
```
Keith-CY, line 54: I haven't reviewed the previous PRs so I wonder why `onCancel` is bound to `confirm`

Keith-CY (approved):

packages/neuron-ui/src/widgets/AlertDialog/index.tsx
yanguoyu: When the type is `failed`, the `onCancel` function is used to close the dialog.

Keith-CY: Got it

WhiteMinds (approved):

silySuper:
/package.

silySuper:
I need to verify that does the  internal node whether support edit  or not now?

yanguoyu:
> I need to verify that does the internal node whether support edit or not now?

of cause

silySuper (approved):

PR #2817, opened 2023-08-29
refactor: remove ckb-sdk-js from dependency list

zhangyouxin:
## Description

Resolve: https://github.com/Magickbase/neuron-public-issues/issues/201

This is the last PR of https://github.com/Magickbase/neuron-public-issues/issues/201, finally removed ckb-sdk-js from dependency list, it has changed:

- fixvec
- Blake2b implementation
- rawTransactionToHash
- ECPairs


yanguoyu (approved):

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -417,8 +417,9 @@ export default class TransactionSender {
     const message = blake2b.digest()
 
     if (!wallet.isHardware()) {
-      const keyPair = new ECPair(privateKeyOrPath)
-      emptyWitness.lock = keyPair.signRecoverable(message)
+      // `privateKeyOrPath` variable here is path only when wallet is hardware, otherwise it's a private key
```
Keith-CY, line 420:
```suggestion
      // `privateKeyOrPath` variable here is a private key because wallet is not a hardware one. Otherwise, it will be a private key path.
```

Reorder the expression because it's in the branch of `(!isHardware())`

Keith-CY (approved):

silySuper:
/package.

packages/neuron-wallet/src/models/transaction-size.ts
```diff
@@ -55,7 +55,8 @@ export default class TransactionSize {
   }
 
   public static outputData(data: string): number {
-    const bytes = serializeFixVec(data)
+    const fixvecCodec = fixvec(number.Uint8)
+    const bytes = byteUtils.hexify(fixvecCodec.pack(Array.from(byteUtils.bytify(data))))
```
homura, line 59:
```suggestion
    const bytes = byteUtils.hexify(fixvecCodec.pack(data))
```

Why do we need an `Array.from` here?

packages/neuron-wallet/tests/services/address.test.ts
```diff
@@ -307,10 +307,10 @@ describe('integration tests for AddressService', () => {
         })
       })
     })
-
     describe('#generateAndSaveForPublicKey', () => {
       describe('with public key info exist for the public key', () => {
-        const publicKey = 'public key'
+        // public key is a valid 32 byte hex string
+        const publicKey = '0x' + '0'.repeat(66)
```
homura, line 313:
```suggestion
        const publicKey = '0x' + '00'.repeat(32)
```

packages/neuron-wallet/src/models/transaction-size.ts
zhangyouxin: `fixvecCodec` here is an `ArrayCodec` which only accepts `ArrayLike` parameters

image

packages/neuron-wallet/tests/services/address.test.ts
zhangyouxin: updated

packages/neuron-wallet/src/models/transaction-size.ts
homura: https://github.com/nervosnetwork/ckb/blob/5a7efe7a0b720de79ff3761dc6e8424b8d5b22ea/util/types/schemas/blockchain.mol#L63C30-L63C30

Can we use the `BytesVec.pack` to calculate the size? I find the `outputs_data` is a `BytesVec`

silySuper:
/package.

packages/neuron-wallet/src/models/transaction-size.ts
zhangyouxin: It's cool that you pointed out this, but the original code was not **exactly** serializing `outputs_data` to a `BytesVec`, instead it calculates part of `BytesVec` size by adding some `offset`: https://github.com/nervosnetwork/neuron/blob/101b7e12b5833b3f561d7eccdc253ec2fd45466f/packages/neuron-wallet/src/models/transaction-size.ts#L57-L60

A real `BytesVec` contains `header` (4 bytes), `offset` (4 bytes), `body` (4 bytes body size, followed by original data), but original code only contains `4 bytes body size` and `4 bytes offset`, which is less than result of `BytesVec.pack(data).bytelength` by 4 bytes. so the result could be like:
```
public static outputData(data: string): number {
  return BytesVec.pack([byteUtils.bytify(data)]).byteLength - 4
}
```
The code looks like magic.


We should be careful here, so maybe we can just exactly do every step as before, and use lumos. but not changing implementation here.

homura (approved):

silySuper:
https://github.com/Magickbase/neuron-public-issues/issues/276

zhangyouxin:
> Please resolve conflicts and merge @zhangyouxin

I'm not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request. anybody authorized please give a hand

PR #2840, opened 2023-09-14
revert: Remove cache block number and filter cell

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/282. 
If Neuron is released soon, This PR should be merged and I will continue to complete the cell storage optimate into the next version.
If not, This PR should not be merged and I will as soon as possible complete the cell storage optimate and open a PR.

### DIFF
--- a/packages/neuron-wallet/src/services/addresses.ts
+++ b/packages/neuron-wallet/src/services/addresses.ts
@@ -201,7 +201,7 @@ export default class AddressService {
     await getConnection().manager.save(publicKeyInfo)
 
     const addressMeta = AddressMeta.fromHdPublicKeyInfoModel(publicKeyInfo.toModel())
-    this.notifyAddressCreated([addressMeta], undefined)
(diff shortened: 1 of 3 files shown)


packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -201,7 +201,7 @@ export default class AddressService {
     await getConnection().manager.save(publicKeyInfo)
 
     const addressMeta = AddressMeta.fromHdPublicKeyInfoModel(publicKeyInfo.toModel())
-    this.notifyAddressCreated([addressMeta], undefined)
+    AddressService.notifyAddressCreated([addressMeta], undefined)
```
yanguoyu, line 204: `generateAndSaveForPublicKey` is a static function, and a static function calls it. If use `this`, when I start Neuron the first time without a database file, it throws an error that can not called with undefined.

PR #2841, opened 2023-09-16
fix: Fix miss saving some inputs and outputs with optimize input and output storage

yanguoyu:
Fixed https://github.com/Magickbase/neuron-public-issues/issues/282.

1. Save entity into `indexer_tx_hash_cache` order by the block number
2. Create a queue to check generate address
3. When starting the sync process, check whether there exist some lock args belonging to the Neuron wallet


Keith-CY (approved):

Keith-CY:
Please have a review @homura @zhangyouxin  @devchenyan @WhiteMinds 
And please have a test simultaneously @silySuper @FrederLu

devchenyan (approved):

FrederLu (approved):

PR #2849, opened 2023-09-25
Fix: fix rc/110.3 bugs, detail in comments

yanguoyu:
1. Fix styles when the desc or symbol is overlong
2. Close dialog when submitting success with migrating to create a new account
3. Reset the deposit value when the dialog's visible changes and show an error dialog when the deposit fails.
4. Ignore update the transaction that calls `get_transaction` failed.

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
@@ -3,10 +3,17 @@ import { CONNECTION_NOT_FOUND_NAME } from '../database/chain/ormconfig'
 import { FailedTransaction, TransactionPersistor } from '../services/tx'
 import RpcService from '../services/rpc-service'
 import NetworksService from '../services/networks'
-import { TransactionStatus } from '../models/chain/transaction'
+import Transaction, { TransactionStatus } from '../models/chain/transaction'
 import TransactionWithStatus from '../models/chain/transaction-with-status'
 import logger from '../utils/logger'
 import { interval } from 'rxjs'
+
+type TransactionDetail = {
+  hash: string
+  tx: Transaction | undefined
+  status: TransactionStatus
+  blockHash: string | null
+}
 
 const getTransactionStatus = async (hash: string) => {
   const url: string = NetworksService.getInstance().getCurrent().remote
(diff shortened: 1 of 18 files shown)


devchenyan (approved):

Keith-CY (approved):

Keith-CY:
`Clear tx` should be performed on initialization, e.g. when the deposit dialog opens, instead of the workflow end. Otherwise, the initial tx may be affected by other workflows. For example,

1. open a deposit dialog and enter a value
2. close the dialog and the tx is cleared
3. do some work and a tx is generated, but the post-treat is missing by accident
4. open a deposit dialog, the tx generated in step 3 will be the initial one

yanguoyu:
> `Clear tx` should be performed on initialization, e.g. when the deposit dialog opens, instead of the workflow end. Otherwise, the initial tx may be affected by other workflows. For example,
> 
> 1. open a deposit dialog and enter a value
> 2. close the dialog and the tx is cleared
> 3. do some work and a tx is generated, but the post-treat is missing by accident
> 4. open a deposit dialog, the tx generated in step 3 will be the initial one
> 
> To be precise, it's a reset of initial tx

Actually, when the deposit dialog opens, the transaction will be refreshed. It shows the last deposit value because the deposit dialog was not destroyed. So the last deposit value exists.  So I reset the deposit value when the dialog visible changes at https://github.com/nervosnetwork/neuron/pull/2849/commits/045d09b43d9ee40f08037e9b39df1b9dc6d40443

silySuper:
Asset Accounts-Receive page shows abnormal when token name is long.

截屏2023-09-26 15 26 14

silySuper:
History page shows abnormal when token name is long.

https://github.com/nervosnetwork/neuron/assets/25971273/3f5831ff-e8f1-4f85-ae4b-f0c9ff485670

yanguoyu:
@silySuper Fixed by https://github.com/nervosnetwork/neuron/pull/2849/commits/a97012f65fcf6430f3da6bff4e58b14ee7fd125f

Keith-CY:
Everything works well on my computer now.

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -42,18 +49,28 @@ const trackingStatus = async () => {
...
+        // ignore error, get failed skip current update
       }
     })
   )
 
-  const failedTxs = txs.filter(tx => tx.status === TransactionStatus.Failed)
-  const successTxs = txs.filter(tx => tx.status === TransactionStatus.Success)
+  const failedTxs = txs.filter(
+    (tx): tx is TransactionDetail & { status: TransactionStatus.Failed } =>
+      !!tx && tx.status === TransactionStatus.Failed
```
Keith-CY, line 68:
```suggestion
      tx?.status === TransactionStatus.Failed
```

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -42,18 +49,28 @@ const trackingStatus = async () => {
...
+  )
+  const successTxs = txs.filter(
+    (tx): tx is TransactionDetail & { status: TransactionStatus.Success } =>
+      !!tx && tx.status === TransactionStatus.Success
```
Keith-CY, line 72:
```suggestion
      tx?.status === TransactionStatus.Success
```

Keith-CY (approved):

silySuper (approved):

PR #2851, opened 2023-09-25
fix: patch label of textfield when it's not a string

Keith-CY:
This PR patches the label of a textfield by a fallback that uses a field name. It should be fixed in the next release that `label` should be limited to a string because it has its own semantic meaning.

Before
image
After
image

### DIFF
--- a/packages/neuron-ui/src/widgets/TextField/index.tsx
+++ b/packages/neuron-ui/src/widgets/TextField/index.tsx
@@ -30,7 +30,7 @@ const TextField = React.forwardRef(
       ...rest
     }: {
       field: string
-      label?: string
+      label?: string | React.ReactNode
       value?: string
       hint?: string
       error?: string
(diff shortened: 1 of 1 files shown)


yanguoyu (approved):

packages/neuron-ui/src/widgets/TextField/index.tsx
```diff
@@ -30,7 +30,7 @@ const TextField = React.forwardRef(
       ...rest
     }: {
       field: string
-      label?: string
+      label?: string | React.ReactNode
```
WhiteMinds, line 33: Perhaps it could be implemented as `label?: string | { node: ReactNode; text: string }`

WhiteMinds (approved):

packages/neuron-ui/src/widgets/TextField/index.tsx
Keith-CY: > Perhaps it could be implemented as `label?: string | { node: ReactNode; text: string }`

I think `label` should be strictly limited to `string`, so the definition will be recovered to `string`, and if a pre-component is necessary, an independent field as `prefix/suffix/preTab/postTab` can be added

devchenyan (approved):

silySuper (approved):

PR #2860, opened 2023-10-07
fix: fixed locktime cell,multisig sync blocknumber, table head style

yanguoyu:
1. Lock time cells should be saved into the local database.
2. For the light client skip update the multisig sync block number if the address is not the current wallet. Because of light client only subscribe current wallet multisig address
3. Fixed able head style.

image

4. Fix styles when the asset name is too long
image

5. Fix some elements that are hidden when Neuron height is the smallest

image


yanguoyu:
@devchenyan @Keith-CY @homura @silySuper 
Please have a view and test, it should merge into rc/110.3

Keith-CY (approved):

devchenyan (approved):

silySuper (approved):

Issue #2868, opened 2023-10-12
[Synchronization] **Always Connecting on Neuron v0.110.3**


yanguoyu:
#### What happened?
For some reason, Neuron will not open the migrate dialog to help users migrate the old ckb data. Then the Neuron will always show `Connecting`.

#### How to determine this will help you?
We can download the debug information by clicking `Help -> Export Debug Information` and unzip it. Then open the `main.log`, if we find some word like this `For optimal performance, CKB wants to migrate the data into new format.` The following actions may be helpful to you, else they may caused by other problems.

#### How to resolve it?
We can find these words `Run "${command}" and confirm by typing "YES" to migrate the data.` after `We strongly recommended you to use the latest stable version of CKB, since the old versions may have unfixed vulnerabilities.` in the main.log.
Copy the **command** and paste it on to the terminal, then press enter and confirm by typing "YES" to migrate the data. After the migration finish, then restart Neuron, it will work well.

Crypto1967:
I have this same problem and I have no idea how to correct this. I should have never updated the wallet. There isn't even a sync status showing up, just says connecting. It's been saying that for over 36 hours. How to I correct this problem or go back to software that works? Error log says downgrade is disallowed. 

[2023-10-14 19:56:11.128] [error] CKB:	run fail: For optimal performance, CKB wants to migrate the data into new format.

Keith-CY:
> I have this same problem and I have no idea how to correct this. I should have never updated the wallet. There isn't even a sync status showing up, just says connecting. It's been saying that for over 36 hours. How to I correct this problem or go back to software that works? Error log says downgrade is disallowed.
> 
> [2023-10-14 19:56:11.128] [error] CKB: run fail: For optimal performance, CKB wants to migrate the data into new format.

Hi @Crypto1967 

We have added a dialog for migration so it can be performed with a click
https://github.com/nervosnetwork/neuron/pull/2869
It's being tested and will be released this week

Crypto1967:
Thanks, please let me know when it's available and instructions to on how to do it. Many thanks. John
    On Sunday, October 15, 2023 at 12:01:16 PM EDT, Chen Yu ***@***.***> wrote:  
 
 



I have this same problem and I have no idea how to correct this. I should have never updated the wallet. There isn't even a sync status showing up, just says connecting. It's been saying that for over 36 hours. How to I correct this problem or go back to software that works? Error log says downgrade is disallowed.

[...]

PR #2869, opened 2023-10-12
fix: Show the migrate ckb data dialog

yanguoyu:
If we need to run `ckb migrate` for the ckb data, the neuron will show a dialog to help users migrate ckb data. Once users have selected migrate, the neuron will auto-migrate next time if the before-migrate version is the same. The migrating version will be saved in the electron local storage, if we want to show the migrate dialog again, we can delete the local storage, which can be found at `/Users/${username}/Library/Application Support/Neuron/Local Storage`.

https://github.com/nervosnetwork/neuron/assets/12881040/3da60666-e6ba-47c3-a392-cbc0276ed4f8

### DIFF
--- a/packages/neuron-ui/src/components/PageContainer/index.tsx
+++ b/packages/neuron-ui/src/components/PageContainer/index.tsx
@@ -68,7 +68,7 @@ const PageContainer: React.FC<ComponentProps> = props => {
     return [
       bestBlockNumber > 0 && cacheTipBlockNumber > 0
         ? `${+((cacheTipBlockNumber * 100) / bestBlockNumber).toFixed(2)}%`
-        : '0',
+        : '0.00%',
       `${cacheTipBlockNumber >= 0 ? localNumberFormatter(cacheTipBlockNumber) : '-'} / ${
         bestBlockNumber >= 0 ? localNumberFormatter(bestBlockNumber) : '-'
       }`,
--- a/packages/neuron-ui/src/services/subjects.ts
+++ b/packages/neuron-ui/src/services/subjects.ts
@@ -37,13 +37,14 @@ const SubjectConstructor = <T>(
   return ipcRenderer
     ? {
         subscribe: (handler: (data: T) => void) => {
-          ipcRenderer.on(channel, (_e: Event, data: T) => {
+          const handlerWrap = (_e: Event, data: T) => {
             handler(data)
-          })
+          }
+          ipcRenderer.on(channel, handlerWrap)
           return {
             unsubscribe: () => {
               if (isMulti) {
-                ipcRenderer.removeListener(channel, handler)
+                ipcRenderer.removeListener(channel, handlerWrap)
               } else {
                 ipcRenderer.removeAllListeners(channel)
               }
--- a/packages/neuron-ui/src/services/subjects.ts
+++ b/packages/neuron-ui/src/services/subjects.ts
@@ -67,7 +67,7 @@ export const Navigation = SubjectConstructor<Subject.URL>('navigation')
 export const SetLocale = SubjectConstructor<(typeof LOCALES)[number]>('set-locale')
 export const DeviceSignIndex = SubjectConstructor<Subject.SignIndex>('device-sign-index')
 export const MultisigOutputUpdate = SubjectConstructor<string>('multisig-output-update')
-export const Migrate = SubjectConstructor<'need-migrate' | 'migrating' | 'failed' | 'finish'>('migrate')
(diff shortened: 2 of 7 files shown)


Keith-CY:
I remembered that `migrating` will be displayed where `progress` sat in the deprecated UI

yanguoyu:
> I remembered that `migrating` will be displayed where `progress` sat in the deprecated UI

Of course, may be because it was too fast that the migrating status was not shown. But I find the migrating status will not change to syncing status, I will fix it and upload the full process video.

yanguoyu:
The full migration video:

https://github.com/nervosnetwork/neuron/assets/12881040/311aac59-a6b4-4bf8-9505-606da734a76d

packages/neuron-ui/src/components/PageContainer/index.tsx
```diff
@@ -68,7 +68,7 @@ const PageContainer: React.FC<ComponentProps> = props => {
     return [
       bestBlockNumber > 0 && cacheTipBlockNumber > 0
         ? `${+((cacheTipBlockNumber * 100) / bestBlockNumber).toFixed(2)}%`
-        : '0',
+        : '0.00%',
```
Keith-CY, line 71: I think `0%` is clear and neat, but just my personal opinion

packages/neuron-ui/src/services/subjects.ts
```diff
@@ -67,7 +67,7 @@ export const Navigation = SubjectConstructor<Subject.URL>('navigation')
 export const SetLocale = SubjectConstructor<(typeof LOCALES)[number]>('set-locale')
 export const DeviceSignIndex = SubjectConstructor<Subject.SignIndex>('device-sign-index')
 export const MultisigOutputUpdate = SubjectConstructor<string>('multisig-output-update')
-export const Migrate = SubjectConstructor<'need-migrate' | 'migrating' | 'failed' | 'finish'>('migrate')
+export const Migrate = SubjectConstructor<'need-migrate' | 'migrating' | 'failed' | 'finish'>('migrate', true)
```
Keith-CY, line 70: What is `true` here for

packages/neuron-ui/src/components/PageContainer/index.tsx
yanguoyu: I want to keep unify with `toFixed(2)`.

packages/neuron-ui/src/services/subjects.ts
yanguoyu: When calling unsubscribe the channel, it will only unsubscribe created by itself, but not unsubscribe all of the channel that name is the same.

Keith-CY: Got it

Keith-CY (approved):

devchenyan (approved):

silySuper:
https://github.com/Magickbase/neuron-public-issues/issues/301

packages/neuron-ui/src/utils/hooks/index.ts
```diff
@@ -405,10 +405,22 @@ export const useToggleChoiceGroupBorder = (containerSelector: string, borderClas
...
+  const [isMigrateDialogShow, setIsMigrateDialogShow] = useState(false)
+  const [hasDismissMigrate, setHasDismissMigrate] = useState(false)
+  const [ckbDataPath, setCkbDataPath] = useState<string>()
+  useEffect(() => {
+    getCkbNodeDataPath().then(res => {
+      if (isSuccessResponse(res) && res.result) {
+        setCkbDataPath(res.result)
+      }
+    })
+  })
```
WhiteMinds, line 418: This doesn't seem to be something that should be executed on every render, did you forget to write deps?

WhiteMinds (changes requested):

packages/neuron-ui/src/utils/hooks/index.ts
yanguoyu: You are right, I miss the deps. It will add the dep when location. path changed.

WhiteMinds (approved):

silySuper (approved):

MarkoSARIC1990:
How can we do migration on Neuron, i cant understand from the video, thank you

yanguoyu:
> How can we do migration on Neuron, i cant understand from the video, thank you

Neuron should open a migration dialog for the user to choose to migrate if migration is needed. If the dialog does not show,
or we can manually migrate refer to https://github.com/nervosnetwork/neuron/issues/2868#issue-1938970527

PR #2890, opened 2023-10-19
fix: Update compatible.csv after update ckb and Neuron version

yanguoyu:
This step should be added to the release steps. We'd better add an action to update it to avoid forgetting.

### DIFF
--- a/compatible.csv
+++ b/compatible.csv
@@ -1,5 +1,6 @@
-CKB,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
-Neuron,,,,,,,,
-0.110,yes,yes,no,no,no,no,no,no
-0.106,no,no,yes,yes,yes,yes,no,no
-0.103,no,no,no,no,no,no,yes,yes
+CKB,0.111,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
+Neuron,,,,,,,,,
+0.111,yes,yes,yes,no,no,no,no,no,no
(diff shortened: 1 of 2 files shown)


Keith-CY (approved):

devchenyan (approved):

compatible.csv
```diff
@@ -1,5 +1,6 @@
-CKB,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
-Neuron,,,,,,,,
-0.110,yes,yes,no,no,no,no,no,no
-0.106,no,no,yes,yes,yes,yes,no,no
-0.103,no,no,no,no,no,no,yes,yes
+CKB,0.111,0.110,0.109,0.108,0.107,0.106,0.105,0.104,0.103
+Neuron,,,,,,,,,
+0.111,yes,yes,yes,no,no,no,no,no,no
```
homura, line 3: Not really related to this PR, but I'd like to ask the source of this table, it seems like it's not the same as the compatibility table from [CKB releases](https://github.com/nervosnetwork/ckb/releases/tag/v0.111.0)

yanguoyu: The 109 version changes an RPC function name(`get_fee_rate_statics` -> `get_fee_rate_statistics`), and the RPC function that the Neuron needs. So we set it to false with the 108 version.
https://github.com/nervosnetwork/ckb/releases/tag/v0.109.0

homura (approved):

PR #2915, opened 2023-10-27
feat: Split full node and light client db file.

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/294.
After this PR, clearing the full node cache will not affect the light client, and vice versa.
If the users change settings for the full node testnet, it will also change for the light client testnet, and vice versa.


silySuper:
In light testnet client,click settings,show error like this:
截屏2023-12-08 10 00 24

packages/neuron-wallet/src/database/chain/ormconfig.ts
```diff
@@ -133,28 +144,54 @@ const connectOptions = async (genesisBlockHash: string): Promise<SqliteConnectio
...
     logger.error(err.message)
   }
 }
 
+export async function initConnection(genesisBlockHash: string) {
+  await initConnectionWithType(genesisBlockHash, 'full')
+  await initConnectionWithType(genesisBlockHash, 'light')
+}
+
+export function migrateDBFile(genesisBlockHash: string) {
```
Keith-CY, line 183: Will the original DB file be removed?

homura, line 183: Recommend renaming this method to a more certain name to ensure that it will work if a later version also needs to be migrated, such as

```ts
declare function migrate_0_112(): Promise<void>;

function migrate() {
  migrate_0_112()
}
```

And in the future version, we can upgrade it by

```ts
function migrate() {
  if (lastVersion <= v0_112) await migrate_0_112()
  if (lastVersion <= v0_11x) await migrate_0_11x()
}
```

packages/neuron-wallet/src/database/chain/connection.ts
```diff
@@ -0,0 +1,18 @@
...
+import NetworksService from '../../services/networks'
+
+export type ConnectionName = 'light' | 'full'
+
+export function getCurrentConnectionName(): ConnectionName {
+  return NetworksService.getInstance().getCurrent()?.type === NetworkType.Light ? 'light' : 'full'
+}
+
+export function getConnection(connectionName?: ConnectionName) {
+  const name: ConnectionName = connectionName ?? getCurrentConnectionName()
```
homura, line 12:
```suggestion
export function getConnection(name: ConnectionName = getCurrentConnectionName()) {
```

packages/neuron-wallet/src/database/chain/subscriber/address-subscriber.ts
```diff
@@ -0,0 +1,10 @@
+import { EventSubscriber } from 'typeorm'
+import UserSettingSubscriber from './user-setting-subscriber'
+import AddressDescription from '../entities/address-description'
+
+@EventSubscriber()
+export default class AddressSubscribe extends UserSettingSubscriber<AddressDescription> {
+  listenTo() {
+    return AddressDescription
+  }
+}
```
homura, line 10: what does the subscriber is used for? It seems to do nothing but listens to changes from `AddressDescription`

yanguoyu: It is used to sync `address-description` from two db files. I will change the `AddressSubscribe` to `AddressDescriptionSubscribe`

homura: Got it, I found that the `UserSettingSubscriber` contains some reusable pieces.

packages/neuron-wallet/src/database/chain/subscriber/user-setting-subscriber.ts
```diff
@@ -0,0 +1,57 @@
+import { EntitySubscriberInterface, InsertEvent, RemoveEvent, UpdateEvent } from 'typeorm'
+import { ConnectionName, getConnection, getCurrentConnectionName } from '../connection'
+
+type Constructor<T> = new (...args: unknown[]) => T
+
+export default abstract class UserSettingSubscriber<Entity extends object>
```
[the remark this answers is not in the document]
yanguoyu: It will be hard to split client-related and non-client-related data, sometimes we needs to join the table from the DB file. And we will needs to change much codes to adapt the more DB file.

packages/neuron-wallet/src/database/chain/ormconfig.ts
yanguoyu: Currently, it will not be removed. Then we can easily revert the migration and start testing again.

yanguoyu: Currently, we do not save the last version. If we need to use the last version, we'd better save the last version in the next version. And then in that next version we can get the last version

Keith-CY: Does that mean the old db is intended for recovery during testing, and will be removed in production?
If so, I would suggest removing the old one at first but leaving a message about how to backup it. By doing so, the migration process can be fully tested, including removing old db.

packages/neuron-wallet/src/database/chain/ormconfig.ts
```diff
@@ -133,28 +144,55 @@ const connectOptions = async (genesisBlockHash: string): Promise<SqliteConnectio
...
+  const originDBFile = dbPath(genesisBlockHash, 'cell')
+  const currentFullDBFile = dbPath(genesisBlockHash, 'full')
+  const currentLightDBFile = dbPath(genesisBlockHash, 'light')
+  if (fs.existsSync(originDBFile) && !fs.existsSync(currentLightDBFile)) {
+    if (!fs.existsSync(currentFullDBFile)) {
+      fs.copyFileSync(originDBFile, currentFullDBFile)
+    }
+    if (!fs.existsSync(currentLightDBFile)) {
```
Keith-CY, line 191: This condition is covered above `if (fs.existsSync(originDBFile) && !fs.existsSync(currentLightDBFile))`

packages/neuron-wallet/src/database/chain/ormconfig.ts
Keith-CY: Does the comment above help @homura

Keith-CY: Repost

yanguoyu:
Please have a review @WhiteMinds @devchenyan @zhangyouxin
Do you have any suggestions about this PR? @homura

devchenyan (approved):

silySuper (approved):

PR #2918, opened 2023-11-02
fix: address input issue

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/311

https://github.com/nervosnetwork/neuron/assets/11756869/0735ccf1-8947-4e11-906f-25b50ab28bc5

### DIFF
--- a/packages/neuron-ui/src/widgets/TextField/index.tsx
+++ b/packages/neuron-ui/src/widgets/TextField/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useRef, useState } from 'react'
+import React, { useCallback, useRef, useState, useEffect } from 'react'
 import Alert from 'widgets/Alert'
 import { ReactComponent as Edit } from 'widgets/Icons/Edit.svg'
 import { EyesClose, EyesOpen } from 'widgets/Icons/icon'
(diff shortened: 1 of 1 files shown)


Keith-CY:
The initial state is weird because text placeholder is not middle aligned, what if set `row=2` when the input field is focused.

Keith-CY:
> The initial state is weird because text placeholder is not middle aligned, what if set `row=2` when the input field is focused.

cc @devchenyan

devchenyan:
> > The initial state is weird because text placeholder is not middle aligned, what if set `row=2` when the input field is focused.
> 
> cc @devchenyan

updated, please review

devchenyan:
please review @yanguoyu @Keith-CY

packages/neuron-ui/src/widgets/TextField/index.tsx
```diff
@@ -55,11 +55,19 @@ const TextField = React.forwardRef(
     ref: React.LegacyRef<HTMLDivElement>
   ) => {
     const [isPasswordHidden, setIsPasswordHidden] = useState(true)
+    const inputRef = useRef<any>()
```
Keith-CY, line 58: `any` should be avoided, could be `HTMLInputElement | null`

packages/neuron-ui/src/widgets/TextField/index.tsx
```diff
@@ -55,11 +55,19 @@ const TextField = React.forwardRef(
...
     const changePasswordHide = useCallback(() => {
       setIsPasswordHidden(v => !v)
     }, [setIsPasswordHidden])
     // FIXME: label should be limited to a string because it has its own semantic meaning
     const labelStr = typeof label === 'string' ? label : field
+
+    useEffect(() => {
+      inputRef.current.focus()
```
Keith-CY, line 66: It's better to return early when `inputRef.current` is `undefined`

yanguoyu (approved):

packages/neuron-ui/src/widgets/TextField/index.tsx
Keith-CY: `inputRef.current` is `undefined` until `<textare />` is rendered, so `undefined` should be handled in the code. Otherwise, `inputRef.current.focus()` will throw.

devchenyan: > `inputRef.current` is `undefined` until `<textare />` is rendered, so `undefined` should be handled in the code. Otherwise, `inputRef.current.focus()` will throw.

fixed, but it doesn't show here.

image

Keith-CY: Got it

Keith-CY (approved):

silySuper (approved):

PR #2921, opened 2023-11-04
feat: Adapt switch internal network type, and notify the user network type is not running external.

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/284.

https://github.com/nervosnetwork/neuron/assets/12881040/e58b8d97-c035-410c-b5e7-6cca5a2731ff


packages/neuron-ui/src/components/NetworkSetting/index.tsx
```diff
@@ -75,20 +83,43 @@ const NetworkSetting = ({ chain = chainState, settings: { networks = [] } }: Sta
     }
   }, [setShowEditorDialog, setNotice, netId])
 
+  const onSwitchNetworkType = useCallback<React.MouseEventHandler<HTMLButtonElement>>(() => {
+    switchCurrentNetworkType()
+  }, [])
```
Keith-CY, line 88: Switching `type` can be transformed into switching `network id` in UI, so logic changes in the backend can be avoided.

packages/neuron-ui/src/containers/Main/hooks.ts
```diff
@@ -311,6 +317,69 @@ export const useSubscription = ({
   }, [walletID, pageNo, pageSize, keywords, isAllowedToFetchList, navigate, dispatch, location.pathname])
 }
 
+export const useCheckNode = (networks: State.Network[], network?: State.Network) => {
+  const [showSwitchNetwork, setShowSwitchNetwork] = useState<boolean>(false)
```
Keith-CY, line 321: Name it with `is~` to imply it's type of boolean

packages/neuron-ui/src/containers/Navbar/index.tsx
```diff
@@ -95,13 +96,16 @@ const Navbar = () => {
...
+  const network = useMemo(() => networks.find(n => n.id === networkID), [networkID, networks])
   useEffect(() => {
-    verifyExternalCkbNode().then(res => {
-      if (isSuccessResponse(res) && res.result) {
-        setVerifyCkbResult(res.result)
-      }
-    })
-  }, [])
+    if (!network?.readonly) {
+      verifyExternalCkbNode().then(res => {
```
Keith-CY, line 102: `Verify external ckb node` should be done in the background if `connected to an unexpected node` should be alerted to users so they can realize it even Neuron is running in the background. We can do this later.

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -130,10 +144,13 @@ export default class NetworksService extends Store {
     }
 
     Object.assign(network, options)
-    Object.assign(
-      network,
-      await CommonUtils.timeout(2000, this.refreshChainInfo(network), network).catch(() => network)
-    )
+    if (!network.readonly) {
+      // readonly network chaininfo can not update auto
```
Keith-CY, line 148:
```suggestion
      // readonly network chaininfo is immutable
```

Keith-CY:
The logic of informing users when an external node is connected while the internal network is selected is too complex.

It can be done by

1. check if the connected node matches the selected network at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/node.ts#L196
2. send the result to `ConnectionStatusSubject` at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/node.ts#L81
3. the result has been subscribed in UI at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/containers/Main/hooks.ts#L200
4. show the alert if necessary

yanguoyu:
> And as mentioned at https://github.com/nervosnetwork/neuron/pull/2921#discussion_r1382425727
switch network type can be transformed into switch network id in the UI, e.g.

So, some networks will be hidden? When the user chooses the mainnet, the testnet network will be hidden. Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?

yanguoyu:
> The logic of informing users when an external node is connected while the internal network is selected is too complex.

I think I can send the `_isCkbNodeExternal` to UI by this event, there are two conditions when the `startedBundledNode` is false, start the internal node failed or we do not try to start the internal node.

packages/neuron-ui/src/components/DataSetting/hooks.ts
```diff
@@ -20,13 +20,13 @@ export const useDataPath = () => {
...
 
-  useDidMount(() => {
+  useEffect(() => {
     getCkbNodeDataPath().then(res => {
       if (isSuccessResponse(res)) {
         setPrevPath(res.result!)
       }
     })
-  })
+  }, [network?.chain, network?.id])
```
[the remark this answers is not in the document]
yanguoyu: When the network chain switch, the ckb data directory will be changed, if we set the different IDs for different network chains, it can be removed. https://github.com/nervosnetwork/neuron/pull/2921#issuecomment-1793590460

Keith-CY:
> > And as mentioned at [#2921 (comment)](https://github.com/nervosnetwork/neuron/pull/2921#discussion_r1382425727)
> > switch network type can be transformed into switch network id in the UI, e.g.
> 
> So, some networks will be hidden? When the user chooses the mainnet, the testnet network will be hidden. Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?

Similar to being hidden
image
Mainnet for internal node and testnet for internal node(two options) are shrinked into one network option with a type switch

In the case of the screenshot, `switch to mainnet` will be treated as `switch to network ckb_mainnet`

Keith-CY:
> > The logic of informing users when an external node is connected while the internal network is selected is too complex.
> 
> I think I can send the `_isCkbNodeExternal` to UI by this event, there are two conditions when the `startedBundledNode` is false, start the internal node failed or we do not try to start the internal node.

My bad, the field works is `isBundledNode` above it, at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/node.ts#L80

So the targeted status is `connected === true && isBundledNode === false`

yanguoyu:
> > > The logic of informing users when an external node is connected while the internal network is selected is too complex.
> > 
> > 
> > I think I can send the `_isCkbNodeExternal` to UI by this event, there are two conditions when the `startedBundledNode` is false, start the internal node failed or we do not try to start the internal node.
> 
> My bad, the field works is `isBundledNode` above it, at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/node.ts#L80
> 
> So the targeted status is `connected === true && isBundledNode === false`

That sounds good, the targeted status should be `connected === true && isBundledNode === true && startedBundledNode === false`
`connected === true` ensures that the node is starting. `isBundledNode === true` ensures that the selected node is an internal node. `startedBundledNode === false` ensures that we do not start an internal node success.

yanguoyu:
> > > And as mentioned at [#2921 (comment)](https://github.com/nervosnetwork/neuron/pull/2921#discussion_r1382425727)
> > > switch network type can be transformed into switch network id in the UI, e.g.
> > 
> > 
> > So, some networks will be hidden? When the user chooses the mainnet, the testnet network will be hidden. Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> 
> Similar to being hidden image Mainnet for internal node and testnet for internal node(two options) are shrinked into one network option with a type switch
> 
> In the case of the screenshot, `switch to mainnet` will be treated as `switch to network id ckb_mainnet`
> 
> > Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> 
> Sure, there's no need for one more API because `switch network id` and `switch network type` do the same work under the hood.

I have tried that, but It seems there are some differences. When we switch the network ID, we should verify and notify the users when an external node is running. But when we switch the internal network type, it's unnecessary and just starts with an internal node.

Keith-CY:
> > > > And as mentioned at [#2921 (comment)](https://github.com/nervosnetwork/neuron/pull/2921#discussion_r1382425727)
> > > > switch network type can be transformed into switch network id in the UI, e.g.
> > > 
> > > 
> > > So, some networks will be hidden? When the user chooses the mainnet, the testnet network will be hidden. Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> > 
> > 
> > Similar to being hidden image Mainnet for internal node and testnet for internal node(two options) are shrinked into one network option with a type switch
> > In the case of the screenshot, `switch to mainnet` will be treated as `switch to network id ckb_mainnet`
> > > Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> > 
> > 
> > Sure, there's no need for one more API because `switch network id` and `switch network type` do the same work under the hood.
> 
> I have tried that, but It seems there are some differences. When we switch the network ID, we should verify and notify the users when an external node is running. But when we switch the internal network type, it's unnecessary and just starts with an internal node.

I didn't get the point. There are two independent processes.

1. Switch network, it sends a message to the background to update the active network, and happens before the node is connected.
2. Verify the connected node, it checks if the `internal network option` is active while the `connected node` is outside Neuron, happens after the node is connected.

They are 2 different logic happening before and after the node is connected, how will they be coupled.

Maybe you're mentioning that `verification` can be skipped when `switch chain type`, but there's an assumption that `switch chain type` definitely connects to the internal node, I'm not sure about that.

yanguoyu:
> > > > > And as mentioned at [#2921 (comment)](https://github.com/nervosnetwork/neuron/pull/2921#discussion_r1382425727)
> > > > > switch network type can be transformed into switch network id in the UI, e.g.
> > > > 
> > > > 
> > > > So, some networks will be hidden? When the user chooses the mainnet, the testnet network will be hidden. Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> > > 
> > > 
> > > Similar to being hidden image Mainnet for internal node and testnet for internal node(two options) are shrinked into one network option with a type switch
> > > In the case of the screenshot, `switch to mainnet` will be treated as `switch to network id ckb_mainnet`
> > > > Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> > > 
> > > 
> > > Sure, there's no need for one more API because `switch network id` and `switch network type` do the same work under the hood.
> > 
> > 
> > I have tried that, but It seems there are some differences. When we switch the network ID, we should verify and notify the users when an external node is running. But when we switch the internal network type, it's unnecessary and just starts with an internal node.
> 
> I didn't get the point. There are two independent processes.
> 
> 1. Switch network, it sends a message to the background to update the active network, and happens before the node is connected.
> 2. Verify the connected node, it checks if the `internal network option` is active while the `connected node` is outside Neuron, happens after the node is connected.
> 
> They are 2 different logic happening before and after the node is connected, how will they be coupled.
> 
> Maybe you're mentioning that `verification` can be skipped when `switch chain type`, but there's an assumption that `switch chain type` definitely connects to the internal node, I'm not sure about that(because they are two independent processes IMO).

When the user starts the mainnet with another port, currently, if they want to switch to the mainnet, I guess we should not check whether there exists an external node and start the testnet not with the current port. So we should also provide an API to switch the network type and keep the mainnet and testnet remote the same. I guess they have the same complex with only switch network ID and use `switchCurrentNetworkType` to switch network type.

packages/neuron-ui/src/components/NetworkSetting/index.tsx
yanguoyu: Fixed by https://github.com/nervosnetwork/neuron/pull/2921/commits/92ae955c62d62bbcc1d4870593bc0f9f57e16e46

Keith-CY:
> > > > > > And as mentioned at [#2921 (comment)](https://github.com/nervosnetwork/neuron/pull/2921#discussion_r1382425727)
> > > > > > switch network type can be transformed into switch network id in the UI, e.g.
> > > > > 
> > > > > 
> > > > > So, some networks will be hidden? When the user chooses the mainnet, the testnet network will be hidden. Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> > > > 
> > > > 
> > > > Similar to being hidden image Mainnet for internal node and testnet for internal node(two options) are shrinked into one network option with a type switch
> > > > In the case of the screenshot, `switch to mainnet` will be treated as `switch to network id ckb_mainnet`
> > > > > Then I can remove the remote API `switchCurrentNetworkType`. Is my understanding correct?
> > > > 
> > > > 
> > > > Sure, there's no need for one more API because `switch network id` and `switch network type` do the same work under the hood.
> > > 
> > > 
> > > I have tried that, but It seems there are some differences. When we switch the network ID, we should verify and notify the users when an external node is running. But when we switch the internal network type, it's unnecessary and just starts with an internal node.
> > 
> > 
> > I didn't get the point. There are two independent processes.
> > 
> > 1. Switch network, it sends a message to the background to update the active network, and happens before the node is connected.
> > 2. Verify the connected node, it checks if the `internal network option` is active while the `connected node` is outside Neuron, happens after the node is connected.
> > 
> > They are 2 different logic happening before and after the node is connected, how will they be coupled.
> > Maybe you're mentioning that `verification` can be skipped when `switch chain type`, but there's an assumption that `switch chain type` definitely connects to the internal node, I'm not sure about that(because they are two independent processes IMO).
> 
> When the user starts the mainnet with another port, currently, if they want to switch to the mainnet, I guess we should not check whether there exists an external node and start the testnet not with the current port. So we should also provide an API to switch the network type and keep the mainnet and testnet remote the same. I guess they have the same complex with only switch network ID and use `switchCurrentNetworkType` to switch network type.

We had a discussion on meeting and there's no disagreement now

Keith-CY:
@WhiteMinds @devchenyan @zhangyouxin @homura please have a review

packages/neuron-ui/src/locales/zh-tw.json
```diff
@@ -1108,6 +1110,15 @@
       "input-place-holder": "輸入同步起始區塊高度，例如：10,100,101",
       "locate-first-tx": "定位第一筆交易",
       "view-block": "查看區塊"
+    },
+    "main": {
+      "external-node-detected-dialog": {
+        "title": "檢測到外部節點",
+        "body-tips-without-network": "\"Internal Node\" 僅用於連接內置節點，請添加新的網絡選項以連接外部節點。",
+        "body-tips-with-network": "\"Internal Node\" 僅用於連接內置節點，請從選擇以下網絡選項或添加新的網絡選項以連接外部節點。",
+        "add-network": "添加網絡",
```
devchenyan, line 1119: 僅用於連接內置節點，請`從`選擇以下網絡選項或添加新的網絡選項以連接外部節點

`從`  is redundant

packages/neuron-ui/src/locales/zh.json
```diff
@@ -1129,6 +1131,15 @@
       "input-place-holder": "输入同步起始区块高度，例如：10,100,101",
       "locate-first-tx": "定位第一笔交易",
       "view-block": "查看区块"
+    },
+    "main": {
+      "external-node-detected-dialog": {
+        "title": "检测到外部节点",
+        "body-tips-without-network": "\"Internal Node\" 仅用于连接内置节点，请添加新的网络选项以连接外部节点。",
+        "body-tips-with-network": "\"Internal Node\" 仅用于连接内置节点，请从选择以下网络选项或添加新的网络选项以连接外部节点。",
+        "add-network": "添加网络",
```
devchenyan, line 1140: 仅用于连接内置节点，请`从`选择以下网络选项或添加新的网络选项以连接外部节点

从 is redundant

packages/neuron-wallet/src/services/settings.ts
```diff
@@ -84,16 +80,34 @@ export default class SettingsService extends Store {
...
     updateApplicationMenu(null)
   }
+
+  migrateDataPath() {
+    const networkChain = NetworksService.getInstance().getCurrent()?.chain
+    const currentCkbDataPath = this.readSync(settingKeys.ckbDataPath)
+    this.writeSync(`${settingKeys.nodeDataPath}_${networkChain}`, currentCkbDataPath)
+    const defaultMainNetworkDir = path.resolve(app.getPath('userData'), 'chains/mainnet')
+    if (networkChain === 'ckb') {
+      this.writeSync(`${settingKeys.nodeDataPath}_ckb_testnet`, path.resolve(app.getPath('userData'), 'chains/testnet'))
```
Keith-CY, line 99: Is this still necessary when full node testnet is not included in this app

devchenyan (approved):

packages/neuron-ui/src/locales/en.json
```diff
@@ -1137,6 +1139,15 @@
       "input-place-holder": "Please enter the start sync block number, eg: 10,100,101",
       "locate-first-tx": "Locate the first transaction",
       "view-block": "View the block"
+    },
+    "main": {
+      "external-node-detected-dialog": {
+        "title": "External node detected",
+        "body-tips-without-network": "\"Internal node\" is reserved for the built-in CKB node, please add a new network option for the external node.",
```
Keith-CY, line 1146:
```suggestion
        "body-tips-without-network": "\"Internal Node\" is reserved for the built-in CKB node, please add a new network option for the external node.",
```

packages/neuron-ui/src/services/remote/remoteApiWrapper.ts
```diff
@@ -91,6 +91,7 @@ type Action =
   | 'get-current-network-id'
   | 'set-current-network-id'
   | 'delete-network'
+  | 'switch-network-type'
```
Keith-CY, line 94: Is `switch-network-type` still used?

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
```diff
@@ -92,7 +92,11 @@ export default class LightConnector extends Connector<CKBComponents.Hash> {
   private async fetchDepCell() {
     const depGroupOutputsData: string[] = await this.getDepTxs()
     const depGroupTxHashes = [
-      ...new Set(depGroupOutputsData.map(v => unpackGroup.unpack(v).map(v => v.tx_hash.toHexString())).flat()),
+      ...new Set(
+        depGroupOutputsData
+          .map(v => unpackGroup.unpack(v).map(v => `0x${v.tx_hash.toString(16).padStart(64, '0')}`))
```
Keith-CY, line 97: In what case `tx hash` should be padded

packages/neuron-ui/src/components/NetworkSetting/index.tsx
```diff
@@ -1,17 +1,27 @@
...
 import RadioGroup from 'widgets/RadioGroup'
 import { useOnWindowResize, useToggleChoiceGroupBorder, getNetworkLabelI18nkey } from 'utils'
+import { AddSimple, Switch } from 'widgets/Icons/icon'
+import Tooltip from 'widgets/Tooltip'
+import { LIGHT_CLIENT_MAINNET, NetworkType } from 'utils/const'
+import { lastShowInternalNodeIds } from 'services/localCache'
 import styles from './networkSetting.module.scss'
 
+const getAnotherNetworkType = (chain: State.Network['chain']): 'testnet' | 'mainnet' => {
+  if (chain === 'ckb' || chain === LIGHT_CLIENT_MAINNET) {
```
Keith-CY, line 19:
```suggestion
  // REFACTOR ME: MAINNET_CLIENT_LIST can be used here.
  if (chain === 'ckb' || chain === LIGHT_CLIENT_MAINNET) {
```

packages/neuron-wallet/src/controllers/networks/index.ts
```diff
@@ -19,6 +19,7 @@ export default class NetworksController {
       .pipe(distinctUntilChanged(), skip(1))
       .subscribe(async (connected: boolean) => {
         if (connected) {
+          logger.debug('Network:\tconnection success')
```
Keith-CY, line 22:
```suggestion
          logger.debug('Network:\tconnected successfully')
```

Keith-CY (approved):

packages/neuron-wallet/src/services/settings.ts
yanguoyu: Only the `else` branch is necessary, the `ckb_testnet` data path can be removed. I will remove it.

packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts
yanguoyu: I use wrong type with Lumos, https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/block-sync-renderer/sync/light-connector.ts#L28-L36 , I will change these to `const unpackGroup = molecule.vector(blockchain.OutPoint)`

silySuper:
https://github.com/Magickbase/neuron-public-issues/issues/313#issuecomment-1797186520

PR #2928, opened 2023-11-10
fix: Avoid unexpected dialog shown.

yanguoyu:
To avoid unexpected dialog that notifies the user of an external node shown.
1. Set the process to null when the started pid equals the current process pid.
2. Stop the get tip number when switching the network and start after starting the node.
3. Add debounce time to start node, that is avoid the user switching the internal network faster.

### DIFF
--- a/packages/neuron-ui/src/utils/getSyncStatus.ts
+++ b/packages/neuron-ui/src/utils/getSyncStatus.ts
@@ -5,6 +5,7 @@ const TEN_MINS = 10 * 60 * 1000
 let blockNumber10MinAgo: number = -1
 let timestamp10MinAgo: number | undefined
 let prevUrl: string | undefined
+let prevNetworkID: string | undefined
 
 export const getSyncStatus = ({
   bestKnownBlockNumber,
@@ -12,17 +13,23 @@ export const getSyncStatus = ({
   cacheTipBlockNumber,
   currentTimestamp,
   url,
+  networkID,
 }: {
   bestKnownBlockNumber: number
   bestKnownBlockTimestamp: number
   cacheTipBlockNumber: number
   currentTimestamp: number
   url: string | undefined
+  networkID: string
 }) => {
-  if ((!timestamp10MinAgo && bestKnownBlockNumber >= 0) || (prevUrl && url !== prevUrl && bestKnownBlockNumber >= 0)) {
+  if (
+    (!timestamp10MinAgo && bestKnownBlockNumber >= 0) ||
+    (((prevUrl && url !== prevUrl) || (prevNetworkID && prevNetworkID !== networkID)) && bestKnownBlockNumber >= 0)
+  ) {
     timestamp10MinAgo = currentTimestamp
     blockNumber10MinAgo = bestKnownBlockNumber
     prevUrl = url
+    prevNetworkID = networkID
   }
 
   const now = Math.floor(currentTimestamp / 1000) * 1000
--- a/packages/neuron-ui/src/utils/getSyncStatus.ts
+++ b/packages/neuron-ui/src/utils/getSyncStatus.ts
@@ -23,7 +23,10 @@ export const getSyncStatus = ({
   networkID: string
 }) => {
   if (
+    // !timestamp10MinAgo means start sync for the first time
     (!timestamp10MinAgo && bestKnownBlockNumber >= 0) ||
+    // prevUrl for change the network remote(change network or eidt the network)
+    // prevNetworkID for change the network, sometime change network the remote is same. light client mainnet -> testnet
     (((prevUrl && url !== prevUrl) || (prevNetworkID && prevNetworkID !== networkID)) && bestKnownBlockNumber >= 0)
   ) {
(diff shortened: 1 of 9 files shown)


packages/neuron-wallet/src/services/ckb-runner.ts
```diff
@@ -142,18 +142,24 @@ export const startCkbNode = async () => {
     }
   })
 
-  ckb.on('error', error => {
+  currentProcess.on('error', error => {
     logger.error('CKB:\trun fail:', error)
     isLookingValidTarget = false
-    ckb = null
+    if (ckb?.pid === currentProcess.pid) {
```
Keith-CY, line 148: `Object.is` should be used here based on the assignment at line 161

```js
ckb = currentProcess
```

And the `ckb = null` when `ckb?.pid === currentProcess.pid` is a bit weird because it implies there will be 2 processes spawned by Neuron

What if send an `exit` signal to the previous one, then spwan a new one.

Besides, why not `ckb = currentProcess` when `ckb.pid !== currentProcess.pid`

Keith-CY:
In what situations will the dialog be displayed mistakenly?

packages/neuron-wallet/src/services/ckb-runner.ts
yanguoyu: If the `ckb` is set as null by another event, the ckb can't stop forever by Neuron.
When the user switches the node faster, sometimes the `ckb.pid !== currentProcess.pid`.

yanguoyu:
> In what situations will the dialog be displayed mistakenly?

When the `connectionStatusSubject` broadcasts true for the last selected node. So I restart the `subscribe` which gets the tip number when the network switches.

Keith-CY:
> > In what situations will the dialog be displayed mistakenly?
> 
> When the `connectionStatusSubject` broadcasts true for the last selected node. So I restart the `subscribe` which gets the tip number when the network switches.

Got it

packages/neuron-wallet/src/services/ckb-runner.ts
Keith-CY: > If the `ckb` is set as null by another event, the ckb can't stop forever by Neuron. When the user switches the node faster, sometimes the `ckb.pid !== currentProcess.pid`.

If so, I think the fix should be replacing `set ckb as null` with `stop ckb`

packages/neuron-wallet/src/services/ckb-runner.ts
```diff
@@ -142,18 +142,24 @@ export const startCkbNode = async () => {
     }
   })
 
-  ckb.on('error', error => {
+  currentProcess.on('error', error => {
     logger.error('CKB:\trun fail:', error)
     isLookingValidTarget = false
-    ckb = null
+    if (Object.is(ckb, currentProcess)) {
+      ckb = null
```
Keith-CY, line 149: I think `stop signal` should be sent to process `ckb` even if it has been sent, because the reason `ckb` and `currentProcess` conflict here is unknown.

packages/neuron-wallet/src/services/ckb-runner.ts
yanguoyu: https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/services/ckb-runner.ts#L160-L171
The ckb has been set as null when stopping the ckb, and it will send a `close` event to the `currentProcess`, so here should add a judgment to avoid setting another ckb to null.

packages/neuron-wallet/src/services/ckb-runner.ts
yanguoyu: Setting the `ckb` as null can not stop the process, I guess adding a condition to set it as null have no side effect.

Keith-CY: Ah, I found I recognized the condition as `!Object.is(ckb, currentProcess)` mistakenly. Now I have no questions

Keith-CY (approved):

packages/neuron-ui/src/utils/getSyncStatus.ts
```diff
@@ -5,24 +5,31 @@ const TEN_MINS = 10 * 60 * 1000
...
   bestKnownBlockTimestamp: number
   cacheTipBlockNumber: number
   currentTimestamp: number
   url: string | undefined
+  networkID: string
 }) => {
-  if ((!timestamp10MinAgo && bestKnownBlockNumber >= 0) || (prevUrl && url !== prevUrl && bestKnownBlockNumber >= 0)) {
+  if (
+    (!timestamp10MinAgo && bestKnownBlockNumber >= 0) ||
+    (((prevUrl && url !== prevUrl) || (prevNetworkID && prevNetworkID !== networkID)) && bestKnownBlockNumber >= 0)
```
homura, line 27: it's better to give meaning to such combined and complicated conditions, e.g.

```ts
if (
  /*sync is started*/ (!timestamp10MinAgo && bestKnownBlockNumber >= 0) ||
  /*network is changed*/ (...)
)
```

PR #2932, opened 2023-11-13
feat: Enable click max when address is empty.

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/308. Use placeholder address to generate tx.

https://github.com/nervosnetwork/neuron/assets/12881040/0187bec5-f50a-45f7-ba27-0e071827649c

### DIFF
--- a/packages/neuron-ui/src/components/Send/hooks.ts
+++ b/packages/neuron-ui/src/components/Send/hooks.ts
@@ -10,7 +10,10 @@ import {
   shannonToCKBFormatter,
   calculateFee,
   validateOutputs,
+  DefaultLockInfo,
 } from 'utils'
+import { scriptToAddress } from '@nervosnetwork/ckb-sdk-utils'
+import { PlaceHolderArgs } from 'utils/const'
 
 let generateTxTimer: ReturnType<typeof setTimeout>
 
@@ -330,10 +333,7 @@ export const useInitialize = (
   const [errorMessage, setErrorMessage] = useState('')
   const [isSendMax, setIsSendMax] = useState(false)
 
-  const outputs = useMemo(
-    () => items.map(item => ({ ...item, disabled: isSendMax || sending })),
-    [items, isSendMax, sending]
-  )
+  const outputs = useMemo(() => items.map(item => ({ ...item, disabled: sending })), [items, sending])
 
   const updateIsSendMax = useCallback(
     (payload: boolean) => {
@@ -358,7 +358,19 @@ export const useInitialize = (
   const updateSendingAllTransaction = useCallback(() => {
     updateTransactionWith(generateSendingAllTx)({
       walletID,
-      items,
+      items: items.map(v => ({
+        ...v,
+        address:
+          v.address ||
+          scriptToAddress(
+            {
+              codeHash: DefaultLockInfo.CodeHash,
+              hashType: DefaultLockInfo.HashType,
+              args: PlaceHolderArgs,
+            },
+            isMainnet
+          ),
+      })),
       price,
       setTotalAmount,
       setErrorMessage,
@@ -413,6 +425,7 @@ export const useInitialize = (
     setErrorMessage,
     isSendMax,
     onSendMaxClick,
+    updateIsSendMax,
   }
 }
 
--- a/packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
+++ b/packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
@@ -8,11 +8,10 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: false,
-      ignoreLastAddress: false,
     },
     exception: false,
   },
-  'Should pass when last address is empty and ignore last address': {
+  'Should throw an error when address is not provided': {
     params: {
       outputs: [
         {
@@ -21,20 +20,6 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: false,
-      ignoreLastAddress: true,
-    },
-    exception: false,
-  },
-  'Should throw an error when not the last address is not provided': {
-    params: {
-      outputs: [
-        {
-          address: '',
-          amount: '100',
-        },
-      ],
-      ignoreLastAmount: false,
-      ignoreLastAddress: false,
     },
     exception: true,
   },
@@ -47,7 +32,6 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: false,
-      ignoreLastAddress: false,
     },
     exception: true,
   },
@@ -60,7 +44,6 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: false,
-      ignoreLastAddress: false,
     },
     exception: true,
   },
@@ -73,7 +56,6 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: false,
-      ignoreLastAddress: false,
     },
     exception: true,
   },
@@ -86,7 +68,6 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: false,
-      ignoreLastAddress: false,
     },
     exception: true,
   },
@@ -99,7 +80,6 @@ const fixtures = {
         },
       ],
       ignoreLastAmount: true,
-      ignoreLastAddress: false,
     },
     exception: false,
   },
--- a/packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
+++ b/packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
@@ -21,7 +21,7 @@ const fixtures = {
       ],
       ignoreLastAmount: false,
     },
-    exception: true,
+    exception: false,
   },
   'Should throw an error when address is invalid': {
     params: {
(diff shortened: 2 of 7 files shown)


packages/neuron-ui/src/components/Send/hooks.ts
```diff
@@ -10,7 +10,10 @@ import {
   shannonToCKBFormatter,
   calculateFee,
   validateOutputs,
+  DefaultLockInfo,
 } from 'utils'
+import { scriptToAddress } from '@nervosnetwork/ckb-sdk-utils'
```
Keith-CY, line 15: `@nervosnetwork/ckb-sdk-utils` has been migrated to `lumos`

packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
```diff
@@ -21,7 +21,7 @@ const fixtures = {
       ],
       ignoreLastAmount: false,
     },
-    exception: true,
+    exception: false,
```
Keith-CY, line 24: The title is `should throw an error when address is not provided` doesn't match `exception: false`

Maybe a specific case of `empty address with checked "send max"` should be added.

packages/neuron-ui/src/utils/validators/outputs.ts
```diff
@@ -10,7 +10,9 @@ export const validateOutputs = (
 ) => {
   for (let i = 0; i < items.length; i++) {
     try {
-      validateAddress(items[i].address || '', isMainnet)
+      if (items[i].address) {
```
Keith-CY, line 13: Only the last address is allowed to be empty because only the last fieldset can be set `send max`

packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
yanguoyu: Added a test case to check that `ignoreLastAddress` is true.

packages/neuron-ui/src/components/Send/hooks.ts
yanguoyu: The `UI` package has not imported Lumos. Maybe we should remove it by the refactor that has removed`@nervosnetwork/ckb-sdk-utils`.

yanguoyu: https://github.com/Magickbase/neuron-public-issues/issues/201#issuecomment-1809635406

packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
Keith-CY: The condition becomes complex with adding a new field `ignoreLastAddress` because the combination of `ignoreLastAmount && ignoreLastAddress` includes 4 cases but only `false && false`, `false && true`, `true && false` are covered.

What if using one flag to skip validation of the last fieldset

yanguoyu: That reminds me! I added `ignoreLastAddress` to ignore the last items when checking whether the last max button is disabled.
How about removing the last item to check the max button's status? Then the `validateOutputs` is unnecessary to change. https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/Send/index.tsx#L137

Keith-CY: I didn't get the meaning of `removing the last item to check the max button's status`

yanguoyu: https://github.com/nervosnetwork/neuron/pull/2932/files#diff-40f3dad0e46ccc5c6d81f579bd16c048e233df26fc686e22a864b55f75da8522R151-R156
Remove the last item of output when checking the max button's status.

Keith-CY: > [#2932 (files)](https://github.com/nervosnetwork/neuron/pull/2932/files#diff-40f3dad0e46ccc5c6d81f579bd16c048e233df26fc686e22a864b55f75da8522R151-R156) Remove the last item of output when checking the max button's status.

LGTM

Keith-CY (approved):

packages/neuron-ui/src/tests/validators/outputs/fixtures.ts
Keith-CY: The logic that `outputs.slice(0, -1)` can be reused where `ignoreLastAmount = true` so the logic of `validateOutputs` can be simplified. We can update it later.

devchenyan (approved):

silySuper:
When click max button repeatedly，an total amount will show non-inevitably and ui become abnormal. @yanguoyu 

https://github.com/nervosnetwork/neuron/assets/25971273/653534b6-88dd-4a3c-85a2-a9e387df09d2

yanguoyu:
> When click max button repeatedly，an total amount will show non-inevitably and ui become abnormal. @yanguoyu
> 
>  2023-11-15.10.37.41.mov

How about click Max after syncing to 100%?

silySuper:
> > When click max button repeatedly，an total amount will show non-inevitably and ui become abnormal. @yanguoyu
> > 2023-11-15.10.37.41.mov
> 
> How about click Max after syncing to 100%?

When sync 100%，the balance does not change again and again,so abnormal  ui does not show  up.Is the reason?

yanguoyu:
> > > When click max button repeatedly，an total amount will show non-inevitably and ui become abnormal. @yanguoyu
> > > 2023-11-15.10.37.41.mov
> > 
> > 
> > How about click Max after syncing to 100%?
> 
> When sync 100%，the balance does not change again and again,so abnormal ui does not show up.Is the reason?

Maybe some transaction was not synced, then the max balance will be larger than the actual.

silySuper:
An small strange point is that :when click max ,the total blance is filled.When click again,it shows like this:
截屏2023-11-15 11 03 01

Maybe click max button second time has cancel input box number and come to primitive status like this is better,isn 't it ? @Danie0918 
截屏2023-11-15 11 06 09

yanguoyu:
> An small strange point is that :when click max ,the total blance is filled.When click again,it shows like this: 截屏2023-11-15 11 03 01
> 
> Maybe click max button second time has cancel input box number and come to primitive status like this is better,isn 't it ? @Danie0918 截屏2023-11-15 11 06 09

IMO, There are differences between `cancel max` with `initialization `, just like entering the username and deleting all the input, we will notify the users that `please enter username`.

image

image

Danie0918:
> Maybe click max button second time has cancel input box number and come to primitive status like this is better,isn 't it ?

For a button, its operation should match the description. There are 2 options here:

1. disable repeat clicks.
2. when clicked again, still perform the action described by the button.

Personally, I prefer the second one, which allows the user to recapture the maximum value when their balance changes.

yanguoyu:
> > Maybe click max button second time has cancel input box number and come to primitive status like this is better,isn 't it ?
> 
> For a button, its operation should match the description. There are 2 options here:
> 
> 1. Disable repeat clicks.
> 2. When clicked again, still perform the action described by the button.
> 
> Personally, I prefer the second one, which allows the user to recapture the maximum value when their balance changes.

Currently, the max button has two functions:
1. Send max if first click
2. Cancel send max if currently is sending max.

Do you mean we'd better cancel sending max by another action? @Danie0918

Danie0918:
> Currently, the max button has two functions:
> 
> 1. Send max if first click
> 2. Cancel send max if currently is sending max.
> 
> Do you mean we'd better cancel sending max by another action?

Canceling the maximum value can be manually modified or deleted by the user, and the max button should be to enter the maximum value.If the cancel function is implemented via a button, the button should change to cancel when clicked.

yanguoyu:
> > Currently, the max button has two functions:
> > 
> > 1. Send max if first click
> > 2. Cancel send max if currently is sending max.
> > 
> > Do you mean we'd better cancel sending max by another action?
> 
> Canceling the maximum value can be manually modified or deleted by the user, and the max button should be to enter the maximum value.If the cancel function is implemented via a button, the button should change to cancel when clicked.
> 
> IMO,The Max button should be entered as the maximum value even if it is clicked repeatedly.

How about changing `Max` to `Reset` if the current is sending the max? @Danie0918

Danie0918:
> > > Currently, the max button has two functions:
> > > 
> > > 1. Send max if first click
> > > 2. Cancel send max if currently is sending max.
> > > 
> > > Do you mean we'd better cancel sending max by another action?
> > 
> > 
> > Canceling the maximum value can be manually modified or deleted by the user, and the max button should be to enter the maximum value.If the cancel function is implemented via a button, the button should change to cancel when clicked.
> > IMO,The Max button should be entered as the maximum value even if it is clicked repeatedly.
> 
> How about changing `Max` to `Reset` if the current is sending the max? @Danie0918

Okay, that's better.

silySuper:
When language is Chinese it shows  like this：


https://github.com/nervosnetwork/neuron/assets/25971273/0cc117a2-d6c9-47cc-b6d9-fde6dfaf9352

Why not shows "最大值" in Chinese？

yanguoyu:
> When language is Chinese it shows like this：
> 
>  2023-11-17.11.07.31.mov 
> Why not shows "最大值" in Chinese？

The tips include `Max`. 
```
您需要有足够的余额来支付找零（大于 61 CKBytes），或者点击 'Max' 按钮发送全部余额。
```

silySuper (approved):

PR #2933, opened 2023-11-13
feat: Unpack the app to `app.asar` for test mode.

yanguoyu:


.github/workflows/package_for_test.yml
```diff
@@ -122,6 +122,7 @@ jobs:
           CSC_LINK: ${{ secrets.MAC_CERTIFICATE_BASE64 }}
           CSC_KEY_PASSWORD: ${{ secrets.MAC_CERTIFICATE_PASSWORD }}
           TEAM_ID: ${{ secrets.TEAM_ID }}
+          USE_HARD_LINKS: false
```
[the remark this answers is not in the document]
yanguoyu: To fix the build error `file already exists ` https://github.com/electron-userland/electron-builder/issues/3179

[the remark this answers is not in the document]
Keith-CY: Got it

packages/neuron-wallet/electron-builder.test.yml
```diff
@@ -0,0 +1,96 @@
...
+directories:
+  buildResources: .
+  output: ../../release
+
+afterSign: scripts/notarize.js
+
+files:
+  - from: "../.."
+    to: "."
+    filter: ["!**/*", ".ckb-version", ".ckb-light-version", "ormconfig.json", "compatible.csv"]
```
Keith-CY, line 16: The `compaitble.csv` should be `compatible.json`. And introducing one more template increases the burden of maintenance. How about setting some fields configurable and generating the configuration on the fly.

yanguoyu: Of course, if we import a lib that can parse and write a `yml` file to create the test yml from the original yml file, it will be better.

yanguoyu:
Now, I use the GitHub action script to create a test build yml file. https://github.com/nervosnetwork/neuron/pull/2933/files#diff-1735bbbe6061183b60bbab86717798ebe202196cdf93573e972973c701114851R113-R120

.github/workflows/package_for_test.yml
```diff
@@ -110,6 +110,15 @@ jobs:
...
           CI: false
 
+      - name: Write electron-build Test yml
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('node:fs')
+            const prodYmlPath = 'packages/neuron-wallet/electron-builder.yml'
+            const testYmlPath = 'packages/neuron-wallet/electron-builder.test.yml'
+            fs.writeFileSync(testYmlPath, fs.readFileSync(prodYmlPath).toString().replace('asar: true', 'asar: false'))
```
Keith-CY, line 120: It looks good, but I wonder why not override the original `yaml` file. Only one file need to be updated by doing so.

Keith-CY (approved):

Keith-CY (approved):

.github/workflows/package_for_test.yml
```diff
@@ -110,6 +110,14 @@ jobs:
         env:
           CI: false
 
+      - name: Write electron-build Test yml
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('node:fs')
+            const ympPath = 'packages/neuron-wallet/electron-builder.yml'
+            fs.writeFileSync(ympPath, fs.readFileSync(ympPath).toString().replace('asar: true', 'asar: false'))
```
homura, line 119: I'd suggest adding a comment in `electron-builder.yml` to make the maintainers notice the source code can be included in a development build

https://github.com/nervosnetwork/neuron/blob/6640e21ea970da3913ccf826ab83d3aea6f646db/packages/neuron-wallet/electron-builder.yml#L5C7-L5C11

homura (approved):

.github/workflows/package_for_test.yml
yanguoyu: Added at https://github.com/nervosnetwork/neuron/pull/2933/files#diff-510827a6f662e6ba2ea07421fbed486a0a38d996a85a9bd3c73a0e241d6f62a7R5

PR #2935, opened 2023-11-14
fix: Fix migrate settings and network.

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/320.

#### How to reproduce this problem

Delete the settings and network file, download the 0.103.1 Neuron, and start. Then update to 0.110.1.

The settings file location:
Mac OS: /Users/${username}/Library/Application Support/Neuron/settings.json
Windows: C:/Users/${username}/AppData/Roaming/Neuron/settings.json

The network file location:
Mac OS: /Users/${username}/Library/Application Support/Neuron/networks/index.json
Windows: C:/Users/${username}/AppData/Roaming/Neuron/networks/index.json

### DIFF
--- a/packages/neuron-wallet/src/services/networks.ts
+++ b/packages/neuron-wallet/src/services/networks.ts
@@ -66,7 +66,8 @@ enum NetworksKey {
 }
 
 const oldDefaultNames = ['Default', 'default node', presetNetworks.networks[0].name]
-// Before 0.106.0 version the default remote's value is http://localhost:8114.
+// Before 0.106.0 version the default remote's value is http://localhost:8114
+// "localhost" was deprecated because of https://github.com/Magickbase/neuron-public-issues/issues/122
 const oldDefaultRemotes = ['http://localhost:8114', BUNDLED_CKB_URL]
 
 export default class NetworksService extends Store {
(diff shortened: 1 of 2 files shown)


packages/neuron-wallet/src/services/networks.ts
```diff
@@ -66,6 +66,7 @@ enum NetworksKey {
 }
 
 const oldDefaultNames = ['Default', 'default node', presetNetworks.networks[0].name]
+const oldDefaultRemotes = ['http://localhost:8114', BUNDLED_CKB_URL]
```
Keith-CY, line 69: `http://localhost:8114` is magical, maybe it can be defined in `constants` with why `localhost` is deprecated.

yanguoyu: `oldDefaultRemotes` is a const variable, I guess adding a comment above the `oldDefaultRemotes` is enough.

Keith-CY: LGTM

packages/neuron-wallet/src/services/networks.ts
```diff
@@ -66,6 +66,8 @@ enum NetworksKey {
 }
 
 const oldDefaultNames = ['Default', 'default node', presetNetworks.networks[0].name]
+// Before 0.106.0 version the default remote's value is http://localhost:8114.
```
Keith-CY, line 69:
```suggestion
// Before 0.106.0 version the default remote's value is http://localhost:8114.
// "localhost" was deprecated because of https://github.com/Magickbase/neuron-public-issues/issues/122
```

Keith-CY (approved):

devchenyan (approved):

silySuper (approved):

PR #2962, opened 2023-11-26
ci: add an action to generate checksum table from release draft

Keith-CY:
This commit adds an action to generate checksum table from release draft.

The action runs a script to download assets into a temporary directory, then run a script to load binaries to generate the checksum table and print it in console.

Ref: https://github.com/Magickbase/neuron-public-issues/issues/326


yanguoyu (approved):

Keith-CY:
@homura @zhangyouxin @WhiteMinds @devchenyan please have a review

.github/workflows/checksums-of-release-draft.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          mkdir -p /tmp/Neuron
+          node ./scripts/download-binaries-from-release-draft.js /tmp/Neuron
+        shell: pwsh
+        env:
+          TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+
+      - name: Checksums
+        run: node ./scripts/generate-checksum-table.js /tmp/Neuron
```
homura, line 33: I found that the script finishes after writing a `checksum.txt` without an upload to GitHub action. Is it expected?

Keith-CY: Yes, the `checksum.txt` is an add-on for custom distribution, it will be generated in the directory where binaries are, so the directory can be published directly(e.g., google driver) and users may check the checksum table online.

In this workflow, the checksum table will be sent to the stdout so I can copy&paste it into the release note.

When we add an action to generate a release note automatically, the checksum table in stdout can be redirected to the next workflow.

homura: thx for your explanation, the distribution workflow will still evolve later, and this is not the final version

PR #2964, opened 2023-11-29
refactor: upgrade lumos to 0.21

homura:
This PR upgraded lumos to 0.21 to work with the `fetch` API


packages/neuron-wallet/src/utils/ckb-rpc.ts
```diff
@@ -2,7 +2,7 @@ import type { ParamsFormatter } from '@ckb-lumos/rpc'
 import type { Block } from '@ckb-lumos/base'
 import { HexString } from '@ckb-lumos/base'
 import { CKBRPC } from '@ckb-lumos/rpc'
-import { Method } from '@ckb-lumos/rpc/lib/method'
+import { Method as Method_ } from '@ckb-lumos/rpc/lib/method'
```
Keith-CY, line 5: Does `Method_` conform to a specific naming convention?

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -855,7 +855,9 @@ export default class TransactionSender {
     const rpc = generateRPC(currentNetwork.remote, currentNetwork.type)
 
     let tx = await rpc.getTransaction(depositOutPoint.txHash)
-    if (tx.txStatus.status !== 'committed') throw new Error('Transaction is not committed yet')
+    if (tx.txStatus.status !== 'committed' || !tx.txStatus.blockHash) {
```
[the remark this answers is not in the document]
homura: I am not sure. But I think we'd better stay the same type description with the [rpc-types](https://github.com/nervosnetwork/ckb/blob/860d56c94c386a5d82be6d7f9f50e38420892f88/util/jsonrpc-types/src/blockchain.rs#L584-L591)

yanguoyu (approved):

packages/neuron-wallet/src/services/transaction-sender.ts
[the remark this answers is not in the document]
Keith-CY: Reasonable

packages/neuron-wallet/src/utils/ckb-rpc.ts
homura: i've renamed to `SdkRpcMethod` for more understandable

Keith-CY (approved):

packages/neuron-wallet/src/utils/ckb-rpc.ts
```diff
@@ -11,11 +11,11 @@ import {
   PayloadInBatchException,
   IdNotMatchedInBatchException,
 } from '@ckb-lumos/rpc/lib/exceptions'
-import https from 'https'
-import http from 'http'
 import { request } from 'undici'
 import CommonUtils from './common'
 import { NetworkType } from '../models/network'
+import { RPCConfig } from '@ckb-lumos/rpc/lib/types/common'
+import { CKBComponents } from '@ckb-lumos/rpc/lib/types/api'
```
Keith-CY, line 18: These two imports should be marked as `import type` and moved to the top

packages/neuron-wallet/src/utils/ckb-rpc.ts
```diff
@@ -131,6 +131,12 @@ export class FullCKBRPC extends CKBRPC {
   }
 }
 
+class Method extends SdkRpcMethod {
```
Keith-CY, line 134: Maybe `SdkRpcMethod` can be used directly

Keith-CY:
type `Node`(https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/types/ckbComponents.d.ts#L47-L48) could be updated because `httpAgent` and `httpsAgent` are removed

packages/neuron-wallet/src/utils/ckb-rpc.ts
homura: the super bond a `rpcConfig` as `super(node, options, rpcConfig)`, so we use an extended class here

packages/neuron-wallet/src/utils/ckb-rpc.ts
homura: switched to `import type`

Keith-CY (approved):

PR #2977, opened 2023-12-07
chore: devtools for developing with a devnet

homura:
This PR is used for working with a devnet. Honestly, I am not sure it's really useful.

The tool was created while developing something unpublished, such as the spore NFT. Recently, I tried to make Neuron's offline-sign feature work with unidentified locks, and these scripts were used again. Therefore, I created this PR hoping to help developers who are experiencing a similar scenario to me.


yanguoyu:
It's a bit hard for me to understand how to use this with Neuron. If I have the full data about testnet, do I also need this?
Additionally, how about providing full dev node data to test with the dev node? Like the E2E test dev node data.

homura:
> It's a bit hard for me to understand how to use this with Neuron. If I have the full data about testnet, do I also need this?

If you already have the full data of testnet, you likely don't need this tool. But there are some exceptions, for example, where a contract has not been deployed. Or if a large number of faucet tokens are needed, and the faucet is not enough to support the test. Another useful feature is quick mining, developers can adjust the `miner.workers.value` and `specs.params.epoch_duration_target=1` to short block interval or epoch duration, this is more flexible than the testnet

> Additionally, how about providing full dev node data to test with the dev node? Like the E2E test dev node data.

Sorry, I don't quite understand this part, could you provide more details?

yanguoyu:
> > Additionally, how about providing full dev node data to test with the dev node? Like the E2E test dev node data.
> 
> Sorry, I don't quite understand this part, could you provide more details?

Provider dev data that includes all of the contracts that Neuron needs. As I know, the E2E test will provide dev data that includes all contracts.

homura:
> Provider dev data that includes all of the contracts that Neuron needs. As I know, the E2E test will provide dev data that includes all contracts.




> Provider dev data that includes all of the contracts that Neuron needs. As I know, the E2E test will provide dev data that includes all contracts.

I guess what you mean is that not all contracts are included in the devtools, therefore, it is better to contain all contracts, such as sudt, cheque, m-NFT, etc. to make the devtools more useful, did I catch you?

Keith-CY:
> > Provider dev data that includes all of the contracts that Neuron needs. As I know, the E2E test will provide dev data that includes all contracts.
> 
> > Provider dev data that includes all of the contracts that Neuron needs. As I know, the E2E test will provide dev data that includes all contracts.
> 
> I guess what you mean is that not all contracts are included in the devtools, therefore, it is better to contain all contracts, such as sudt, cheque, m-NFT, etc. to make the devtools more useful, did I catch your idea?

Maybe the E2E runtime environment can be used for this case
https://github.com/gpBlockchain/ckb-neuron-sync-test

homura:
> Maybe the E2E runtime environment can be used for this case
> [gpBlockchain/ckb-neuron-sync-test](https://github.com/gpBlockchain/ckb-neuron-sync-test)

Indeed, this is similar to the e2e scenario, we can build the development environment based on this e2e repo.

Keith-CY:
> > Maybe the E2E runtime environment can be used for this case
> > [gpBlockchain/ckb-neuron-sync-test](https://github.com/gpBlockchain/ckb-neuron-sync-test)
> 
> Indeed, this is similar to the e2e scenario, we can build the development environment based on this e2e repo.

The E2E setup will be moved into Neuron Repo except for the chain data, the chain data are expected to be shared among projects. But the setup should support synchronizing chain data from the E2E repo.

PR #2978, opened 2023-12-08
refactor: refactor sign method to allow unidentified lock scripts

Keith-CY:
The sign method is refactored to allow unidentified lock scripts
while skipped lock scripts are prevented outside the sign method

Notice that skipped lock scripts validation is not adopted when
a multisig transaction because it's explictly thrown
when no private key is found. It may be intended.

### DIFF
--- a/packages/neuron-wallet/src/services/transaction-sender.ts
+++ b/packages/neuron-wallet/src/services/transaction-sender.ts
@@ -67,12 +67,12 @@ export default class TransactionSender {
     walletID: string = '',
     transaction: Transaction,
     password: string = '',
-    skipLastInputs: boolean = true,
+    skipLastInput: boolean = true,
     skipSign = false
   ) {
     const tx = skipSign
       ? Transaction.fromObject(transaction)
-      : await this.sign(walletID, transaction, password, skipLastInputs)
+      : await this.sign(walletID, transaction, password, skipLastInput)
 
     return this.broadcastTx(walletID, tx)
   }
@@ -109,7 +109,7 @@ export default class TransactionSender {
     walletID: string = '',
     transaction: Transaction,
     password: string = '',
-    skipLastInputs: boolean = true,
+    skipLastInput: boolean = true,
     context?: RPC.RawTransaction[]
   ) {
     const wallet = this.walletService.get(walletID)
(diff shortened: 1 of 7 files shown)


yanguoyu (approved):

packages/neuron-wallet/src/services/transaction-sender.ts
```diff
@@ -183,13 +189,32 @@ export default class TransactionSender {
...
-      // A 65-byte empty signature used as placeholder
-      witnessesArgs[0].witnessArgs.setEmptyLock()
 
       const privateKey = findPrivateKey(witnessesArgs[0].lockArgs)
 
+      if (!privateKey) {
+        metadata.locks.skipped.add(lockHash)
+        witnessSigningEntries.forEach((entry, idx) => {
+          if (entry.lockHash === lockHash) {
+            const rawWitness = tx.witnesses[idx]
```
homura, line 208: Could the `tx.witnesses[idx]` be a `''` instead `'0x'`? If we are not sure about it, maybe we can replace this line with

```ts
const rawWitness = tx.witnesses[idx] || '0x'
```

Keith-CY: This branch is for skipped input/witness pair, and if the witness is a string, it will be returned without modification. From my perspective, it's weird to replace `''` with `'0x'` because it does some work but it's meaningless.

Keith-CY: How would you think of the explanation

homura: > This branch is for skipped input/witness pair, and if the witness is a string, it will be returned without modification. From my perspective, it's weird to replace '' with '0x' because it does some work but it's meaningless.

The reason I mentioned this is because I tried to skip unidentified locks locally but failed due to an empty [witness `""`](https://github.com/nervosnetwork/neuron/commit/27fcc6bd2dc45148b5f537af6d62abb231574e02#diff-6b4e52aade78db37f085688194f39382d8411f142f051ac17a199f52f6953babR248) in witnesses. And I was not quite sure which was wrong. If you haven't encountered the issue, maybe we don't need a default witness placeholder

homura (approved):

PR #2994, opened 2023-12-20
feat: Optimize the setting of sync start block in light client mode.

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/304.
1. Show the last setted start block number
image

2. If the current set start block number is less than before

image

3. If the current set start block number is bigger than the header tip block number

image

### DIFF
--- a/packages/neuron-ui/src/components/PageContainer/hooks.ts
+++ b/packages/neuron-ui/src/components/PageContainer/hooks.ts
@@ -1,8 +1,36 @@
-import { useCallback, useEffect, useState } from 'react'
+import { TFunction } from 'i18next'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { importedWalletDialogShown } from 'services/localCache'
-import { isDark, openExternal, updateWallet, setTheme as setThemeAPI } from 'services/remote'
+import { isDark, openExternal, setTheme as setThemeAPI, updateWalletStartBlockNumber } from 'services/remote'
 import { Migrate } from 'services/subjects'
 import { getExplorerUrl, isSuccessResponse } from 'utils'
+
+const waitConfirmTime = 5
+
+const useCountDown = (second: number) => {
+  const [countdown, setCountdown] = useState(0)
+  const countDecreaseIntervalRef = useRef<ReturnType<typeof setInterval>>()
+  const resetCountDown = useCallback(() => {
+    clearInterval(countDecreaseIntervalRef.current)
+    setCountdown(second)
+    const decrement = () => {
+      countDecreaseIntervalRef.current = setInterval(() => {
+        setCountdown(v => {
+          if (v > 0) {
+            return v - 1
+          }
+          clearInterval(countDecreaseIntervalRef.current)
+          return v
+        })
+      }, 1_000)
+    }
+    decrement()
+  }, [setCountdown, countDecreaseIntervalRef])
+  return {
+    countdown,
+    resetCountDown,
+  }
+}
 
 export const useSetBlockNumber = ({
   firstAddress,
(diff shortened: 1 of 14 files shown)


packages/neuron-ui/src/types/Controller/index.d.ts
```diff
@@ -35,7 +35,12 @@ declare namespace Controller {
     newPassword?: string
     name?: string
     device?: any
+  }
+
+  interface UpdateWalletStartBlockNumberParams {
+    id: string
     startBlockNumber?: string
+    clearAllData?: boolean
```
Keith-CY, line 43: Maybe `clearAllData` can be avoided by comparing `new block number` and `previously set block number` in `neuron-wallet`

Keith-CY:
Please have a review @homura @WhiteMinds @devchenyan @zhangyouxin

Keith-CY:
Please have a review @homura @WhiteMinds @devchenyan @zhangyouxin

packages/neuron-ui/src/types/Controller/index.d.ts
yanguoyu: I guess it's better to clear all data when starting a smaller block number, because some addresses may not be derived with a bigger block number in the previous.

packages/neuron-ui/src/components/PageContainer/hooks.ts
```diff
@@ -1,34 +1,65 @@
...
+      if (Number.isNaN(+blockNumber)) {
+        setCountdown(0)
+        return
+      }
+      setStartBlockNumber(+blockNumber > headerTipNumber ? headerTipNumber.toString() : blockNumber)
+      setBlockNumberErr(+blockNumber > headerTipNumber ? t('set-start-block-number.reset-to-header-tip-number') : '')
+      resetCountDown()
+    },
+    [resetCountDown, setCountdown, initStartBlockNumber, headerTipNumber, t]
+  )
```
homura, line 62: Mixing a UI hook with a non-UI hook may not be a good practice. Maybe we should split them into two hooks

```jsx
const { current, resetCountdown } = useCountdown(second)
const { blockNumber, setBlockNumber } = useSetBlockNumber

// reset countdown to ensure the user has enough time to confirm 
//  if the block number is right
useEffect(() => {
  resetCountdown();
}, [blockNumber])

<BlockNumberInput 
  onChange={(e) => {
    if(!isValidBlockNumber(e.target.value)) return;
    setBlockNumber(e.target.value)
  }}
/>
```

packages/neuron-ui/src/types/Controller/index.d.ts
Keith-CY: Sure, but my point is that the `clear all data` is not necessary in the API, it can be defined in the service, instead of exposing a field for other callers, because `if data should be removed` is determined by the service, not an external caller.

The logic should be `if the new start point is earlier than the previous one, then remove data` instead of `if the caller asks to remove data, then remove data`

homura (approved):

packages/neuron-ui/src/components/PageContainer/hooks.ts
```diff
@@ -1,34 +1,72 @@
...
+const useCountDown = (second: number) => {
+  const [countdown, setCountdown] = useState(0)
+  const countDecreaseTimeoutRef = useRef<ReturnType<typeof setTimeout>>()
+  const resetCountDown = useCallback(() => {
+    clearTimeout(countDecreaseTimeoutRef.current)
+    setCountdown(second)
+    const decrement = () => {
+      countDecreaseTimeoutRef.current = setTimeout(() => {
+        setCountdown(v => (v > 0 ? v - 1 : v))
+        decrement()
```
Keith-CY, line 19: Why is `decrement` called recursively

packages/neuron-ui/src/locales/en.json
```diff
@@ -1124,10 +1125,13 @@
     },
     "set-start-block-number": {
       "title": "Set the start block number",
+      "warn": "Warning: Transactions before synchronizing the starting block will not be synchronized",
```
Keith-CY, line 1128:
```suggestion
      "warn": "Warning: Transactions before the start block of synchronization will not be synchronized",
```

packages/neuron-ui/src/locales/en.json
```diff
@@ -1124,10 +1125,13 @@
...
       "tip": "The start sync block number is configuration when use light client ckb node, then you can set it and quick the sync",
       "input-place-holder": "Please enter the start sync block number, eg: 10,100,101",
       "locate-first-tx": "Locate the first transaction",
-      "view-block": "View the block"
+      "view-block": "View the block",
+      "set-less-than-before": "The block number you set is smaller than the previous set, the synchronization will restart, please proceed with caution.",
+      "reset-to-header-tip-number": "Unable to set a block number larger than the header tip block number, reset to the header tip block number."
```
Keith-CY, line 1134:
```suggestion
      "reset-to-header-tip-number": "Unable to set a block number greater than the header tip block number, reset to the header tip block number."
```

packages/neuron-ui/src/locales/en.json
Keith-CY, line 1: Update in French should be included since https://github.com/nervosnetwork/neuron/pull/3012

We may use ChatGPT for the internationalization.

packages/neuron-wallet/src/controllers/api.ts
```diff
@@ -332,15 +333,42 @@ export default class ApiController {
...
+          const lastSetStartBlockNumber = WalletsService.getInstance().getCurrent()?.toJSON().startBlockNumber
+          if (lastSetStartBlockNumber && +params.startBlockNumber < +lastSetStartBlockNumber) {
+            await CKBLightRunner.getInstance().clearNodeCache()
+          } else {
+            resetSyncTaskQueue.asyncPush(true)
+          }
+        } else {
+          throw new Error('Only Light client can set start block number')
+        }
+        return res
```
Keith-CY, line 356: The nested `if statement` can be avoided by `return early pattern`, e.g.

```ts
if (network.type !== NetworkType.Light) {
  throw new Error(...)
}

if (...) {
  clearCache...
} else {
  resetSyncTaskQueue...
}

return res
```

devchenyan (approved):

packages/neuron-ui/src/components/PageContainer/hooks.ts
yanguoyu: Because I use `setTimeout` to count down. Now I change it to `setInterval` and remove the recursively called.

packages/neuron-ui/src/components/PageContainer/hooks.ts
```diff
@@ -1,34 +1,71 @@
...
+      }, 1_000)
+    }
+    decrement()
+  }, [setCountdown, countDecreaseIntervalRef])
+  return {
+    countdown,
+    resetCountDown,
+  }
+}
+
```
Keith-CY, line 28: timer should be cleared when the countdown ends

Keith-CY (approved):

silySuper (approved):

PR #3000, opened 2023-12-22
fix: Hot/cold combination of a wallet doesn't work smooth

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/288

https://github.com/nervosnetwork/neuron/assets/11756869/49f5365e-5d90-42c3-bd03-93aed8388376


packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const onBack = useGoBack()
+
+  const navigate = useNavigate()
+  const onBroadcast = useCallback(async () => {
+    setIsBroadcasting(true)
+    try {
+      const res = await broadcastTransaction({
+        ...json,
+        walletID: wallet!.id,
```
Keith-CY, line 42: Unlike `broadcast in offline sign dialog`, the transaction to broadcast may not belong to the current wallet because it's been signed.

packages/neuron-ui/src/locales/en.json
```diff
@@ -112,6 +112,8 @@
         "partially-signed": "Partially Signed",
         "signed": "Signed"
       },
+      "import-signed-transaction": "Import signed transaction",
```
Keith-CY, line 115:
```suggestion
      "import-signed-transaction": "Import a signed transaction",
```

packages/neuron-ui/src/locales/en.json
```diff
@@ -121,7 +123,10 @@
         "broadcast": "Broadcast",
         "sign": "Sign and export",
         "cancel": "Cancel"
-      }
+      },
+      "broadcast-transaction": "Broadcast Transaction",
+      "import-unsigned-transaction": "Import unsigned transaction",
+      "import-unsigned-transaction-detail": "You have imported a unsigned transaction, please confirm and try again."
```
Keith-CY, line 129:
```suggestion
      "import-unsigned-transaction": "Import an unsigned transaction",
      "import-unsigned-transaction-detail": "You have imported an unsigned transaction, please confirm and try again."
```

packages/neuron-wallet/src/controllers/app/menu.ts
```diff
@@ -361,6 +362,18 @@ const updateApplicationMenu = (mainWindow: BrowserWindow | null) => {
           loadTransaction(URL.OfflineSign, json, filePath)
         },
       },
+      {
+        label: t('application-menu.tools.broadcast-transaction'),
+        enabled: hasCurrentWallet && !isXpubWallet,
```
Keith-CY, line 367: `Offline sign` should be disabled when the wallet is `xpub wallet` because it cannot sign a transaction, but `broadcast a transaction` doesn't rely on the ability to sign and can be enabled for all wallets.

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      })
+      if (isSuccessResponse(res)) {
+        navigate(RoutePath.History)
```
yanguoyu, line 45: https://github.com/nervosnetwork/neuron/pull/3000/files#r1435815632. Then I guess here we'd better not navigate to another page.

Keith-CY: Agree, as mentioned at https://github.com/nervosnetwork/neuron/pull/3000#discussion_r1435815632, the transaction to broadcast may not be related to any wallet in Neuron. So navigation to history could be misleading.

The best solution is as follows:
1. if the transaction has a cell of the current wallet, go to the history; Otherwise,
2. show a `View in Explorer` button to visit the transaction on CKB Explorer

Keith-CY:
Any update on the conversations above?

packages/neuron-wallet/src/controllers/offline-sign.ts
```diff
@@ -196,4 +196,13 @@ export default class OfflineSignController {
       }
     }
   }
+
+  public async broadcastTransactionOnly({ transaction }: OfflineSignJSON) {
```
yanguoyu, line 200: https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/offline-sign.ts#L158 has a type parameter. I don't know if ignoring `type` will have another side effect.

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+import styles from './broadcastTransaction.module.scss'
+
+const BroadcastTransaction = () => {
+  const {
+    app: { loadedTransaction = {} },
+    chain: { networkID },
+    settings: { networks },
+  } = useGlobalState()
+
+  const [wallet, setWallet] = useState<State.Wallet | null>(null)
```
Keith-CY, line 26: No wallet should be included if this component is only for broadcasting a transaction unless it should be redirected to the history when the broadcasted transaction includes addresses of the current wallet, but I didn't see this logic in this component.

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const [isBroadCasting, setIsBroadcasting] = useState(false)
```
Keith-CY, line 27: `isBroadCasting` should be `isBroadcasting`

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const [broadCastedTxHash, setBroadCastedTxHash] = useState('')
```
Keith-CY, line 28: `broadCastedTxHash` should be `broadcastedTxHash`

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    setIsBroadcasting(true)
+
+    try {
+      const res = await broadcastTransactionOnly({
+        ...json,
+      })
+      if (isSuccessResponse(res)) {
+        getLiveCells().then(cellRes => {
```
Keith-CY, line 60: I may have deleted a message from @yanguoyu mistakenly, it mentioned that the condition could be `if the broadcasted transaction hash in history of the current wallet`

packages/neuron-ui/src/locales/en.json
Keith-CY, line 1: Update in French should be included since https://github.com/nervosnetwork/neuron/pull/3012

We may use ChatGPT for the internationalization.

packages/neuron-ui/src/locales/en.json
```diff
@@ -120,8 +122,12 @@
       "actions": {
         "broadcast": "Broadcast",
         "sign": "Sign and export",
-        "cancel": "Cancel"
-      }
+        "cancel": "Cancel",
+        "view-in-explorer": "View in Explorer"
+      },
+      "broadcast-transaction": "Broadcast Transaction",
+      "import-unsigned-transaction": "Import an unsigned transaction",
```
Keith-CY, line 129: I would suggest informing users with `Import a transaction to sign` because it delivers the required action and the goal. cc @Danie0918 @Sven-TBD

packages/neuron-ui/src/services/remote/offline.ts
```diff
@@ -40,7 +40,8 @@ export type BroadcastProps = OfflineSignJSON & { walletID: string }
 
 export const exportTransactionAsJSON = remoteApi<OfflineSignJSON, void>('export-transaction-as-json')
 export const signTransactionOnly = remoteApi<OfflineSignJSON, void>('sign-transaction-only')
-export const broadcastTransaction = remoteApi<BroadcastProps, void>('broadcast-transaction-only')
+export const broadcastTransaction = remoteApi<BroadcastProps, void>('broadcast-transaction')
+export const broadcastTransactionOnly = remoteApi<OfflineSignJSON, string>('broadcast-transaction-only')
```
Keith-CY, line 44: How about typing it with `SignedTransaction` instead of `OfflineSignJSON` because `OfflineSignJSON` reminds me of `transaction to offline-sign`

packages/neuron-wallet/src/controllers/app/menu.ts
```diff
@@ -361,6 +362,18 @@ const updateApplicationMenu = (mainWindow: BrowserWindow | null) => {
           loadTransaction(URL.OfflineSign, json, filePath)
         },
       },
+      {
+        label: t('application-menu.tools.broadcast-transaction'),
+        enabled: hasCurrentWallet,
```
Keith-CY, line 367: `Broadcast transaction` doesn't reply on `which wallet is active` because it doesn't require an action of signing transactions

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
devchenyan: The wallet name needs to be displayed at the top.

image

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+const BroadcastTransaction = () => {
+  const {
+    app: { loadedTransaction = {} },
+    chain: { networkID },
+    settings: { networks },
+    wallet,
+  } = useGlobalState()
+
+  const [isBroadcasting, setIsBroadcasting] = useState(false)
+  const [broadcastedTxHash, setBroadCastedTxHash] = useState<string | null>('')
```
Keith-CY, line 22: `setBroadCastedTxHash` should be  `setBroadcastedTxHash`

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+      if (!wallet?.id) {
+        setBroadCastedTxHash(res.result)
+        return
+      }
+
+      if (res.result) {
+        getTransactionList({
+          walletID: wallet.id,
+          pageNo: 1,
+          pageSize: 1,
```
Keith-CY, line 65: Maybe the size could be a bit larger for fault tolerance. Or the `search` function in the history could be used to check if the `res.result` is in the history.

devchenyan: > Maybe the size could be a bit larger for fault tolerance. Or the `search` function in the history could be used to check if the `res.result` is in the history.

The logic of methods related to search in history is not suitable for the current scenario.

https://github.com/nervosnetwork/neuron/blob/4907867165926ada3240b390e27c808c1103324b/packages/neuron-ui/src/components/History/index.tsx#L48C21-L48C37

packages/neuron-wallet/src/controllers/offline-sign.ts
```diff
@@ -196,4 +196,47 @@ export default class OfflineSignController {
       }
     }
   }
+
+  public async broadcastTransactionOnly({
```
yanguoyu, line 200: Now, `broadcastTransactionOnly` is very similar to `broadcastTransaction`

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
Keith-CY: How about using `get transaction by wallet id and tx hash` at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/src/controllers/api.ts#L465

yanguoyu (approved):

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
devchenyan: https://github.com/nervosnetwork/neuron/blob/4907867165926ada3240b390e27c808c1103324b/packages/neuron-wallet/src/controllers/transactions.ts#L97C13-L97C13

As long as the wallet and transaction exist, transaction information will be returned regardless of whether the transaction is related to that wallet or not.

Keith-CY: My fault, but `get-transaction-list` should work, it fetches transactions belong to a wallet by a specific keyword, and the keyword could be the tx hash

devchenyan: > My fault, but `get-transaction-list` should work, it fetches transactions belong to a wallet by a specific keyword, and the keyword could be the tx hash

Yes, it is now using `get-transaction-list` and hash for queries.

Keith-CY: LOL, I must have to buy a new glasses.

Keith-CY (approved):

packages/neuron-ui/src/components/BroadcastTransaction/index.tsx
Danie0918: How about we remove the 'Wallet' field?

Danie0918: > How about we remove the 'Wallet' field?

cc @Keith-CY

Keith-CY: > > How about we remove the 'Wallet' field?
> 
> cc @Keith-CY

I will consider the `wallet` field useless in the dialog because there's no strong relationship between the transaction to broadcast and the active wallet. So, removing the wallet field looks good to me.

silySuper:
1.Why not calls unsigned transaction here?
截屏2024-01-29 15 25 18

devchenyan:
> 1.Why not calls unsigned transaction here? 截屏2024-01-29 15 25 18


This was suggested to be revised in the above comment.
https://github.com/nervosnetwork/neuron/pull/3000#discussion_r1449431287

silySuper (approved):

PR #3001, opened 2023-12-24
feat: Same wallet import optimization

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/291

https://github.com/nervosnetwork/neuron/assets/11756869/9a91e81a-ec62-40df-8d27-b51bac5a0065

https://github.com/nervosnetwork/neuron/assets/11756869/c7dbd15a-a964-4de3-8b80-b6dcfaad7d30


packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -374,18 +375,60 @@ export default class WalletService {
...
+      throw new ImportingExitingWallet(JSON.stringify({ extendedKey: props.extendedKey, id }))
+    }
+
     this.listStore.writeSync(this.walletsKey, [...this.getAll(), wallet.toJSON()])
 
     this.setCurrent(wallet.id)
     return wallet
   }
 
+  public replace = async (id: string, tmpId: string) => {
```
yanguoyu, line 397: `id` and `tmpId` should be more clear.

packages/neuron-ui/src/components/ReplaceDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { useNavigate } from 'react-router-dom'
+import { useTranslation } from 'react-i18next'
+import Dialog from 'widgets/Dialog'
+import RadioGroup from 'widgets/RadioGroup'
+import { useState as useGlobalState, useDispatch, updateWalletProperty, setCurrentWallet } from 'states'
+import { RoutePath, isSuccessResponse } from 'utils'
+import { replaceWallet } from 'services/remote'
+import styles from './replaceDuplicateWalletDialog.module.scss'
+
+const useReplaceDuplicateWallet = () => {
```
yanguoyu, line 11: `useReplaceDuplicateWallet` this function is only used in `packages/neuron-ui/src/components/WalletWizard/index.tsx`, why not move to there.

packages/neuron-ui/src/components/DetectDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (deletableWallets.includes(currentID)) {
+      ids = ids.filter(item => item !== currentID)
+      ids.push(currentID)
+    }
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const id of ids) {
+      // eslint-disable-next-line no-await-in-loop
+      await getRequest(id)
+    }
```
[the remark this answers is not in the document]
devchenyan: yes,  these requests need to be requested one by one. Deleting each wallet requires entering its corresponding password.

packages/neuron-ui/src/components/ReplaceDuplicateWalletDialog/index.tsx
devchenyan: `importedWallet` is used in three files

devchenyan: > useReplaceDuplicateWallet

My mistake, it is `useReplaceDuplicateWallet` that is used in three files, not `importedWallet`.

yanguoyu (approved):

Danie0918:
@WhiteMinds @homura Please have a review.

packages/neuron-wallet/src/exceptions/wallet.ts
```diff
@@ -78,6 +78,13 @@ export class WalletFunctionNotSupported extends Error {
   }
 }
 
+export class ImportingExitingWallet extends Error {
```
homura, line 81: The `Exiting` may be `Existing`. I suggest the name `DuplicateImportWallet` for easy understanding

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -374,18 +375,55 @@ export default class WalletService {
...
+    const wallet = this.get(existingWalletId)
+    if (!wallet || !this.importedWallet) {
+      throw new WalletNotFound(existingWalletId)
+    }
+
+    const newWallet = this.importedWallet?.toJSON()
+    if (importedWalletId !== newWallet.id) {
+      throw new WalletNotFound(importedWalletId)
+    }
+    if (wallet.toJSON().extendedKey !== newWallet.extendedKey) {
```
homura, line 407: It is recommended to use `bytes.equals(wallet.extendedKey, newWallet.extendedKey)` to avoid the case issue, like `"0xFF" !== "0xff"`

packages/neuron-ui/src/components/DetectDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { requestPassword } from 'services/remote'
+import styles from './detectDuplicateWalletDialog.module.scss'
+
+const DetectDuplicateWalletDialog = ({ onClose }: { onClose: () => void }) => {
+  const {
+    wallet: { id: currentID = '' },
+    settings: { wallets = [] },
+  } = useGlobalState()
+  const dispatch = useDispatch()
+  const [deletableWallets, setDeletableWallets] = useState<string[]>([])
```
homura, line 15: `duplicatedWallets` may be more precise

packages/neuron-ui/src/components/DetectDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+            actionType: 'delete',
+            onSuccess: async () => {
+              await resolve(id)
+            },
+          },
+        })
+      })
+    }
+
+    let ids = [...deletableWallets]
```
homura, line 71: any reason for the shallow copy here? maybe the `Set` is helpful here to remove the duplicated ids

```ts
const ids = new Set(deletableWallets)
```

devchenyan: There is no need to remove the duplicated ids here. Just need to move the currentID to the last position of ids.

packages/neuron-wallet/src/services/wallets.ts
devchenyan: These two values should be exactly the same, without ignoring case differences.

packages/neuron-ui/src/components/DetectDuplicateWalletDialog/index.tsx
homura: I see. But this piece is a bit confusing for me, it would be more understandable if we refactor it like this

```ts
const requestToDeleteIds = duplicatedWallets
  .filter(item => item !== currentID)
  .concat(duplicatedWallets.include(currentID) ? [currentID] : [])
```

packages/neuron-ui/src/components/DetectDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+        obj[item.extendedKey] = [item]
+      }
+    })
+
+    return Object.values(obj).filter(list => list.length > 1)
+  }, [wallets])
+
+  const handleGroupChange = useCallback(
+    (checked: string) => {
+      const [extendedKey, id] = (checked as string).split('_')
```
homura, line 35:
```suggestion
      const [extendedKey, id] = checked.split('_')
```

the `as string` definition here is redundant since `checked: string` is already defined in the arguments

packages/neuron-ui/src/components/DetectDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+      const list: string[] = []
+      wallets.forEach(item => {
+        if (item.extendedKey === extendedKey) {
+          if (item.id !== id) {
+            list.push(item.id)
+          }
+        } else if (duplicatedWallets.includes(item.id)) {
+          list.push(item.id)
+        }
+      })
```
homura, line 45:
```suggestion
      wallets.filter(item => {
        if (item.extendedKey === extendedKey) {
          return item.id !== id
        } else {
          return duplicatedWallets.includes(item.id)
        }
      }).map(({ id }) => id);
```

packages/neuron-ui/src/components/ReplaceDuplicateWalletDialog/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const group = useMemo(() => wallets.filter(item => item.extendedKey === extendedKey), [wallets, extendedKey])
+
+  const handleGroupChange = useCallback(
+    (checked: string) => {
+      setSelectedId(checked as string)
+    },
+    [setSelectedId]
+  )
+
+  const onConfirm = useCallback(async () => {
```
homura, line 80:
```suggestion
  const onConfirm = useCallback(async () => {
```

the `async` here is unnecessary because the method body has no `await`

packages/neuron-wallet/tests/services/wallets.test.ts
```diff
@@ -155,6 +156,29 @@ describe('wallet service', () => {
         addressType: AddressType.Receiving,
       },
     }
+
+    wallet5 = {
+      name: 'wallet-test5',
+      id: '',
+      extendedKey: 'a',
```
homura, line 163:
```suggestion
      extendedKey: wallet1.extendedKey,
```

to explicitly state that this is a duplicated wallet

homura (approved):

Keith-CY:
Conflicted

silySuper (approved):

PR #3002, opened 2023-12-24
fix: Optimize the error of update Neuron

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/336

image


Keith-CY (approved):

yanguoyu (approved):

Keith-CY:
https://github.com/nervosnetwork/neuron/assets/7271329/9b9cf750-ac1b-4ed3-b684-b4c6af069329

The dialog shows repeatedly when it's dismissed by `esc` button

BTW, the `confirm` button should be white text color when it's in dark mode

devchenyan:
> Screen.Recording.2023-12-26.at.18.58.19.mov 
> The dialog shows repeatedly when it's dismissed by `esc` button
> 
> BTW, the `confirm` button should be white text color when it's in dark mode

In the design criteria, the cancel button should be white and the confirm button should be black text color when it's in dark mode.
image

silySuper (approved):

PR #3004, opened 2023-12-29
Enable playwright

Keith-CY:


Keith-CY:
I found the `request for review` has been proposed 2 weeks ago, but the CI failed, please check @silySuper

PR #3010, opened 2024-01-04
fix: Remove walletId for `get-multisig-config` API

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/348
1. Remove `walletId` for `get-multisig-config` API
2. Disabled `Approve` button when the current wallet can not sign for the multisig config.

https://github.com/nervosnetwork/neuron/assets/12881040/32906371-9c4e-4421-951f-2edfe9fd372f

### DIFF
--- a/packages/neuron-ui/src/components/MultisigAddress/index.tsx
+++ b/packages/neuron-ui/src/components/MultisigAddress/index.tsx
@@ -28,10 +28,12 @@ import { ReactComponent as Edit } from 'widgets/Icons/Edit.svg'
 import { Download, Search } from 'widgets/Icons/icon'
 import { HIDE_BALANCE, NetworkType } from 'utils/const'
 import { onEnter } from 'utils/inputDevice'
+import getMultisigSignStatus from 'utils/getMultisigSignStatus'
 import { useSearch, useConfigManage, useExportConfig, useActions, useSubscription } from './hooks'
 
 import styles from './multisigAddress.module.scss'
 
+const ApproveKey = 'approve'
 const tableActions = [
   {
     key: 'info',
@@ -46,7 +48,7 @@ const tableActions = [
     icon: <Transfer />,
   },
   {
-    key: 'approve',
+    key: ApproveKey,
     icon: <Confirm />,
   },
 ]
@@ -56,7 +58,7 @@ const MultisigAddress = () => {
   useOnLocaleChange(i18n)
   useExitOnWalletChange()
   const {
-    wallet: { id: walletId },
+    wallet: { id: walletId, addresses },
     chain: { networkID },
     settings: { networks = [] },
   } = useGlobalState()
@@ -281,6 +283,7 @@ const MultisigAddress = () => {
                   dataIndex: 'action',
                   align: 'left',
                   render(_, __, item) {
+                    const { canSign } = getMultisigSignStatus({ multisigConfig: item, addresses })
                     return (
                       <div className={styles.action}>
                         <Tooltip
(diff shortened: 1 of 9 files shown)


devchenyan (approved):

packages/neuron-wallet/src/services/sync-progress.ts
```diff
@@ -27,13 +27,13 @@ export default class SyncProgressService {
       .createQueryBuilder()
       .update(SyncProgress)
       .set({ delete: true })
-      .where({ walletId: Not(In(existWalletIds)) })
+      .where({ walletId: Not(In(existWalletIds)), addressType: SyncAddressType.Default })
```
Keith-CY, line 30: What is `addressType` for in this query

packages/neuron-ui/src/components/MultisigAddress/index.tsx
```diff
@@ -296,7 +299,7 @@ const MultisigAddress = () => {
                                   key={key}
                                   data-key={key}
                                   onClick={onClickItem(item)}
-                                  disabled={disabled}
+                                  disabled={key === ApproveKey ? !canSign || disabled : disabled}
```
Keith-CY, line 302: Availability should be set in the action list at https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-ui/src/components/MultisigAddress/index.tsx#L120

```ts
        disabled: selectIds.length > 1 && item.key !== 'delete',
```

yanguoyu: It depends on the data source, so it can not set value there.

Keith-CY: > It depends on the data source, so it can not set value there.

The available address list can be a condition of `listActionOptions`, or to say, `listActionOptions` changes when the address list changes

packages/neuron-wallet/src/services/sync-progress.ts
yanguoyu: In the previous, the multisig-config relates to the wallet. So we can remove other wallets' light client sync except for the current wallet. But now the multisig-config is a global config. So we should filter the wallet and ignore multisig configs. The `addressType` is a query for default `lockArgs` to filter multisig `lockArgs`.

packages/neuron-ui/src/components/MultisigAddress/index.tsx
yanguoyu: `listActionOptions` is a global option for every item. But every item's `Approve` disable status may be different.

Keith-CY: got it

packages/neuron-wallet/src/services/sync-progress.ts
Keith-CY: Got it

Keith-CY (approved):

silySuper:
After switching  different wallets for several times,sync process is stoped,even I have exited and open app again.

https://github.com/nervosnetwork/neuron/assets/25971273/a8d91ece-8fa1-43a6-992a-d6588cfcc61c

[Uploading neuron_debug_1705371474620.zip…]()

yanguoyu:
> After switching different wallets for several times,sync process is stoped,even I have exited and open app again.
> 
>  2024-01-16.10.21.14.mov 
> Uploading neuron_debug_1705371474620.zip…

Does the multisig sync to 99.95% block height? If not, it's normal because the light client should wait for the least sync block number.

silySuper:
This is multisig sync,already to  99.95% block height
截屏2024-01-16 10 28 21

yanguoyu:
> This is multisig sync,already to 99.95% block height 截屏2024-01-16 10 28 21

Thanks, I will find out what happened, besides, the debug information does not upload success in the up comment.

silySuper:
截屏2024-01-16 10 32 34
debug log shows this ,does this affect?

[neuron_debug_1705372285343.zip](https://github.com/nervosnetwork/neuron/files/13944690/neuron_debug_1705372285343.zip)

silySuper (approved):

PR #3014, opened 2024-01-10
Add left time for sync progress

yanguoyu:
1. Add left time for sync progress
image
image

2. Move the global dialog to the button when clicking `Refresh`

image


Keith-CY (approved):

yanguoyu:
@devchenyan @WhiteMinds @homura  Please have a review.

devchenyan (approved):

silySuper:
1.In light client mode,left time shown after sync 0.2%,in expect after sync is begin it should show

https://github.com/nervosnetwork/neuron/assets/25971273/a4604df7-e6db-4513-94f7-ee9ade03adfc


2.In light client mode,left time is increasing when sync process is increasing actually,in expect left time and sync process data should  inverse ratio.

https://github.com/nervosnetwork/neuron/assets/25971273/0d5582c2-339f-47d8-aacf-b12a40d441e5

3.By the way ,another UI problem:
when window is minimum，language after English(United States)  can not show and chose.Maybe add scroll bar is better.

https://github.com/nervosnetwork/neuron/assets/25971273/1f897fc4-202c-4b9b-8920-349905bb28e7

yanguoyu:
> 3.By the way ,another UI problem:
when window is minimum，language after English(United States) can not show and chose.Maybe add scroll bar is better.

I believe it will be better to open an new issue to track this problem.

yanguoyu:
Because the sync speed depends on the latest tip block number, network speed, and synced block number so it will keep correcting during synchronization.

silySuper:
Ok,Maybe No.1 and No.2 will cause users confused .@Danie0918 should know  this actual effect.
No.3 bug  new link: https://github.com/Magickbase/neuron-public-issues/issues/357

Danie0918:
> Ok,Maybe No.1 and No.2 will cause users confused .@Danie0918 should know this actual effect. No.3 bug new link: [Magickbase/neuron-public-issues#357](https://github.com/Magickbase/neuron-public-issues/issues/357)

This is the estimated time, and it's understandable that the remaining time changes, and I'll consider adding instructions subsequently.

silySuper (approved):

PR #3025, opened 2024-01-13
fix: The sync child process broken and not restart

devchenyan:
fix issue: https://github.com/Magickbase/neuron-public-issues/issues/309

WX20240113-235625

After the bug is fixed, errors are captured to ensure continuous synchronization.
image
image


yanguoyu:
After catching the rejecting error, will the sync process continue running?

devchenyan:
> After catching the rejecting error, will the sync process continue running?

When the error is not captured, the synchronization is interrupted
image


After capturing the error, the synchronization process does not get interrupted and continues to synchronize.
image

Keith-CY (approved):

yanguoyu (approved):

Keith-CY (approved):

silySuper (approved):

PR #3052, opened 2024-03-02
fix: When address is not enough, stop update localSavedBlockNumber

yanguoyu:
This is an issue from [`discord`](https://discord.com/channels/956765352514183188/1084671124211769364/1212615721897689088)

The reason is the generated address and sync process are not synchronized. So when there is not enough address, the sync process will stop saving `localSavedBlockNumber`. Then after the new address is derived, it will start syncing from the right block number.

### DIFF
--- a/packages/neuron-wallet/src/services/addresses.ts
+++ b/packages/neuron-wallet/src/services/addresses.ts
@@ -95,19 +95,16 @@ export default class AddressService {
     receivingAddressCount: number = DefaultAddressNumber.Receiving,
     changeAddressCount: number = DefaultAddressNumber.Change
   ): Promise<AddressInterface[] | undefined> {
-    const [unusedReceivingAddresses, unusedChangeAddresses] = await this.getGroupedUnusedAddressesByWalletId(walletId)
-    const unusedReceivingCount = unusedReceivingAddresses.length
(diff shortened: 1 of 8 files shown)


yanguoyu:
The unsigned package can be found at https://github.com/yanguoyu/neuron/actions/runs/8123096389

Keith-CY (approved):

Keith-CY:
@devchenyan @homura @silySuper please have a review

devchenyan (approved):

packages/neuron-wallet/src/services/addresses.ts
```diff
@@ -140,6 +137,20 @@ export default class AddressService {
     return allGeneratedAddresses
   }
 
+  public static async getNeedGenerateAddressCount(
+    walletId: string,
+    receivingAddressCount: number = DefaultAddressNumber.Receiving,
+    changeAddressCount: number = DefaultAddressNumber.Change
+  ) {
```
homura, line 144: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit

I'd suggest renaming it to `getAddressCountsToFillGapLimit` for better clarity

homura (approved):

silySuper (approved):

PR #3055, opened 2024-03-04
fix: Fix sending sudt to a new acp cell with extra 142 CKB by offline sign

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/361
When testing this issue, it's better to send all of your ckb balance to one address, if your wallet has some ckb cells, maybe the last cell skip sign will be ok.


Danie0918:
@devchenyan @homura Please have a review.

devchenyan (approved):

Keith-CY (approved):

silySuper (approved):

PR #3069, opened 2024-03-13
Fix cell manage with hard wallet

yanguoyu:
1. Support lock or unlock when using a hard wallet

https://github.com/nervosnetwork/neuron/assets/12881040/08552ecd-c134-4b9f-b774-34f6d040dec4

2. Show the tx hash in the cell info dialog

image


Danie0918:
@devchenyan @homura @Keith-CY Please have a review.

homura:
IMO, it is best to minimize the use of passwords or hardware to reduce the risk of exposing private keys

I am not sure whether it is essential to logically unlock client-side cells through a hardware device since these cells are not physically locked but rather prevented from being spent by the wallet

yanguoyu:
> IMO, it is best to minimize the use of passwords or hardware to reduce the risk of exposing private keys
> 
> I am not sure whether it is essential to logically unlock client-side cells through a hardware device since these cells are not physically locked but rather prevented from being spent by the wallet

https://github.com/Magickbase/neuron-public-issues/issues/218#issuecomment-1696950793 This the suggestion from the PM

packages/neuron-ui/src/components/CellManagement/hooks.ts
```diff
@@ -362,3 +375,117 @@ export const usePassword = () => {
...
+
+export const useHardWallet = ({ wallet, t }: { wallet: State.WalletIdentity; t: TFunction }) => {
+  const isWin32 = useMemo(() => {
+    return getPlatform() === 'win32'
+  }, [])
+  const [error, setError] = useState<string | undefined>()
+  const disconnectStatus = t('hardware-verify-address.status.disconnect')
+  const ckbAppNotFoundStatus = t(CkbAppNotFoundException.message)
+  const isNotAvailable = useMemo(() => {
+    return error === disconnectStatus || error === ckbAppNotFoundStatus
```
Keith-CY, line 387: Why not compare error with `error code` to avoid effect from `i18n json`

silySuper (approved):

devchenyan (approved):

PR #3101, opened 2024-04-03
fix: Migrate dialog does not show in v0.114.2

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/374


Danie0918:
@Keith-CY @homura  Please have a review.

packages/neuron-wallet/src/services/ckb-runner.ts
```diff
@@ -124,7 +124,10 @@ export const startCkbNode = async () => {
   currentProcess.stderr?.on('data', data => {
     const dataString: string = data.toString()
     logger.error('CKB:\trun fail:', dataString)
-    if (dataString.includes('CKB wants to migrate the data into new format')) {
+    if (
+      dataString.includes('CKB wants to migrate the data into new format') ||
+      dataString.includes('CKB recommends migrating your data into a new format')
```
Keith-CY, line 129: It's better to define them as `ENUM` to make them clear

Keith-CY (approved):

_typos.toml
```diff
@@ -1,6 +1,8 @@
 [default.extend-words]
 thur = "thur"
 numer = "numer"
+HD = "HD"
```
Keith-CY, line 4: Which word is this for

silySuper:
I have start 0.113.0 full node on local,and open test package above,but can not see migrate tip.
image

What should I do to have migrate effect?

Keith-CY:
> I have start 0.113.0 full node on local,and open test package above,but can not see migrate tip. image
> 
> What should I do to have migrate effect?

Replacing the chain data(default location `/Users/{User name}/Library/Application Support/Neuron/chains/testnet/data`) with one synced by CKB@v0.113.0 and CKB@v0.110.2 should trigger the migration.

silySuper:
This is my files in chains path,no testnet files.My testnet files is in node files
截屏2024-04-07 17 10 15
截屏2024-04-07 17 11 22

Keith-CY:
> This is my files in chains path,no testnet files.My testnet files is in node files 截屏2024-04-07 17 10 15 截屏2024-04-07 17 11 22

The path I mentioned above is the default one. It can be set in `Settings => Data => CKB Node &Config`, so you can replace the `testnet/data` in your custom node data directory to trigger the migration.

silySuper:
截屏2024-04-07 17 44 09
截屏2024-04-07 17 44 17
 Only internal node has this path settings,and the file name is mainnet ,not testnet.So you mean I can change this internal node path to my 0.113 testnet data ,that I can trigger the migration?

Keith-CY:
> 截屏2024-04-07 17 44 09 截屏2024-04-07 17 44 17 Only internal node has this path settings,and the file name is mainnet ,not testnet.So you mean I can change this internal node path to my 0.113 testnet data ,that I can trigger the migration?

IMO, the user-specified full node(3rd option in this case) should be able to set the data directory, it may be a bug, need confirmation from @Danie0918 

You can replace the `data` in your own node directory to trigger the migration if you're using the self-hosted node.

Danie0918:
> > 截屏2024-04-07 17 44 09 截屏2024-04-07 17 44 17 Only internal node has this path settings,and the file name is mainnet ,not testnet.So you mean I can change this internal node path to my 0.113 testnet data ,that I can trigger the migration?
> 
> IMO, the user-specified full node(3rd option in this case) should be able to set the data directory, it may be a bug, need confirmation from @Danie0918
> 

This should be a bug that needs to be fixed.

silySuper:
> > > 截屏2024-04-07 17 44 09 截屏2024-04-07 17 44 17 Only internal node has th,is path settings,and the file name is mainnet ,not testnet.So you mean I can change this internal node path to my 0.113 testnet data ,that I can trigger the migration?
> > 
> > 
> > IMO, the user-specified full node(3rd option in this case) should be able to set the data directory, it may be a bug, need confirmation from @Danie0918
> 
> This should be a bug that needs to be fixed.

OK, @devchenyan

yanguoyu:
> > > 截屏2024-04-07 17 44 09 截屏2024-04-07 17 44 17 Only internal node has this path settings,and the file name is mainnet ,not testnet.So you mean I can change this internal node path to my 0.113 testnet data ,that I can trigger the migration?
> > 
> > 
> > IMO, the user-specified full node(3rd option in this case) should be able to set the data directory, it may be a bug, need confirmation from @Danie0918
> 
> This should be a bug that needs to be fixed.

I don't think so. If the selected node is not an internal node, it can not be set the data path. At that the users start the ckb node, so we will not show the migrate dialog and show the data path. And we can not get the ckb data path, all of the ckb node and data are controlled by the users.

yanguoyu (approved):

silySuper:
What is the conclusion?Need to add path setting in testnet node? @Danie0918

yanguoyu:
> What is the conclusion?Need to add path setting in testnet node? @Danie0918

Migrate dialog only shows when selecting the `Internal node`, so you should test by selecting the `Internal Node`.

silySuper:
I have replace data to  /Volumes/My\ Passport/ckb节点/ckb_v0.108.1_x86_64-apple-darwin-portable/testnet/data ，but can not see migrate dialog.


https://github.com/nervosnetwork/neuron/assets/25971273/628d9146-7756-4749-923e-e0261e97fa1f

silySuper:
@devchenyan please check this issue~

devchenyan:
> @devchenyan please check this issue~

1. Completely uninstall neuron 
2. Use a low version of neuron to synchronise data
3. Use the test package to synchronise data, trigger the migration.

silySuper:
> > @devchenyan please check this issue~
> 
> @silySuper
> 
> 1. Completely uninstall neuron
> 2. Use a low version of neuron to synchronise data
> 3. Use the test package to synchronise data, trigger the migration.
actually I think neuron v114.2 is the test package because of the title,does old testnet data create by old version(v108) can not  trigger the migration, before which lower version can  use?

packages/neuron-wallet/src/services/ckb-runner.ts
```diff
@@ -26,6 +26,11 @@ const platform = (): string => {
   }
 }
 
+enum NeedMigrateMsg {
+  Wants = 'CKB wants to migrate the data into new format',
+  Recommends = 'CKB recommends migrating your data into a new format',
```
homura, line 31:
```suggestion
  // https://github.com/nervosnetwork/ckb/blob/31bcecddc8ad5b7d0a30399ef55a5ce0a359bb7e/shared/src/shared_builder.rs#L79
  Recommends = 'CKB recommends migrating your data into a new format',
```

silySuper:
> > > @devchenyan please check this issue~
> > 
> > 
> > @silySuper
> > 
> > 1. Completely uninstall neuron
> > 2. Use a low version of neuron to synchronise data
> > 3. Use the test package to synchronise data, trigger the migration.
> >    actually I think neuron v114.2 is the test package because of the title,does old testnet data create by old version(v108) can not  trigger the migration, before which lower version can  use? @devchenyan

截屏2024-04-18 11 46 30

When last version is not null and !v108  both true,can trigger the migration ?

silySuper (approved):

silySuper:
Maybe  internal node data path choose can add tips to tell users which directory can be used is better @Danie0918

Danie0918:
> Maybe internal node data path choose can add tips to tell users which directory can be used is better @Danie0918

This can only be suggested to the user and may still cause this problem. Any other solutions? @devchenyan @yanguoyu

yanguoyu:
> > Maybe internal node data path choose can add tips to tell users which directory can be used is better @Danie0918
> 
> This can only be suggested to the user and may still cause this problem. Any other solutions? @devchenyan @yanguoyu

I believe the user only needs to know the root directory of the ckb data, it's unnecessary to understand the meaning of each directory of the ckb data root directory.

Issue #3108, opened 2024-04-08
[Dev] Module not found: Error: You attempted to import .../neuron/node_modules/react-refresh/runtime.js which falls outside of the project src/ directory


twhy:
## Description
Running Neuron Wallet in dev mode with `yarn start` results in `Module not found Error`

## Environment
```
OS: Mac OS 14.4.1
Yarn: 1.22.21
Node: v20.11.1
```

## Reproduce Steps
1. Git clone repo
2. Follow README.md to install all dependencies
3. Run `yarn start`

## Terminal logs
```shell
yarn start
yarn run v1.22.21
$ concurrently "cross-env BROWSER=none yarn run start:ui" "wait-on -t 10000 http://127.0.0.1:3000 && yarn run start:wallet"
$ cd packages/neuron-ui && yarn run start
$ cross-env DISABLE_ESLINT_PLUGIN=true GENERATE_SOURCEMAP=false react-app-rewired start
[0] Browserslist: caniuse-lite is outdated. Please run:
[0]   npx update-browserslist-db@latest
[0]   Why you should do it regularly: https://github.com/browserslist/update-db#readme
[0] (node:99432) [DEP_WEBPACK_DEV_SERVER_ON_AFTER_SETUP_MIDDLEWARE] DeprecationWarning: 'onAfterSetupMiddleware' option is deprecated. Please use the 'setupMiddlewares' option.
[0] (Use `node --trace-deprecation ...` to show where the warning was created)
[0] (node:99432) [DEP_WEBPACK_DEV_SERVER_ON_BEFORE_SETUP_MIDDLEWARE] DeprecationWarning: 'onBeforeSetupMiddleware' option is deprecated. Please use the 'setupMiddlewares' option.
[0] Starting the development server...
[0] 
[0] [BABEL] Note: The code generator has deoptimised the styling of /Users/twhy/Code/nervos/neuron/packages/neuron-ui/src/widgets/Icons/VerificationFailure.svg as it exceeds the max of 500KB.
[0] [BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.
[0] Failed to compile.
[0] 
[0] Module not found: Error: You attempted to import /Users/twhy/Code/nervos/neuron/node_modules/react-refresh/runtime.js which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
[0] You can either move it inside src/, or add a symlink to it from project's node_modules/.
[...]

PR #3109, opened 2024-04-08
fix: Fix module not found error in dev mode (#3108)

twhy:
Fix https://github.com/nervosnetwork/neuron/issues/3108

### DIFF
--- a/packages/neuron-ui/package.json
+++ b/packages/neuron-ui/package.json
@@ -100,6 +100,7 @@
     "jest-styled-components": "7.2.0",
     "react-app-rewired": "2.2.1",
     "react-scripts": "5.0.1",
+    "react-refresh": "0.11.0",
     "react-test-renderer": "18.2.0",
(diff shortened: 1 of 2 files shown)


Keith-CY:
It's caused by conflicts between `react-refresh@v0.11.0` and `react-refresh@v0.14.0` and can be fixed by specifying the version in `package.json > resolutions` at https://github.com/nervosnetwork/neuron/blob/develop/package.json#L67

I would suggest using `react-refresh@v0.14.0` because `react-scripts` accepts `react-refresh@^v0.11.0`, which includes `react-refresh@v0.14.0`

```diff
- "react-i18next": ">=11.16.4",
+ "react-i18next": ">=11.16.4",
+ "react-refresh": "0.14.0"
```

twhy:
@Keith-CY  Your suggestion is much better, I'll close this PR and create a new one.

PR #3111, opened 2024-04-08
refactor: replace shannonToCKB with lumos formatUnit

twhy:
How this PR was implemented

1.  Rewrite `shannonToCKB` with `@ckb-lumos/bi/formatUnit` as below.
```typescript
import { formatUnit, ckbDecimals } from '@ckb-lumos/bi'

const shannonToCKB = (shannon: bigint) => {
  return new Intl.NumberFormat('en-US', {
    useGrouping: false,
    signDisplay: 'always',
    minimumFractionDigits: ckbDecimals,
    maximumFractionDigits: ckbDecimals,
  }).format(Number(formatUnit(shannon, 'ckb')))
}

export default shannonToCKB
```
2. Run `shannonToCKB.test.ts` to make sure the new implementation pass all tests.
image

3. Go to `to-csv-row.ts`(the only file using `shannonToCKB`), rewrite
```ts
amount = shannonToCKB(BigInt(tx.value ?? ''))
```
 as 
```ts
amount = new Intl.NumberFormat('en-US', {
  useGrouping: false,
  signDisplay: 'always',
  minimumFractionDigits: ckbDecimals,
  maximumFractionDigits: ckbDecimals,
}).format(Number(formatUnit(BigInt(tx.value ?? '0'), 'ckb'))) 
```
`BigInt('')` and `BigInt('0')` are both evaluated to `0n`, but `BigInt('0')` is more explicit.

4. Delete `shannonToCKB.ts`  and `shannonToCKB.test.ts`


homura:
Looks good. Since version 0.22, Lumos has provided an entry package `@ckb-lumos/lumos` that includes almost all common functions. It's suggested to import only `@ckb-lumos/lumos` for more convenience

```ts
import { formatUnit } from '@ckb-lumos/lumos/utils'
formatUnit(...)

// or
import { utils } from '@ckb-lumos/lumos'
utils.formatUnit(...)
```

twhy:
@homura  That's very convinient! 
But the neuron-wallet is still using `0.21.1`
https://github.com/nervosnetwork/neuron/blob/develop/packages/neuron-wallet/package.json#L45
So we probably should create a separate PR to upgrade the lumos dependencies and update the imports after this one get merged.

homura (approved):

packages/neuron-wallet/src/utils/to-csv-row.ts
```diff
@@ -41,7 +41,12 @@ const toCSVRow = (
       txType = +tx.sudtInfo.amount <= 0 ? `UDT ${SEND_TYPE}` : `UDT ${RECEIVE_TYPE}`
     }
   } else {
-    amount = shannonToCKB(BigInt(tx.value ?? ''))
+    amount = new Intl.NumberFormat('en-US', {
+      useGrouping: false,
+      signDisplay: 'always',
+      minimumFractionDigits: ckbDecimals,
+      maximumFractionDigits: ckbDecimals,
+    }).format(Number(formatUnit(BigInt(tx.value ?? '0'), 'ckb')))
```
Keith-CY, line 49: `Number` has precision issue, e.g.

```
new Intl.NumberFormat('en-US', {
      useGrouping: false,
      signDisplay: 'always',
      minimumFractionDigits: 8,
      maximumFractionDigits: 8,
    }).format(Number('123456789012345678901234567890'))
// +123456789012345680000000000000.00000000
```

`formatUnit` returns `string` which is enough for this case.

Keith-CY: It's better to enclose the `type cast` in a function and guard it by test cases.

I just checked the function of `foramtUnit`(https://github.com/ckb-js/lumos/blob/bb1fd0a233ac69ce1976e6b9036ea7fa17476457/packages/bi/src/index.ts#L197) and found that this method transformed `shannon` to `CKB` well. Why is `Intl.NumberFormat` used here without `usingGrouping`?

If `Intl.NumberFormat` is just for the positive sign, it can be fixed easily.

twhy: `Intl.NumberFormat` is used for two reasons
1. Add sign `+` when necessary
2. Make sure `amount` has 8 fraction digits. 
e.g. `formatUnit(BigInt(0), 'ckb')` returns '0.0' instead of '0.00000000' 
and `formatUnit(BigInt(12345678900000000), 'ckb')` returns `"123456789.0"` instead of `"+123456789.00000000"`

The `useGrouping: false` option makes sure `amount` has no commas.

Here are the `shannonToCKB.test.ts` test cases
```ts
const fixtures: { shannon: bigint; expected: string }[] = [
  {
    shannon: BigInt(0),
    expected: '+0.00000000',     // Received: "0.0" without Intl.NumberFormat
  },
  {
    shannon: BigInt(1234567),
    expected: '+0.01234567',
  },
  {
    shannon: BigInt(-1234567),
    expected: '-0.01234567',
  },
  {
    shannon: BigInt(12345678),
    expected: '+0.12345678',
  },
  {
    shannon: BigInt(-12345678),
    expected: '-0.12345678',
  },
  {
    shannon: BigInt(123456789),
    expected: '+1.23456789',
  },
  {
    shannon: BigInt(-123456789),
    expected: '-1.23456789',
  },
  {
    shannon: BigInt(12345678900000000),
    expected: '+123456789.00000000',  // Received: "123456789.0" without Intl.NumberFormat
  },
  {
    shannon: BigInt(-12345678900000000),
    expected: '-123456789.00000000',  // Received: "-123456789.0" without Intl.NumberFormat
  },
]
```

Keith-CY: Got it, then I would suggest adding a method to enclose the `type cast`

twhy:
```ts
```

homura: Maybe it is good enough to rewrite the `NumberFormat` declaration for this scenario

```ts
declare global {
  namespace Intl {
    interface NumberFormat {
      format(value: number | string): string;
    }
  }
}
```

twhy: Where to put this declaration ? When I try to put it inside `to-csv-row.ts`
image


I actually think `as any` or `as unknown as number` is not that bad because at langage level the `Intl.NumberFormat.prototype.format()` method does accept string. 

It's the typings are not well written, or too strict.

image

ninja-fe: > Where to put this declaration ? When I try to put it inside to-csv-row.ts

image

homura: > Where to put this declaration ? When I try to put it inside to-csv-row.ts

image

twhy: OK. Cool!  @Keith-CY What do you think about this appoach ?

homura: > All dirty works are concealed in the formatAmount method, and it can be guaranteed by unit tests.

Doing it in this way does indeed make it easier to read, without relying on TypeScript's tricks

twhy: > > Where to put this declaration ? When I try to put it inside `to-csv-row.ts` image
> > I actually think `as any` or `as unknown as number` is not that bad IN THIS CASE because at langage level the `Intl.NumberFormat.prototype.format()` method does accept string.
> > It's the typings are not well written, or too strict.
> > image
> > @Keith-CY Can you show a little example about how to enclose the `type cast` by adding a method in this case ? I'm a bit confused.
> 
> Simply add a method as
> 
> ```ts
> const formatAmount = (amount: string | number | ...) => 
>     new Intl.NumberFormat('en-US', {
>       useGrouping: false,
>       signDisplay: 'always',
>       minimumFractionDigits: ckbDecimals,
>       maximumFractionDigits: ckbDecimals,
>     }).format(formatUnit(BigInt(amount ?? '0'), 'ckb') as any)
> ```
> 
> All dirty works are concealed in the `formatAmount` method, and it can be guaranteed by unit tests.

OK! I will do it.

packages/neuron-wallet/tests/utils/to-csv-row/fixtures.json
Keith-CY, line 1: These cases are for rendering various types of assets. The amount formtter should be tested specifically.
Otherwise, it would be hard to add more cases for the formatter, e.g. adding a case of handling overflow

twhy: OK!  
Maybe I should NOT have deleted `shannonToCKB.ts` and `shannonToCKB.test.ts`.
What I should have done is just to rewrite `shannonToCKB` method as below.
```ts
import { formatUnit, ckbDecimals } from '@ckb-lumos/bi'

const shannonToCKB = (shannon: bigint) => {
  return new Intl.NumberFormat('en-US', {
    useGrouping: false,
    signDisplay: 'always',
    minimumFractionDigits: ckbDecimals,
    maximumFractionDigits: ckbDecimals,
  }).format(formatUnit(shannon, 'ckb') as any)
}

export default shannonToCKB
```
And in `to-csv-row.ts`, keep the original code. Also keep `shannonToCKB.test.ts` as tests.
```ts
amount = shannonToCKB(BigInt(tx.value ?? '0'))
```

The `formatAmount` is just the same as `shannonToCKB`.

twhy: I created PR https://github.com/nervosnetwork/neuron/pull/3114 to replace this one. We should probably close this PR, but the discussion is very helpful.

PR #3120, opened 2024-04-12
fix: When set start block number is less than before, remove all of the sync data.

yanguoyu:
When we set a smaller number than before, the UI shows `The block number you set is smaller than the previous set, the synchronization will restart, please proceed with caution.` But actually, it does not remove all of the sync data.

`The last set startBlockNumber should be get before updating the wallet` will fix it.


Keith-CY (approved):

yanguoyu:
@silySuper please test it.

silySuper:
In my impression,start block number will effect by after clicking  clear all sync data ,can not effect after changed（I thought that was the logic before）.Now is effected after be changed.

silySuper (approved):

PR #3126, opened 2024-04-15
refactor: rewrite sudtValueToAmount with formatUnit in neuron-ui

twhy:


Keith-CY (approved):

packages/neuron-ui/src/utils/formatters.ts
```diff
@@ -236,45 +237,18 @@ export const sudtValueToAmount = (
...
-      .join('')}${decimalFraction}`
-  }
-  return `${sign}${+unsignedSUDTValue === 0 ? '0' : unsignedSUDTValue}`
+  const val = value === null || Number.isNaN(+value) ? '0' : value
+  const [int, dec = ''] = formatUnit(val, +decimal).split('.')
+  const fmt = new Intl.NumberFormat('en-US', {
+    useGrouping: showCommaSeparator,
+    signDisplay: showPositiveSign ? 'always' : 'auto',
+  })
+  return `${fmt.format(int as any)}${dec ? `.${dec}` : ''}`
```
homura, line 251:
```suggestion
  // use any type to avoid TS errors since string is not listed in the args IntlFormatter.prototype.format definition but it works
  return `${fmt.format(int as any)}${dec ? `.${dec}` : ''}`
```

homura (approved):

PR #3128, opened 2024-04-17
refactor: rewrite shannonToCKBFormatter with formatUnit

twhy:
The third param `delimiter: string = ','` has been replaced by `showCommaSeparator = true` to make things simpler.

PR #3129, opened 2024-04-17
refactor: rewrite shannonToCKBFormatter with formatUnit

twhy:
1. The third param `delimiter: string = ','` has been replaced by `showCommaSeparator = true` to make things simpler.
2. `showPositiveSign = true` takes effect ONLY when shannon value is > 0 . @Keith-CY


homura:
Only one line was changed compared to the last PR at #3128 

```diff
- signDisplay: showPositiveSign ? 'always' : 'auto',
+ signDisplay: showPositiveSign && +shannon > 0 ? 'always' : 'auto',
```

Since it's a small change, you don't need to close and reopen a new one. You can commit the change after commenting on it. GitHub will mark the previous comment as outdated if you modify the code

twhy:
The unit tests pass on Node.js 20 but not  Node.js 18. I will fix this.

twhy:
## Problem Description
`Intl.NumberFormat.prototype.format()` 
1. DOES NOT format very large number properly in Node.js 18.x (**V8 10.2**).
2. Function properly in Node.js 20.x (**V8 11.3**) and Chromium 120+, **V8 12.x**. 

See it in previous comment.
Keep it simple: **It function property in V8 11.3+**

## Can the code in this PR run properly ?
**YES**.
The code is in `neuron-ui` which run in Chromium 120, **V8.12.x**, and it works as expected.
See it in the screenshot in previous comment.

## Why the Unit Tests workflow failed ?
Unit Tests workflow is run in Node.js 18.12.0 (**V8 10.2**). 
See it in Problem Description.

## How to fix it ?
### Solution 1
1. Upgrade Electron to version 29
2. Update Unit Tests workflow to use Node.js 20.x

### Solution 2
1. Keep using Electron 28.1
2. Split Unit Tests workflow into two workflows
   - Unit tests in `neuron-ui` runs in Node.js 20.x
   - Unit tests in `neuron-wallet` runs in Node.js 18.x

## Support
✅  `Intl.NumberFormat.prototype.format()` works well
❌  `Intl.NumberFormat.prototype.format()` NOT as expected

| Electron | Chromium                      | Node.js                                           |  Note |
|:--------:|:-----------------------:|:--------------------------------:| :---------------|
|   28       |  `120.0.6099.56`, `V8 12.0` ✅ | `18.18.2`, `V8 10.2` ❌ |  We're using version 28.1 |
|   29        |  `122.0.6261.39`, `V8 12.2` ✅ |  `20.9.0`, `V8 11.3`  ✅| Latest version 30.0.0 |

## Discussion
@Keith-CY @homura @yanguoyu @devchenyan 

Please leave comments about your ideas or solutions, or pick one solution from above. Thanks!

homura:
The UDT amount data structure is a u128 while the capacity is a u64. For the shannonToCkb function, the number should not exceed 18446744073709551615. I believe this is an overthink.

yanguoyu:
I guess we will update Electron finally, so it's maybe a solution.

twhy:
This PR has been reviewed before, and all checks have passed after we upgrade Unit Tests Node.js version to 20.11.1
It should be mergable now.

Keith-CY (approved):

homura (approved):

PR #3142, opened 2024-05-07
refactor: refactor ckbCore.utils.xyz with lumos methods

twhy:


packages/neuron-ui/src/components/HDWalletSign/index.tsx
```diff
@@ -56,21 +56,12 @@ const HDWalletSign = ({ tx }: { tx: State.DetailedTransaction }) => {
         } else {
           try {
             if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
-              address = ckbCore.utils.bech32Address(cell.lock.args, {
-                prefix: addressPrefix,
-                type: ckbCore.utils.AddressType.HashIdx,
-                codeHashOrCodeHashIndex: '0x00',
-              })
+              address = generateAddress(cell.lock, { config })
```
Keith-CY, line 59: The address is not derived from the lock, it's actually derived from
```yaml
code_hash: SECP256k1_blake160 # https://github.com/ckb-js/ckb-sdk-js/blob/develop/packages/ckb-sdk-utils/src/address/index.ts#L85
hash_type: type
args: lock.args
```

This usage comes from a deprecated format mentioned at https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0021-ckb-address-format/0021-ckb-address-format.md#deprecated-short-payload-format

packages/neuron-ui/src/components/HDWalletSign/index.tsx
```diff
@@ -56,21 +56,12 @@ const HDWalletSign = ({ tx }: { tx: State.DetailedTransaction }) => {
...
-                args: cell.lock.args,
-                prefix: addressPrefix,
-                type:
-                  cell.lock.hashType === 'data'
-                    ? ckbCore.utils.AddressType.DataCodeHash
-                    : ckbCore.utils.AddressType.TypeCodeHash,
-                codeHash: cell.lock.codeHash,
-              })
+              address = generateAddress(
+                { ...cell.lock, hashType: cell.lock.hashType === 'data' ? 'data' : 'type' },
```
Keith-CY, line 62: Now `hash type` has more than `data`, `type`, so it should be explicitly specified instead of using `type` by default. Error should be thrown if no valid `hash type` is matched.

Ref:
- Valid hash type: https://github.com/nervosnetwork/rfcs/blob/198fc90ab7582953ed85a6655e88e51346857475/rfcs/0051-ckb2023/0051-ckb2023.md?plain=1#L25-L26

packages/neuron-ui/src/components/HDWalletSign/index.tsx
twhy: Let's double check before commit, here's the new code.
```ts
if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
  address = generateAddress(
    {
      args: cell.lock.args,
      codeHash: config.SCRIPTS.SECP256K1_BLAKE160.CODE_HASH,
      hashType: config.SCRIPTS.SECP256K1_BLAKE160.HASH_TYPE,
    },
    { config }
  )
}
```
FYI, `generateAddress` generates deprecated addresses which matches the original code bahavior.

packages/neuron-ui/src/components/HDWalletSign/index.tsx
twhy: 1. When `hashType` is `data` or `type`, use `generateAddress` to generate deprecated address which matches the original code behavior.
2. When 'hashType' is `data1` or `data2`, use `encodeToAddress` to generate `0x00` format type new full address.
3. Throw an error when `hashType` is invalid.

```ts
try {
  if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
    address = generateAddress(
      {
        args: cell.lock.args,
        codeHash: config.SCRIPTS.SECP256K1_BLAKE160.CODE_HASH,
        hashType: config.SCRIPTS.SECP256K1_BLAKE160.HASH_TYPE,
      },
      { config }
    )
  } else if (['data', 'type', 'data1', 'data2'].includes(cell.lock.hashType)) {
    address = ['data', 'type'].includes(cell.lock.hashType)
      ? generateAddress(cell.lock, { config })
      : encodeToAddress(cell.lock, { config })
  } else {
    throw new Error(`Invalid lock script hash type ${cell.lock.hashType}`)
  }
} catch (err) {
  console.error(err)
}
```

packages/neuron-ui/src/components/HDWalletSign/index.tsx
Keith-CY: My bad, seems that the condition of the `if-branch` has limited the `lock.codeHash` and `lock.hashType` to `config.SCRIPTS.SECP256K1_BLAKE160.CODE_HASH` and `config.SCRIPTS.SECP256K1_BLAKE160.HASH_TYPE` so no change is required

packages/neuron-ui/src/components/HDWalletSign/index.tsx
Keith-CY: LGTM, but `generateAddress` can be replaced by `encodeToAddress`, because they are equavelant to the same lock script, just in different format.

twhy: OK. I will commit this.
```ts
try {
  if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
    address = generateAddress(cell.lock, { config })
  } else if (['data', 'type', 'data1', 'data2'].includes(cell.lock.hashType)) {
    address = encodeToAddress(cell.lock, { config })
  } else {
    throw new Error(`Invalid lock script hash type ${cell.lock.hashType}`)
  }
} catch (err) {
  console.error(err)
}
```

yanguoyu (approved):

Keith-CY (approved):

packages/neuron-ui/src/components/ApproveMultisigTxDialog/index.tsx
```diff
@@ -16,7 +17,7 @@ import styles from './approveMultisigTx.module.scss'
 const Cell = React.memo(
   ({ cell, isMainnet }: { cell: State.DetailedInput | State.DetailedOutput; isMainnet: boolean }) => {
     const address = useMemo(
-      () => (cell.lock ? ckbCore.utils.scriptToAddress(cell.lock, isMainnet) : ''),
+      () => (cell.lock ? encodeToAddress(cell.lock, { config: isMainnet ? predefined.LINA : predefined.AGGRON4 }) : ''),
```
homura, line 20: There is already a script/address conversion API implemented in Neuron, maybe we can use it directly. https://github.com/nervosnetwork/neuron/blob/4d1b3bbb3540a833c974b770f6758daf0efbb291/packages/neuron-wallet/src/utils/scriptAndAddress.ts#L5

twhy: Good to know it. 
But `scriptAndAddress.ts` is in `neuron-wallet` so we can't use it in this PR, 
because the changes in this PR are in `neuron-ui`.

homura: I noticed that there are several instances of `{ config: isMainnet ? predefined.LINA : predefined.AGGRON4 })` in this PR. 

Copy-pasting code with logic over three times is not recommended as it can increase complexity. Moreover, `predefined.LINA` and `predefined.AGGRON4` are outdated network names since version 0.22, which were changed to `MAINNET` and `TESTNET`. It is better to refactor the code to avoid increasing complexity

packages/neuron-ui/src/components/HDWalletSign/index.tsx
```diff
@@ -56,21 +56,11 @@ const HDWalletSign = ({ tx }: { tx: State.DetailedTransaction }) => {
         } else {
           try {
             if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
-              address = ckbCore.utils.bech32Address(cell.lock.args, {
-                prefix: addressPrefix,
-                type: ckbCore.utils.AddressType.HashIdx,
-                codeHashOrCodeHashIndex: '0x00',
-              })
+              address = generateAddress(cell.lock, { config })
+            } else if (['data', 'type', 'data1', 'data2'].includes(cell.lock.hashType)) {
```
homura, line 60: Is there a reason for validating the hashType instead of directly validating encodeToAddress(cell.lock, { config })? 

Note that encodeToAddress will check if a script is valid.

twhy: It's based on an earlier conversation.
image


This is definitely simpler
```ts
if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
  address = generateAddress(cell.lock, { config })
} else {
  address = encodeToAddress(cell.lock, { config })
}
```
Erorr (if any)  thrown
```ts
export function ValidateScript(
  ...
  throw new Error(`${debugPath}.hashType must be either data or type!`);
```

The current code is less simple but has more detailed error message.
```ts
throw new Error(`Invalid lock script hash type ${cell.lock.hashType}`)
```

I think they both work, so we just need to pick one. @homura

twhy: Pushed changes. Simple is better.
```ts
if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
  address = generateAddress(cell.lock, { config })
} else {
  address = encodeToAddress(cell.lock, { config })
}
```

homura: cool, it looks fine

twhy:
@Keith-CY 
Since `encodeToAddress` checks script's `hashType` and throws an error if `hashType` is invalid,
the code is now simplified as below.  We are ready to merge this PR after CI checks completes.
```ts
if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
  address = generateAddress(cell.lock, { config })
} else {
  address = encodeToAddress(cell.lock, { config })
}
```

https://github.com/ckb-js/lumos/blob/develop/packages/helpers/src/index.ts#L239
image

https://github.com/ckb-js/lumos/blob/develop/packages/toolkit/src/validators.js#L80
image

packages/neuron-ui/src/components/ApproveMultisigTxDialog/index.tsx
twhy: We can replace `predefined.LINA` `predefined.AGGRON4` with `MAINNET` and `TESTNET` 
when we upgrade to lumos 0.22, so let's put that aside for now.

For `{ config: isMainnet ? predefined.LINA : predefined.AGGRON4 })`   
Let's see the problem from another angle, I can write wrapper functions like
```ts
import { encodeToAddress, generateAddress as _generateAddress } from "@ckb-lumos/helpers";

function generateAddress(script: Script, isMainnet: boolean) {
   return _generateAddress(script, { config: isMainnet ? predefined.LINA : predefined.AGGRON4 }))
}

function scriptToAddress(script: Script, isMainnet: boolean) {
   return encodeToAddress(script, { config: isMainnet ? predefined.LINA : predefined.AGGRON4 }))
}

... Wrap other lumos methods
```
It becomes
`Project code` calling `Wrappers of Lumos methods` calling `Lumos methods`
This means Lumos doesn't solve my problem directly, 
I have to wrap it inside another function to to make things work.

But what we/developers want is
`Project code` calling `Lumos methods`
This means Lumos solves my problem directly.

homura: It is common in development to use 3rdparty libraries within an application. Neuron has an address-related method that involves only two networks - the mainnet and non-mainnet. This makes it possible for Neuron to only require a boolean to control the address conversion instead of a complete `Config`.

Therefore, it is reasonable to have a wrapper for the address conversion in Neuron that uses the following syntax:

```ts
declare function somethingAddress(value: Something, isMainnet: boolean): Address
```

twhy: Added `neuron-ui/src/utils/scriptAndAddress.ts` and calling `scriptToAddress` & `addressToScript` now.

@homura Can you trigger the CI workflows ?

```ts
import { predefined } from '@ckb-lumos/config-manager'
import { encodeToAddress, parseAddress, generateAddress } from '@ckb-lumos/helpers'

const { LINA: MAINNET, AGGRON4: TESTNET } = predefined

const CONFIGS = {
  [MAINNET.PREFIX]: MAINNET,
  [TESTNET.PREFIX]: TESTNET,
}

export const scriptToAddress = (
  script: CKBComponents.Script,
  { isMainnet = true, deprecated = false }: { isMainnet?: boolean; deprecated?: boolean }
) => {
  const config = { config: isMainnet ? MAINNET : TESTNET }
  return deprecated ? generateAddress(script, config) : encodeToAddress(script, config)
}

export const addressToScript = (address: string, { isMainnet = true }: { isMainnet?: boolean } = {}) => {
  const prefix = address.slice(0, 3)
  const config = CONFIGS[prefix] ?? (isMainnet ? MAINNET : TESTNET)
  return parseAddress(address, { config })
}

```

Keith-CY:
> @Keith-CY 
> 
> Since `encodeToAddress` checks script's `hashType` and throws an error if `hashType` is invalid,
> 
> the code is now simplified as below.  We are ready to merge this PR after CI checks completes.
> 
> ```ts
> 
> if (cell.lock.codeHash === systemCodeHash && cell.lock.hashType === 'type') {
> 
>   address = generateAddress(cell.lock, { config })
> 
> } else {
> 
>   address = encodeToAddress(cell.lock, { config })
> 
> }
> 
> ```
> 
> 
> 
> https://github.com/ckb-js/lumos/blob/develop/packages/helpers/src/index.ts#L239
> 
> image
> 
> 
> 
> https://github.com/ckb-js/lumos/blob/develop/packages/toolkit/src/validators.js#L80
> 
> image
> 
> 

LGTM

packages/neuron-ui/src/utils/scriptAndAddress.ts
```diff
@@ -0,0 +1,23 @@
...
+const { LINA: MAINNET, AGGRON4: TESTNET } = predefined
+
+const CONFIGS = {
+  [MAINNET.PREFIX]: MAINNET,
+  [TESTNET.PREFIX]: TESTNET,
+}
+
+export const scriptToAddress = (
+  script: CKBComponents.Script,
+  { isMainnet = true, deprecated = false }: { isMainnet?: boolean; deprecated?: boolean }
```
homura, line 13:
```suggestion
  { isMainnet = true, version = 'CKB2021' }: { isMainnet?: boolean; version?: 'CKB2019' | 'CKB2021' }
```

homura (approved):

packages/neuron-ui/src/utils/scriptAndAddress.ts
twhy: `version` makes more sense when we have multiple versions of addresses. 
e.g. `'CKB2019 | 'CKB2021' | 'CKB2025'`

Right now we just have the full new address format or the deprecated address format.
So `deprecated` is enough and it's convenient and easy to understand and passed in as param.

`true` | `false`  vs `CKB2019` | `CKB2021`

We can change this to version when CKB does have another new version of adress format.

PR #3144, opened 2024-05-08
feat: Export the start block number and slice the light client log.

yanguoyu:
1. Export the start block number https://github.com/Magickbase/neuron-public-issues/issues/386
2. Slice the export light client log https://github.com/Magickbase/neuron-public-issues/issues/378

### DIFF
--- a/packages/neuron-wallet/src/controllers/export-debug.ts
+++ b/packages/neuron-wallet/src/controllers/export-debug.ts
@@ -13,6 +13,7 @@ import SettingsService from '../services/settings'
 import { generateRPC } from '../utils/ckb-rpc'
 import { CKBLightRunner } from '../services/light-runner'
 import { LIGHT_CLIENT_MAINNET, LIGHT_CLIENT_TESTNET } from '../utils/const'
+import WalletsService from '../services/wallets'
 
 export default class ExportDebugController {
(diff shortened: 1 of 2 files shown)


packages/neuron-wallet/src/controllers/export-debug.ts
```diff
@@ -95,30 +97,31 @@ export default class ExportDebugController {
         release,
         vcredist,
       },
+      wallets: wallets.map(v => ({
+        id: v.id,
+        name: v.name,
```
Keith-CY, line 102: `Name` of a wallet is kind of private and it's better not to be exposed in public, and `id` is sensitive too, how about masking some chars of the id in the log

yanguoyu: I will remove the name field. Id is a random string created by `uuid`, I don't know whether id is sensitive

yanguoyu: I have removed the `walletId` in `hd_public_key_info.csv` and used `index` to recognize it as one wallet, and also match the start block number that the user sets.

Keith-CY (approved):

Danie0918:
@homura @devchenyan Please have a review.

devchenyan (approved):

PR #3151, opened 2024-05-15
fix: Replace `Asset` to `Amount` and better 'Asset' data format

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/335#issuecomment-1996306163

For [Optimize input and output storage](https://github.com/nervosnetwork/neuron/pull/2672), Neuron can not recognize whether the transaction is only sent to the current wallet. So, `For transactions sent to yourself, 'Asset' needs to be shown as '--'.` will not implement.

### DIFF
--- a/packages/neuron-ui/src/locales/es.json
+++ b/packages/neuron-ui/src/locales/es.json
@@ -135,7 +135,7 @@
       "activity": "Actividad",
       "datetime": "Fecha y Hora",
       "status": "Estado",
-      "amount": "Cantidad",
+      "asset": "Activos",
       "address": "Dirección",
       "sent": "Enviado",
       "sending": "Enviando",
@@ -305,7 +305,7 @@
         "name": "Nombre de la Billetera",
         "type": "Tipo",
         "balance": "Saldo",
-        "amount": "Cantidad",
+        "asset": "activos",
         "timestamp": "Tiempo",
         "status": "Estado",
         "operation": "Operación"
--- a/packages/neuron-wallet/src/services/tx/transaction-service.ts
+++ b/packages/neuron-wallet/src/services/tx/transaction-service.ts
@@ -256,6 +256,7 @@ export class TransactionsService {
       .addSelect('input.transactionHash', 'transactionHash')
       .addSelect('input.outPointTxHash', 'outPointTxHash')
       .addSelect('input.outPointIndex', 'outPointIndex')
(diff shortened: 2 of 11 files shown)


packages/neuron-ui/src/locales/es.json
```diff
@@ -134,7 +134,7 @@
       "activity": "Actividad",
       "datetime": "Fecha y Hora",
       "status": "Estado",
-      "amount": "Cantidad",
+      "asset": "activos",
```
Keith-CY, line 137: Does `activos` work the same as `Activos` here?

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -373,13 +380,17 @@ export class TransactionsService {
...
     const sums = new Map<string, bigint>()
-    const daoFlag = new Map<string, boolean>()
+    const daoInfo = new Map<string, { inputs: typeof inputs; outputs: typeof outputs }>()
     outputs.map(o => {
       const s = sums.get(o.transactionHash) || BigInt(0)
       sums.set(o.transactionHash, s + BigInt(o.capacity))
 
       if (o.daoData) {
-        daoFlag.set(o.transactionHash, true)
+        if (daoInfo.get(o.transactionHash)) {
```
Keith-CY, line 389:
```suggestion
        if (daoInfo.has(o.transactionHash)) {
```

packages/neuron-wallet/src/services/tx/transaction-service.ts
```diff
@@ -391,7 +402,11 @@ export class TransactionsService {
         return dc.txHash === i.outPointTxHash && dc.index === i.outPointIndex
       })
       if (result) {
-        daoFlag.set(i.transactionHash, true)
+        if (daoInfo.get(i.transactionHash)) {
```
Keith-CY, line 405:
```suggestion
        if (daoInfo.has(i.transactionHash)) {
```

Danie0918:
@homura @devchenyan Please have a review.

devchenyan (approved):

homura (commented):
The PR looks good to me, but the title is somewhat confusing. Maybe a clearer title such as "correct tx history display for dao" would be more appropriate?

homura (approved):

silySuper (approved):

PR #3155, opened 2024-05-17
fix: dev network show mistake sync status

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/359

### DIFF
--- a/packages/neuron-wallet/src/controllers/sync-api.ts
+++ b/packages/neuron-wallet/src/controllers/sync-api.ts
@@ -108,8 +108,13 @@ export default class SyncApiController {
     const rpcService = new RpcService(network.remote, network.type)
     try {
       const syncState = await rpcService.getSyncState()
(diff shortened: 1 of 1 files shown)


yanguoyu:
After throwing an error when `bestKnownBlockNumber` is zero, will the dev sync status show normal?

packages/neuron-wallet/src/controllers/sync-api.ts
```diff
@@ -108,8 +108,13 @@ export default class SyncApiController {
     const rpcService = new RpcService(network.remote, network.type)
     try {
       const syncState = await rpcService.getSyncState()
+      const bestKnownBlockNumber = parseInt(syncState.bestKnownBlockNumber, 16)
+
+      if (bestKnownBlockNumber === 0) {
+        throw new Error()
+      }
```
Keith-CY, line 115: Why is the error thrown when the best known block number is 0

PR #3160, opened 2024-05-20
perf: Optimize synchronization logic for multiple wallets in the light client.

yanguoyu:
Refer to https://github.com/Magickbase/neuron-public-issues/issues/352

I added a [threshold](https://github.com/nervosnetwork/neuron/compare/develop...yanguoyu:opt-light-sync?expand=1#diff-f2ef23eb2dbce7d890a6dc0dc96c7dd8784a29db3313faebefea5718de60227aR26) for different wallets that can sync simultaneously. If the block number difference between the synchronization block number of other wallets and the current wallet block number is not less than this threshold value, they can be synchronized together.

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/light-synchronizer.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/light-synchronizer.ts
@@ -23,6 +23,7 @@ import NetworksService from '../../services/networks'
 import { getConnection } from '../../database/chain/connection'
 
 const unpackGroup = molecule.vector(blockchain.OutPoint)
+const THRESHOLD_BLOCK_NUMBER_IN_DIFF_WALLET = 100_000
 
 export default class LightSynchronizer extends Synchronizer {
   private lightRpc: LightRPC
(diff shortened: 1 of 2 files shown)


packages/neuron-wallet/src/block-sync-renderer/sync/light-synchronizer.ts
```diff
@@ -198,32 +211,30 @@ export default class LightSynchronizer extends Synchronizer {
...
-            +existSyncscripts[scriptToHash(v.script)].blockNumber < +(walletStartBlockMap[v.walletId] ?? 0)
-        )
-        .map(v => {
-          const blockNumber = Math.max(
-            parseInt(walletStartBlockMap[v.walletId] ?? '0x0'),
-            walletMinBlockNumber?.[v.walletId] ?? 0
+        .filter(v => {
+          return (
+            !existSyncScripts[scriptToHash(v.script)] ||
+            +existSyncScripts[scriptToHash(v.script)].blockNumber < walletMinBlockNumber[v.walletId]
```
Keith-CY, line 227: `scriptToHash(v.script)` can be assigned to a variable to avoid hashing it again, or simply

```js
+(exsitSyncScript[scriptToHash(v.script)]?.blockkNumber ?? 0) < walletMinBlockNumber[v.walletId]
```

devchenyan (approved):

Keith-CY (approved):

silySuper:
This pr means that when I change to wallet which block height is bigger than the current wallet and sync for a while,the lower wallet block height will sync also?I can not see this effect.This is my operation video.
https://we.tl/t-3YiWvKPEkr

yanguoyu:
> This pr means that when I change to wallet which block height is bigger than the current wallet and sync for a while,the lower wallet block height will sync also?I can not see this effect.This is my operation video. https://we.tl/t-3YiWvKPEkr

You may have misunderstood. When the current wallet synced block number is lower, after this current wallet reaches the other higher wallet's synced block height, they will sync simultaneously.

silySuper (approved):

PR #3162, opened 2024-05-21
fix: Use hex to avoid unexpected string.

yanguoyu:
Use the `hex` type when transferring the Buffer to the string to avoid unexpected strings.

### DIFF
--- a/packages/neuron-wallet/src/services/settings.ts
+++ b/packages/neuron-wallet/src/services/settings.ts
@@ -88,7 +88,7 @@ export default class SettingsService extends Store {
 
   private generateEncryptString(str: string) {
     if (safeStorage.isEncryptionAvailable()) {
-      return safeStorage.encryptString(str).toString('utf-8')
+      return safeStorage.encryptString(str).toString('hex')
     }
     const hash = crypto.createHash('sha256')
     hash.update(str)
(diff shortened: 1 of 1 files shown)


Keith-CY (approved):

packages/neuron-wallet/src/services/settings.ts
```diff
@@ -88,7 +88,7 @@ export default class SettingsService extends Store {
 
   private generateEncryptString(str: string) {
     if (safeStorage.isEncryptionAvailable()) {
-      return safeStorage.encryptString(str).toString('utf-8')
+      return safeStorage.encryptString(str).toString('hex')
```
homura, line 91: this change might cause the decryption to fail because the decrypted string is no longer the same as before

yanguoyu: This function has not been released, it's a new future in the next version. So I guess it does not matter.

homura: Got it. Can we return the raw buffer instead of the encoded string? Then compare the password with `Buffer.prototype.equals`

yanguoyu: The encoded string needs to be cached on the disk. So it needs to transfer to a string.

homura (approved):

devchenyan (approved):

silySuper (approved):

PR #3163, opened 2024-05-22
refactor: replace LumosCell, LumosCellQuery with Cell, QueryOptions

twhy:


packages/neuron-wallet/src/models/chain/live-cell.ts
```diff
@@ -64,8 +64,8 @@ export default class LiveCell {
       : null
 
     return new LiveCell(
-      cell.outPoint.txHash,
-      cell.outPoint.index,
+      cell.outPoint!.txHash,
+      cell.outPoint!.index,
```
homura, line 68: The `Cell` in Lumos does not distinguish between whether it is an input or output cell. Metaphorically, a `LiveCell` implies that a cell must be an input cell. Therefore, we should throw an error here if a cell does not have an `OutPoint`, meaning it is not an input cell

```js

asserts(cell.outPoint, `Cell must be a live cell`)

return new LiveCell(
  cell.outPoint.txHash,
  cell.outPoint.index,
)

```

packages/neuron-wallet/src/services/live-cell-service.ts
```diff
@@ -16,17 +16,13 @@ export default class LiveCellService {
...
-    data: string | null
-  ): Promise<LumosCell[]> {
+  private async getLiveCellsByScript(lock: Script | null, type: Script | null, data: string | null): Promise<Cell[]> {
     if (!lock && !type) {
       throw new Error('at least one parameter is required')
     }
 
-    const query: LumosCellQuery = { lock, type, data }
-    const liveCells: LumosCell[] = await queryIndexer(query)
+    const query = { lock, type, data } as QueryOptions
```
homura, line 24:
```suggestion
    const query: QueryOptions = { lock, type, data } 
```

or 

```suggestion
    const query = { lock, type, data } satisfies QueryOptions
```

The "as" might cause an error in this context

yanguoyu: May be we can transfer `null` to `undefined`

Keith-CY (approved):

packages/neuron-wallet/src/models/chain/live-cell.ts
twhy: @homura  Is this OK ? 

Since the original code only type checks `cell.outPoint`, so I only replace `Cell` with `CellWithOutPoint`.

packages/neuron-wallet/src/services/live-cell-service.ts
twhy: @homura  As described above, `as` is the only working option for now.

It's just a `null` or `defined` problem here, so using `as` is quite safe. And `getLiveCellsByScript` has good type check on its params.

homura (approved):

twhy:
@Keith-CY This PR can be merged now.

PR #3173, opened 2024-05-28
fix: Fix import hard wallet.

yanguoyu:


packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -34,6 +34,20 @@ import AddressParser from '../models/address-parser'
 import MultisigConfigModel from '../models/multisig-config'
 import { generateRPC } from '../utils/ckb-rpc'
```
twhy, line 36: We need to add some comments explains why the `compressPublicKey()` is needed when importing hardware wallets.

packages/neuron-wallet/src/controllers/wallets.ts
```diff
@@ -34,6 +34,20 @@ import AddressParser from '../models/address-parser'
...
 
+const UNCOMPRESSED_KEY_LENGTH = 130
```
twhy, line 37: We need to add some comments explains why the `compressPublicKey()` is needed when importing hardware wallets.

twhy (approved):

twhy:
@yanguoyu  Thank you for fixing this!

Keith-CY:
Please add tests for this case

silySuper:
s plus import wallet name the wallet is fixed now.
But when find device,it throw an error ,does this error related to this issue?
截屏2024-05-28 15 02 45

packages/neuron-wallet/src/controllers/wallets.ts
yanguoyu: Added

yanguoyu:
> Please add tests for this case



> Please add tests for this case

Added

silySuper (approved):

packages/neuron-wallet/tests/services/hardware/ledger.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    })
+
+    it('if the return pk is zip', async () => {
+      getWalletExtendedPublicKeyMock.mockResolvedValueOnce({ public_key: zipPk, chain_code: '0x' })
+      const ledger = new Ledger({} as any)
+      await ledger.connect()
+      const res = await ledger.getExtendedPublicKey()
+      expect(res.publicKey).toBe(zipPk)
+    })
+  })
```
Keith-CY, line 38:
```suggestion
  describe('test getExtendedPublicKey', () => {
    it('if the return pk is not compressed', async () => {
      getWalletExtendedPublicKeyMock.mockResolvedValueOnce({ public_key: pk, chain_code: '0x' })
      const ledger = new Ledger({} as any)
      await ledger.connect()
      const res = await ledger.getExtendedPublicKey()
      expect(res.publicKey).toBe(zipPk)
    })

    it('if the return pk is compressed', async () => {
      getWalletExtendedPublicKeyMock.mockResolvedValueOnce({ public_key: zipPk, chain_code: '0x' })
      const ledger = new Ledger({} as any)
      await ledger.connect()
      const res = await ledger.getExtendedPublicKey()
      expect(res.publicKey).toBe(zipPk)
    })
  })
```

Keith-CY (approved):

packages/neuron-wallet/tests/services/hardware/ledger.test.ts
Keith-CY: Feel free to merge if these suggestions are not accepted @yanguoyu

PR #3183, opened 2024-06-06
fix: Dev network show mistake sync status

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/359

> https://github.com/nervosnetwork/ckb/issues/4472
> 
> `sync state` only init with local on start, and will update on sync with other node, never with self mined. dev node never has network communicate.

 If it is a devnet node, set `tip_number` to `best_known_block_number` to avoid showing sync failure.

### DIFF
--- a/packages/neuron-wallet/src/controllers/sync-api.ts
+++ b/packages/neuron-wallet/src/controllers/sync-api.ts
@@ -108,8 +108,12 @@ export default class SyncApiController {
     const rpcService = new RpcService(network.remote, network.type)
     try {
       const syncState = await rpcService.getSyncState()
+      let bestKnownBlockNumber = parseInt(syncState.bestKnownBlockNumber, 16)
+      if (network.chain === 'ckb_dev') {
+        bestKnownBlockNumber = parseInt(await rpcService.getTipBlockNumber(), 16)
+      }
       return {
-        bestKnownBlockNumber: parseInt(syncState.bestKnownBlockNumber, 16),
+        bestKnownBlockNumber,
         bestKnownBlockTimestamp: +syncState.bestKnownBlockTimestamp,
       }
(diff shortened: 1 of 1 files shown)


yanguoyu (approved):

Danie0918:
@homura @twhy Please have a review.

packages/neuron-wallet/src/controllers/sync-api.ts
```diff
@@ -108,8 +108,12 @@ export default class SyncApiController {
...
       const syncState = await rpcService.getSyncState()
+      let bestKnownBlockNumber = parseInt(syncState.bestKnownBlockNumber, 16)
+      if (network.chain === 'ckb_dev') {
+        bestKnownBlockNumber = parseInt(await rpcService.getTipBlockNumber(), 16)
+      }
       return {
-        bestKnownBlockNumber: parseInt(syncState.bestKnownBlockNumber, 16),
+        bestKnownBlockNumber,
         bestKnownBlockTimestamp: +syncState.bestKnownBlockTimestamp,
       }
```
twhy, line 118: This is just a suggestion, as I always use `const` when possible.
```ts
const bestKnownBlockNumber = (network.chain === 'ckb_dev')
 ? await rpcService.getTipBlockNumber()
 : syncState.bestKnownBlockNumber

return {
  bestKnownBlockNumber: parseInt(bestKnownBlockNumber, 16),
  bestKnownBlockTimestamp: +syncState.bestKnownBlockTimestamp,
}
```

twhy (approved):

homura (approved):
LGTM, but I suggest changing the title to 'fix: wrong sync status when using devnet'

Keith-CY:
@silySuper please check

PR #3184, opened 2024-06-12
fix: Skip package if the last commit is after the comment

yanguoyu:
I am checking no commit after the trigger comment. If someone pushes a new commit, the package should be repacked again to include the latest commit.

https://github.com/yanguoyu/neuron/actions/runs/9484668688 is the failed example.
https://github.com/yanguoyu/neuron/actions/runs/9484691876 is the success example.

### DIFF
--- a/.github/workflows/package_for_test.yml
+++ b/.github/workflows/package_for_test.yml
@@ -46,6 +46,37 @@ jobs:
         with:
           ref: refs/pull/${{ github.event.issue.number }}/merge
 
+      - name: Ensure no more commits after the triggering comment
+        uses: actions/github-script@v7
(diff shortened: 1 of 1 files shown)


.github/workflows/package_for_test.yml
```diff
@@ -46,6 +46,37 @@ jobs:
         with:
           ref: refs/pull/${{ github.event.issue.number }}/merge
 
+      - name: Check without commit after comment
```
Keith-CY, line 49: This job name could be clearer, e.g.

`ensure no more commits after the triggering comment`

Keith-CY (approved):

devchenyan (approved):

PR #3189, opened 2024-06-18
fix: Warning background in pop-up window

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/396

image


packages/neuron-ui/src/widgets/Dialog/dialog.module.scss
```diff
@@ -50,6 +50,7 @@
   position: relative;
   max-height: calc(100vh - 260px);
   overflow-y: scroll;
+  scrollbar-width: none;
```
yanguoyu, line 53: This will affect the users who use a mouse without wheel.

Keith-CY (approved):

packages/neuron-ui/src/widgets/Dialog/dialog.module.scss
```diff
@@ -49,7 +49,6 @@
   padding: 20px 16px;
   position: relative;
   max-height: calc(100vh - 260px);
```
Keith-CY, line 51: How does it look when the message is long and wrapped so the height is over the `max-height`

devchenyan: > How does it look when the message is long and wrapped so the height is over the `max-height`

image

Keith-CY: It looks good for me because we won't have such a long error message

Danie0918:
@yanguoyu Please have a review.

packages/neuron-ui/src/widgets/Dialog/dialog.module.scss
```diff
@@ -49,7 +49,6 @@
...
-  overflow-y: scroll;
```
yanguoyu, line 52: How about using `overflow-y: auto`? `overflow-y` 's initial value is visible

yanguoyu (approved):

yanguoyu:
@silySuper For this PR, the dialog component CSS has been changed. So It's better to check whether all of the dialog styles work well.

silySuper (approved):

PR #3195, opened 2024-06-21
fix: Show the tip of looking for valid target

yanguoyu:
Fix https://github.com/Magickbase/neuron-public-issues/issues/398
Currently, use the `sync_status` to get the `best_known_block_number `, and check whether the `best_known_block_number` is less than the `assume-valid-target block` number.

After https://github.com/nervosnetwork/ckb/pull/4486 merged, use RPC result to check the status


packages/neuron-wallet/src/controllers/sync-api.ts
```diff
@@ -137,6 +139,14 @@ export default class SyncApiController {
 
     const remainingBlocksToCache = bestKnownBlockNumber - cacheTipNumber
     const remainingBlocksToIndex = bestKnownBlockNumber - indexerTipNumber
+    const setAssumeValidTarget =
+      app.isPackaged &&
+      process.env.CKB_NODE_ASSUME_VALID_TARGET &&
+      process.env.CKB_NODE_ASSUME_VALID_TARGET_BLOCK_NUMBER
+        ? +process.env.CKB_NODE_ASSUME_VALID_TARGET_BLOCK_NUMBER
+        : undefined
```
Keith-CY, line 147: It could be clearer if the var name ends with `block number` and the `else branch` is `Number.NEGATIVE_INFINITY`

yanguoyu: I guess undefined is better which means it does not set this option.

Keith-CY: > I guess undefined is better which means it does not set this option.

Make sense

Danie0918:
@devchenyan @homura @twhy Please have a review.

devchenyan (approved):

Keith-CY (approved):

Keith-CY:
The note `Looking for assumed valid target` appears.

image

PR #3199, opened 2024-06-30
feat: Periodic validation of pending transactions

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/402

### DIFF
--- a/packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
@@ -7,6 +7,7 @@ import TransactionWithStatus from '../models/chain/transaction-with-status'
 import logger from '../utils/logger'
 import { getConnection } from '../database/chain/connection'
 import { interval } from 'rxjs'
+import TxStatus from '../models/chain/tx-status'
 
 type TransactionDetail = {
   hash: string
@@ -18,7 +19,8 @@ type TransactionDetail = {
 const getTransactionStatus = async (hash: string) => {
   const network = NetworksService.getInstance().getCurrent()
   const rpcService = new RpcService(network.remote, network.type)
-  const txWithStatus: TransactionWithStatus | undefined = await rpcService.getTransaction(hash)
+  const txWithStatus: TransactionWithStatus | undefined | { transaction: null; txStatus: TxStatus } =
+    await rpcService.getTransaction(hash)
   if (!txWithStatus) {
     return {
       tx: txWithStatus,
@@ -33,6 +35,13 @@ const getTransactionStatus = async (hash: string) => {
       blockHash: txWithStatus.txStatus.blockHash,
     }
   }
+  if (txWithStatus.txStatus.isRejected()) {
+    return {
+      status: TransactionStatus.Failed,
+      isRejected: true,
+      blockHash: null,
+    }
+  }
   return {
     tx: txWithStatus.transaction,
     status: TransactionStatus.Pending,
@@ -41,16 +50,16 @@ const getTransactionStatus = async (hash: string) => {
 }
 
 const trackingStatus = async () => {
-  const pendingTransactions = await FailedTransaction.pendings()
+  const pendingOrFailedTransactions = await FailedTransaction.pendingOrFaileds()
   await FailedTransaction.processAmendFailedTxs()
 
-  if (!pendingTransactions.length) {
+  if (!pendingOrFailedTransactions.length) {
     return
   }
 
-  const pendingHashes = pendingTransactions.map(tx => tx.hash)
+  const pendingOrFailedHashes = pendingOrFailedTransactions.map(tx => tx.hash)
   const txs = await Promise.all(
-    pendingHashes.map(async hash => {
+    pendingOrFailedHashes.map(async hash => {
       try {
         const txWithStatus = await getTransactionStatus(hash)
         return {
@@ -58,6 +67,7 @@ const trackingStatus = async () => {
           tx: txWithStatus.tx,
           status: txWithStatus.status,
           blockHash: txWithStatus.blockHash,
+          isRejected: !!txWithStatus.isRejected,
         }
       } catch (error) {
         // ignore error, get failed skip current update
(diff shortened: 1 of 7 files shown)


Danie0918:
@yanguoyu @homura Please have a review.

packages/neuron-wallet/src/services/rpc-service.ts
```diff
@@ -29,6 +30,23 @@ export default class RpcService {
     return BlockHeader.fromSDK(result)
   }
 
+  public async getTransactionIncludeRejected(
```
yanguoyu, line 33: There exists a comment on the `getTransaction` function.

```
  /**
   * TODO: rejected tx should be handled
   * {
   *   transaction: null,
   *   txStatus: { blockHash: null, status: 'rejected' }
   * }
   */

How about using `getTransaction` to handle the `reject` condition and remove the comment?
```

packages/neuron-wallet/src/services/rpc-service.ts
```diff
@@ -29,6 +30,23 @@ export default class RpcService {
...
+      return TransactionWithStatus.fromSDK(result)
+    }
+    if (result.txStatus.status === TxStatusType.Rejected) {
+      logger.warn(`Transaction[${hash}] was rejected`)
+      return {
+        transaction: null,
+        txStatus: TxStatus.fromSDK(result.txStatus),
+      }
+    }
+    return undefined
```
homura, line 47: Is the branch `if (result.txStatus.status === TxStatusType.Rejected)` reachable? If a transaction is rejected, the `result.transaction` is not nil but with a `status: rejected`. It goes into the `if(result?.transaction)` branch no matter if it is rejected

devchenyan: > Is the branch `if (result.txStatus.status === TxStatusType.Rejected)` reachable? If a transaction is rejected, the `result.transaction` is not nil but with a `status: rejected`. It goes into the `if(result?.transaction)` branch no matter if it is rejected

A rejected transaction: 

```
{
  transaction: null,
  txStatus: {
    blockHash: null,
    status: 'rejected',
    reason:
      '{"type":"RBFRejected","description":"RBF rejected: replaced by tx Byte32(0xccf7d9b2f739d5b366235c5a17b2d55bb423d3e4ebbbaa6dd446bb1c84c36a60)"}',
  },
  timeAddedToPool: null,
  cycles: null,
  fee: null,
  min_replace_fee: null,
}
```

packages/neuron-wallet/src/services/rpc-service.ts
devchenyan: > There exists a comment on the `getTransaction` function.
> 
> ```
>   /**
>    * TODO: rejected tx should be handled
>    * {
>    *   transaction: null,
>    *   txStatus: { blockHash: null, status: 'rejected' }
>    * }
>    */
> 
> How about using `getTransaction` to handle the `reject` condition and remove the comment?
> ```

Considering the normal call to the getTransaction method, there is no need to separately determine whether the transaction is rejected or not, only whether the return value is undefined or not
So the getTransactionIncludeRejected method is used separately to support the case where rejected transactions need to be handled.

yanguoyu: If the caller does not need the reject status, it can ignore that return result. And check whether the transaction is null or the result is undefined.

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -33,6 +35,13 @@ const getTransactionStatus = async (hash: string) => {
       blockHash: txWithStatus.txStatus.blockHash,
     }
   }
+  if (txWithStatus.txStatus.isRejected()) {
+    return {
+      status: TransactionStatus.Failed,
```
yanguoyu, line 40: Why not add a `Reject` status for `status` to replace adding a variable named `isRejected`?

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -66,16 +76,26 @@ const trackingStatus = async () => {
   )
 
   const failedTxs = txs.filter(
-    (tx): tx is TransactionDetail & { status: TransactionStatus.Failed } => tx?.status === TransactionStatus.Failed
+    (tx): tx is (TransactionDetail | null) & { status: TransactionStatus.Failed; isRejected: boolean } =>
+      tx?.status === TransactionStatus.Failed
   )
   const successTxs = txs.filter(
-    (tx): tx is TransactionDetail & { status: TransactionStatus.Success } => tx?.status === TransactionStatus.Success
+    (tx): tx is (TransactionDetail | null) & { status: TransactionStatus.Success; isRejected: boolean } =>
```
yanguoyu, line 83: When the status is `Success`, the tx must not be null

packages/neuron-wallet/src/block-sync-renderer/tx-status-listener.ts
```diff
@@ -66,16 +76,26 @@ const trackingStatus = async () => {
...
+    (tx): tx is (TransactionDetail | null) & { status: TransactionStatus.Failed; isRejected: boolean } =>
+      !!tx?.isRejected
   )
 
   if (failedTxs.length) {
     await FailedTransaction.updateFailedTxs(failedTxs.map(tx => tx.hash))
   }
 
+  if (rejectedTxs.length) {
+    await FailedTransaction.deleteFailedTxs(rejectedTxs.map(tx => tx.hash))
```
yanguoyu, line 96: If the tx may be `null`, `tx.hash` may throw an exception.

homura (approved):

silySuper:
How to make a transaction be rejected?

silySuper:
What is the relationship between fail and reject?It seems that tx only has this four type?
export enum TxStatusType {
  Pending = 'pending',
  Proposed = 'proposed',
  Committed = 'committed',
  Rejected = 'rejected',
}

silySuper (approved):

Issue #3095, opened 2024-03-24
Sending CKB leads to error unknown error 0x9405


olafwriggers:
Sending CKB from Ledger Nano S leads to unknown error 0x9504. Sending to other Ledger Nano S wallet, Binance and Kucoin fails.

Neuron light client and ledger devices have the newest updates.

<img width="364" alt="image" src="https://github.com/nervosnetwork/neuron/assets/138812918/435af26b-d25b-4213-9fcf-e5a7f3af0b76">

Keith-CY:
> Sending CKB from Ledger Nano S leads to unknown error 0x9504. Sending to other Ledger Nano S wallet, Binance and Kucoin fails.
> 
> Neuron light client and ledger devices have the newest updates.
> 
[...]

PR #3212, opened 2024-07-26
refactor: log error when ledger signing failed

homura:
Resolve https://github.com/Magickbase/neuron-public-issues/issues/418

This PR logs error details when the ledger fails to sign a transaction to help debug the error like https://github.com/nervosnetwork/neuron/issues/3095


packages/neuron-wallet/src/services/hardware/ledger.ts
```diff
@@ -78,12 +78,25 @@ export default class Ledger extends Hardware {
...
+    const signature = await this.ledgerCKB!.signTransaction(hdPath, rawTx, witnesses, context, this.defaultPath).catch(
+      error => {
+        const errorMessage = error instanceof Error ? error.message : String(error)
+        logger.error(
+          'Ledger: failed to sign the transaction ',
+          errorMessage,
+          ' HD path:',
+          hdPath,
+          ' raw transaction:',
+          JSON.stringify(rawTx),
```
Keith-CY, line 91: It might be sensitive to expose transactions in the log. Is it possible to mask some fields, e.g., `args` to be `abc****xyz` and `capacity` to be `0x1*****`(keep the char count consistent)

homura: If a transaction is intended to be sent, it becomes public. I don't think it matters to log a raw transaction in the log file

Keith-CY: The transaction goes public but the sender in reality is a secret. However, the transaction exposed in the log builds the connection between the transaction and the person.

Keith-CY: Maybe we can encrypt the sensitive info and decrypt it in private workspace.

homura: It's quite challenging to fully sanitize a log file to protect all user privacy, as information such as `args`, `data`, and `witnesses` can be linked to a user's public key(public key hash or lock script hash). If we fully or partially hide these parts, it is probably not going to help debug

IMO, the logged data only includes failed transactions, which doesn't reveal too much information. Moreover, it's more effective to track the user through transaction patterns rather than getting logs directly from the user. This is because regular UTxO transactions contain change addresses, and the location of the change UTxOs is relatively consistent

Keith-CY: > It's quite challenging to fully sanitize a log file to protect all user privacy, as information such as `args`, `data`, and `witnesses` can be linked to a user's public key(public key hash or lock script hash). If we fully or partially hide these parts, it is probably not going to help debug
> 
> IMO, the logged data only includes failed transactions, which doesn't reveal too much information. Moreover, it's more effective to track the user through transaction patterns rather than getting logs directly from the user. This is because regular UTxO transactions contain change addresses, and the location of the change UTxOs is relatively consistent

Failed transactions are supposed to be on the chain, so it reveals users' purpose completely, and in advance.

I'm not persuaded by the reason of `more effective to track the users` so the info can be public. As mentioned above, the info can be encrypted, by doing so, the info is still masked while the raw transaction is accessible to limited developers.

I treat the failed transaction as sensitive info because one key feature of Neuron is that the next transaction is unpredictable. If the failed transaction is exposed to the public, it becomes predictable.

packages/neuron-wallet/src/services/log-encryption.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+const JSONSerializer = (_key: string, value: any) => {
+  if (typeof value === 'bigint') {
+    return String(value) + 'n'
+  }
+  return value
+}
+
+logger.info(LogEncryption.getInstance().encrypt('hello world'))
```
yanguoyu, line 73: I guess this is a test

packages/neuron-wallet/src/utils/machineid.ts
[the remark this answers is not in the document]
homura: Yes, the machine ID is unique. If we don't use a unique ID, the encrypted key will change every time we use it. After the discussion, I'll prepend the encrypted AES key to every line of the encrypted log to make the implementation easier to understand

packages/neuron-wallet/src/services/log-encryption.ts
homura: I'll remove it

packages/neuron-wallet/.env
```diff
@@ -120,3 +120,10 @@ MULTISIG_CODE_HASH=0x5c5069eb0857efc65e1bca0c07df34c31663b3622fd3876c876320fc963
 CKB_NODE_ASSUME_VALID_TARGET='0x6dd077b407d019a0bce0cbad8c34e69a524ae4b2599b9feda2c7491f3559d32c'
 CKB_NODE_ASSUME_VALID_TARGET_BLOCK_NUMBER=13007704
 CKB_NODE_DATA_SIZE=56
+
+# openssl rsa -in key.pem -outform PEM -pubout -out public.pem
+# LOG_ENCRYPTION_PUBLIC_KEY="-----BEGIN PUBLIC KEY-----
+# 256-bit RSA KEY IN PEM FORMAT
+# -----END PUBLIC KEY-----"
+# TODO
+LOG_ENCRYPTION_PUBLIC_KEY="-----BEGIN RSA PUBLIC KEY-----\n base64 RSA public key here \n-----END RSA PUBLIC KEY-----"
```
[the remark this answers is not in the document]
homura: The environment variable has a higher priority than `.env`. I left the field in `.env` config mainly as a hint that this field needs to be configured. We can leave it empty here.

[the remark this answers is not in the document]
Keith-CY: It's better to leave it, or mention it in a file that won't be tracked by git by default. Otherwise, users may set their own key in the file and publish it on github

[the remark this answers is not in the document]
homura: I don't think the reason makes sense. It's much more profitable to just change those script configs above to your own deployed malicious contracts than it is to steal the error logs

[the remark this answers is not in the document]
Keith-CY: > I don't think the reason makes sense. It's much more profitable to just change those script configs above to your own deployed malicious contracts than it is to steal the error logs

Removing `public key` from the `.env` is not for defending malicious code, it's for careless developers. There are always developers uploading their own key to GitHub.

homura:
@Keith-CY @yanguoyu I've refactored this PR, and it's ready for review

yanguoyu (approved):

scripts/admin/decrypt-log/run.ts
```diff
@@ -0,0 +1,10 @@
+import { LogDecryption } from "../../../packages/neuron-wallet/src/services/log-encryption";
+
+const ADMIN_PRIVATE_KEY = process.env
+  .ADMIN_PRIVATE_KEY!.split(/\r?\n/)
+  .map((line) => line.trim())
+  .join("\n");
+const LOG_MESSAGE = process.env.LOG_MESSAGE!;
+
+const decryption = new LogDecryption(ADMIN_PRIVATE_KEY);
+console.log(decryption.decrypt(LOG_MESSAGE));
```
Keith-CY, line 10: What if using `gpg` to decrypt the `aes key` so the private key can be guaranteed by `gpg`

homura: If the key is protected by PGP, an admin can use its GPG client to decrypt the message instead of using this script. The script here is an option.

Keith-CY: Got it, please add an annotation for it because the private key is not supposed to be used directly.

Keith-CY (approved):

scripts/admin/decrypt-log/run.ts
homura: I've included a new section to provide instruction on using a protected key

PR #3216, opened 2024-07-31
fix: Exception on checking addresses

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/416

image


yanguoyu (approved):

packages/neuron-wallet/src/locales/en.ts
```diff
@@ -141,6 +141,7 @@ export default {
         '{{path}} has no CKB Node config and storage, press confirm to synchronize from scratch',
       'light-client-sudt-acp-error': "Light client mode doesn't support sending assets to other's asset account",
       'could-not-connect-service': 'Could not connect to the service, please try again later.',
+      'address-required': 'The address cannot be null.',
```
Keith-CY, line 144: `null` is too technical for users, perhaps `empty` is better

Keith-CY (approved):

Keith-CY:
In what case will the address be `null`? For the last empty address in the receiver list?

devchenyan:
> In what case will the address be `null`? For the last empty address in the receiver list?

image

When the receiving address is not entered

PR #3220, opened 2024-08-02
fix: Ignore deleted wallet when ignore duplicate script.

yanguoyu:
Fix https://github.com/Magickbase/neuron-public-issues/issues/417#issuecomment-2264388476


devchenyan (approved):

Keith-CY:
I didn't get the connection between `ignore deleted wallet when ignore duplicate script` and `3.In Windows , set start block number is not effective.` which is mentioned at https://github.com/Magickbase/neuron-public-issues/issues/417#issuecomment-2264388476

yanguoyu:
> I didn't get the connection between `ignore deleted wallet when ignore duplicate script` and `3.In Windows , set start block number is not effective.` which is mentioned at [Magickbase/neuron-public-issues#417 (comment)](https://github.com/Magickbase/neuron-public-issues/issues/417#issuecomment-2264388476)

When finding the duplicate script that belongs to the deleted wallet the new wallet script will not save to the DB at https://github.com/nervosnetwork/neuron/pull/3220/files#diff-da1daa9d927e16165a0cf5cd378178bb7e100c4251d54a4f76f0c5437c5ac9e4L25, then the synced block number will not increase.

silySuper (approved):

PR #3231, opened 2024-08-08
chore: make run.ts independent from neuron

homura:
This PR makes the log decryption independent of Neuron's source code


Keith-CY (approved):

Keith-CY:
This commit can be added to the rc branch https://github.com/nervosnetwork/neuron/tree/rc/v0.117.0

yanguoyu (approved):

PR #3233, opened 2024-08-11
 ci: inject environment variables from github action

Keith-CY:
Inject LOG_ENCRYPTION_PUBLIC_KEY from github action to
encrypt sensitive log


Keith-CY:
Please have a review @devchenyan @homura

PR #3234, opened 2024-08-11
feat: encrypt blake160 list in the log

Keith-CY:
Encrypt the blake160 list when admin pk is available to keep the address list private


homura (approved):
LGTM

yanguoyu (approved):

PR #3236, opened 2024-08-12
fix: remove CRs when inject environment variables 

Keith-CY:
Injecting environment variables with `CR` makes the key mismatched.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -106,7 +106,7 @@ jobs:
         # Inject LOG_ENCRYPTION_PUBLIC_KEY to encrypt sensitive log
       - name: Inject environment variables
         run: |
-          echo "\nLOG_ENCRYPTION_PUBLIC_KEY=${{ secrets.LOG_ENCRYPTION_PUBLIC_KEY }}\n" >> packages/neuron-wallet/.env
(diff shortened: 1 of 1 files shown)


Keith-CY:
Packages with `CR` injected: https://github.com/nervosnetwork/neuron/actions/runs/10345635847
Packages without `CR` injected: https://github.com/nervosnetwork/neuron/actions/runs/10346108387

homura (approved):
Or can we inject an empty line like this?

```sh
echo -en "\n" >> packages/neuron-wallet/.env
echo "LOG_ENCRYPTION_PUBLIC_KEY=${{ secrets.LOG_ENCRYPTION_PUBLIC_KEY }}" >> packages/neuron-wallet/.env
```

Keith-CY:
> Or can we inject an empty line like this?
> 
> ```shell
> echo -en "\n"
> echo "LOG_ENCRYPTION_PUBLIC_KEY=${{ secrets.LOG_ENCRYPTION_PUBLIC_KEY }}" >> packages/neuron-wallet/.env
> ```

This line is injected when the app is being packaged, IMO it's unnecessary to add the empty line

Keith-CY:
Please have a review @yanguoyu @devchenyan

yanguoyu (approved):

PR #3239, opened 2024-08-22
fix: the remaining time display in Nervos DAO is abnormal

devchenyan:
issue: https://github.com/Magickbase/neuron-public-issues/issues/421

image

### DIFF
--- a/packages/neuron-ui/src/components/NervosDAORecord/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAORecord/index.tsx
@@ -91,7 +91,7 @@ export const DAORecord = ({
     epochParser(withdrawEpoch || currentEpoch)
   )
 
-  const leftEpochs = Math.max(compensationEndEpochValue - currentEpochValue, 0)
+  const leftEpochs = Math.max((compensationEndEpochValue || 0) - currentEpochValue, 0)
 
   let leftHours = ''
   let leftDays = ''
--- a/packages/neuron-ui/src/components/NervosDAORecord/index.tsx
+++ b/packages/neuron-ui/src/components/NervosDAORecord/index.tsx
@@ -138,7 +138,7 @@ export const DAORecord = ({
       }
       case CellStatus.Deposited: {
         if (!compensationEndEpochValue) {
-          message = ''
+          message = t('nervos-dao.compensation-period.stage-messages.calculating-the-compensation-cycle')
         } else if (leftHours) {
           message = t('nervos-dao.compensation-period.stage-messages.next-compensation-cycle-hours', {
             hours: leftHours || '--',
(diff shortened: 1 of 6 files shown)


yanguoyu (approved):

packages/neuron-ui/src/components/NervosDAORecord/index.tsx
```diff
@@ -91,7 +91,7 @@ export const DAORecord = ({
     epochParser(withdrawEpoch || currentEpoch)
   )
 
-  const leftEpochs = Math.max(compensationEndEpochValue - currentEpochValue, 0)
+  const leftEpochs = Math.max((compensationEndEpochValue || 0) - currentEpochValue, 0)
```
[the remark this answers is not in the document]
devchenyan: My mistake. When depositEpoch is `''`,  compensationEndEpochValue becomes `NaN` and should not be converted to 0 for use. 
If the `depositEpoch` in depositEpochList is not retrieved in time from the RPC request, the message should be hidden to avoid displaying incorrect information.

https://github.com/nervosnetwork/neuron/blob/e9eab5ee3851106181675e4adc4a19b2b1fad044/packages/neuron-ui/src/components/NervosDAO/hooks.ts#L358

packages/neuron-ui/src/components/NervosDAORecord/index.tsx
```diff
@@ -137,7 +137,9 @@ export const DAORecord = ({
         break
       }
       case CellStatus.Deposited: {
-        if (leftHours) {
+        if (!compensationEndEpochValue) {
+          message = ''
```
Keith-CY, line 141: @Danie0918 we can add a easy-to-understand message for this case

Danie0918: Is this the tip in the calculation?

If so, here's a tip to add `Calculating the next compensation cycle`

Keith-CY: > Is this the tip in the calculation?
> 
> If so, here's a tip to add `Calculating the next compensation cycle`

In some cases, the data for calculation are missing, e.g. fail to get the tip block number from the CKB node, the remaining time cannot be displayed until the tip block number is available. Users should be informed to wait for the calculation instead of an incorrectly estimated remaining time.

Danie0918: cc @devchenyan

Keith-CY: Maybe it's `the current compensation cycle`, the remaining time to the next compensation cycle.

Danie0918: That would be `Calculating the compensation cycle`.

Keith-CY: > That would be `Calculating the compensation cycle`.

LGTM

devchenyan: > That would be `Calculating the compensation cycle`.

image

Keith-CY (approved):

PR #3275, opened 2024-12-10
feat: Support import hardware wallet 24-bit seed

yanguoyu:
This PR will not merged. I open a draft PR.

### DIFF
--- a/packages/neuron-ui/src/components/WalletWizard/hooks.ts
+++ b/packages/neuron-ui/src/components/WalletWizard/hooks.ts
@@ -1,9 +1,9 @@
-import { useState, useCallback } from 'react'
+import { useState, useCallback, useEffect } from 'react'
 
 const MNEMONIC_SENTENCE_WORDS = 12
 
-export const useInputWords = () => {
-  const [inputsWords, setInputsWords] = useState<string[]>(new Array(MNEMONIC_SENTENCE_WORDS).fill(''))
+export const useInputWords = (wordsCount: number = MNEMONIC_SENTENCE_WORDS) => {
+  const [inputsWords, setInputsWords] = useState<string[]>(new Array(wordsCount).fill(''))
   const onChangeInput = useCallback(
     (
       e:
@@ -37,6 +37,9 @@ export const useInputWords = () => {
     },
     [setInputsWords]
   )
+  useEffect(() => {
+    setInputsWords(new Array(wordsCount).fill(''))
+  }, [wordsCount])
   return {
     inputsWords,
     onChangeInput,
--- a/packages/neuron-ui/src/components/WalletWizard/index.tsx
+++ b/packages/neuron-ui/src/components/WalletWizard/index.tsx
@@ -76,6 +76,7 @@ const initState: WithWizardState = {
   password: '',
   confirmPassword: '',
   name: '',
+  isHardware: false,
 }
 
 const submissionInputs = [
@@ -182,6 +183,8 @@ const Welcome = ({ rootPath = '/wizard/', wallets = [], dispatch }: WizardElemen
 
 Welcome.displayName = 'Welcome'
 
+const LEDGER_WORDS_COUNT = 24
+
 const Mnemonic = ({ state = initState, rootPath = '/wizard/', dispatch }: WizardElementProps) => {
   const { generated, imported } = state
   const navigate = useNavigate()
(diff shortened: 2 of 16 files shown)


Keith-CY:
Is this feature ready for testing?

yanguoyu:
> Is this feature ready for testing?

Sure

packages/neuron-wallet/src/services/wallets.ts
```diff
@@ -221,6 +215,59 @@ export class FileKeystoreWallet extends Wallet {
...
+  }
+
+  public getNextReceivingAddresses = async (): Promise<AddressInterface[]> => {
+    const address = await AddressService.getFirstAddressByWalletId(this.id)
+    if (address) {
+      return [address]
+    }
+
+    return []
+  }
```
Keith-CY, line 264: These methods should throw because the wallet of the root path won't derive more.

Keith-CY:
> > Is this feature ready for testing?
> 
> Sure

@silySuper Please have a test with 24-word mnemonics and check if
1. is the address the same as when neuron is connected to ledger
2. transaction can be signed and submitted to the chain

silySuper:
https://github.com/user-attachments/assets/f15c5403-4169-4b86-b21e-ab57fd818131

It is abnormal that send transaction does not need ledger confirm when use wallet imported by 24 keywords.

silySuper:
https://github.com/user-attachments/assets/d03aa0c6-a461-4b54-b05c-bdf0c6c41013

It is abnormal that send transaction does not need ledger confirm when use wallet imported by 24 keywords.

silySuper:
Have cofirmed that it can use password directly.
Verified.

Keith-CY:
Please mark this PR ready for review so we can merge it into an RC branch for the official package.

packages/neuron-wallet/src/services/wallets.ts
yanguoyu: These methods are copied from HardwareWallet, and they will be called.

yanguoyu:
/package
Fixed the wallet that imports from seed is unnecessary to show `verify address`.

silySuper:
verify function is deleted.

Keith-CY:

Keith-CY:

packages/neuron-ui/src/components/WalletWizard/hooks.ts
```diff
@@ -1,9 +1,9 @@
-import { useState, useCallback } from 'react'
+import { useState, useCallback, useEffect } from 'react'
 
 const MNEMONIC_SENTENCE_WORDS = 12
 
-export const useInputWords = () => {
-  const [inputsWords, setInputsWords] = useState<string[]>(new Array(MNEMONIC_SENTENCE_WORDS).fill(''))
+export const useInputWords = (wordsCount?: number) => {
```
homura, line 5: `wordCount = MNEMONIC_SENTENCE_WORDS` to avoid following `wordCount ?? MNEMONIC_SENTENCE_WORDS`

packages/neuron-ui/src/components/WalletWizard/index.tsx
```diff
@@ -193,8 +196,9 @@ const Mnemonic = ({ state = initState, rootPath = '/wizard/', dispatch }: Wizard
     [MnemonicAction.Verify]: 'wizard.replenish-your-seed',
     [MnemonicAction.Import]: 'wizard.input-your-seed',
   }[type]
-  const { inputsWords, onChangeInput, setInputsWords } = useInputWords()
   const [searchParams] = useSearchParams()
+  const isHardware = searchParams.get('isHardware') === 'true'
```
homura, line 200: it would be clearer here to use `wordCount` as the `props` because  the 24-word-mnemonic feature is exclusive for hardware wallets

yanguoyu: I guess `isHardware` is better, then we know the `hardware` wallet should import 24 words.

devchenyan (approved):

homura (approved):
This PR needs to be manually tested with the 24-word mnemonic feature

Keith-CY:
> This PR needs to be manually tested with the 24-word mnemonic feature

@silySuper should have done the test, please confirm again

silySuper:
> 2024-12-11.15.08.26.mov 
> It is abnormal that send transaction does not need ledger confirm when use wallet imported by 24 keywords.

Function has verified  manually yesterday,this wallet in video is created by 24-words which was from my ledger . @Keith-CY @homura

PR #3288, opened 2024-12-18
ci: add a spam comment detection

Keith-CY:
When a comment is added in an issue

1. if the comment has specific words, it will be deleted
2. if the comment has unexpected links, an annotation about safety will be appended.

### DIFF
--- a/.github/workflows/spam-comment-detection.yaml
+++ b/.github/workflows/spam-comment-detection.yaml
@@ -1,0 +1,45 @@
+name: Spam detection
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  detect-spam:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Node.js
+        uses: actions/setup-node@main
+
+      - name: Check for Spam
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const comment = process.env.COMMENT_BODY
+            const spam_words = ['Neuron Support Request']
+            const comment_id = process.env.COMMENT_ID
+            const issue_number = process.env.ISSUE_NUMBER
+            const owner = process.env.REPO_OWNER
+            const repo = process.env.REPO_NAME
+            const EXTERNAL_LINK_REGEXT = /https:\/\/(?!((\w+\.)?github\.com|github\.com|(\w+\.)?magickbase\.com|(\w+\.)?nervos\.org))/gi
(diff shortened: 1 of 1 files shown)


yanguoyu (approved):

.github/workflows/spam-comment-detection.yaml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Set up Node.js
+        uses: actions/setup-node@main
+
+      - name: Check for Spam
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const comment = process.env.COMMENT_BODY
+            const spam_words = ['Neuron Support Request']
```
yanguoyu, line 19: Could we add this variable to GitHub action env?

.github/workflows/spam-comment-detection.yaml
```diff
@@ -0,0 +1,45 @@
...
+            const comment_id = process.env.COMMENT_ID
+            const issue_number = process.env.ISSUE_NUMBER
+            const owner = process.env.REPO_OWNER
+            const repo = process.env.REPO_NAME
+            const EXTERNAL_LINK_REGEXT = /https:\/\/(?!((\w+\.)?github\.com|github\.com|(\w+\.)?magickbase\.com|(\w+\.)?nervos\.org))/gi
+            if (spam_words.some(w => comment.includes(w))) {
+              console.info(`Spam comment: ${comment}`)
+              github.rest.issues.deleteComment({ owner, repo, comment_id })
+            } else if (EXTERNAL_LINK_REGEXT.test(comment)) {
```
yanguoyu, line 28: Should we skip the verification when the creator is a collaborator?

.github/workflows/spam-comment-detection.yaml
Keith-CY: Update by loading spam word list from the secrets

https://github.com/nervosnetwork/neuron/pull/3288/files#diff-c369c155176306e9be1a7df1156a1408083a4b28c9bc392758336e0ec9ff8e05R18

.github/workflows/spam-comment-detection.yaml
Keith-CY: I think this annotation should be for all links and users should be cautious even if the message is from a collaborator.

Keith-CY:

homura (approved):

PR #3292, opened 2025-01-06
ci: set proper permission for spam detection action

Keith-CY:
The default permission of Neuron repo is read-only so the permission write should be specify explicitly


yanguoyu (approved):

Keith-CY:

PR #3326, opened 2025-03-07
fix: HardWallet connection tip

devchenyan:
issue: 
- https://github.com/Magickbase/neuron-public-issues/issues/445

https://github.com/user-attachments/assets/5d97939f-5185-41cc-8af5-4099fa706fb4


packages/neuron-ui/src/components/CellManagement/hooks.ts
```diff
@@ -424,14 +428,16 @@ export const useHardWallet = ({ wallet, t }: { wallet: State.WalletIdentity; t:
   const ensureDeviceAvailable = useCallback(
     async (device: State.DeviceInfo) => {
       try {
-        const connectionRes = await connectDevice(device)
+        const connectionRes = await connectDevice({ ...device, walletID: wallet.id })
```
Keith-CY, line 431: Why should `walletID` be overriden by `wallet.id`

devchenyan: The `device` does not contain `walletID ` field.

Keith-CY (approved):

silySuper (approved):

PR #3327, opened 2025-03-08
feat: Replace react-script to vite

devchenyan:
issue:
- https://github.com/Magickbase/neuron-public-issues/issues/356


package.json
```diff
@@ -53,6 +53,7 @@
     "eslint-config-prettier": "9.1.0",
     "eslint-plugin-prettier": "4.2.1",
     "husky": "8.0.3",
+    "jest": "27.5.1",
```
homura, line 56: Several `@jest/global` modules have been replaced with `vitest` in this PR. Why is `jest` added as a new dependency here?

packages/neuron-ui/src/components/GeneralSetting/index.tsx
```diff
@@ -4,9 +4,9 @@ import { useSearchParams } from 'react-router-dom'
 import Dialog from 'widgets/Dialog'
 import LanguageDialog from 'components/LanguageDialog'
 import AlertDialog from 'widgets/AlertDialog'
-import { ReactComponent as VersionLogo } from 'widgets/Icons/VersionLogo.svg'
-import { ReactComponent as ArrowNext } from 'widgets/Icons/ArrowNext.svg'
-import { ReactComponent as Update } from 'widgets/Icons/Update.svg'
+import VersionLogo from 'widgets/Icons/VersionLogo.svg?react'
```
homura, line 7: we might consider using the `exportType: "named"` option to minimize the impact of the changes in this PR

https://github.com/pd4d10/vite-plugin-svgr/issues/44#issuecomment-1877593407

package.json
devchenyan: jest is used for testing in neuron-wallet. If this dependency is not added, jest will not be available in the `node_modules/bin/`, and tests cannot be executed.

packages/neuron-ui/src/components/GeneralSetting/index.tsx
devchenyan: We also have the practice of directly importing SVGs for use as `<img src={svgStr}`. If we uniformly adopt this approach, it would be challenging to accommodate both scenarios, and the method of using `?react` is the more recommended way of importing.  https://www.npmjs.com/package/vite-plugin-svgr/v/4.2.0

Keith-CY (approved):

homura (approved):

PR #3332, opened 2025-03-27
fix: display a message if the external node is not running with --indexer

devchenyan:
issue:
- https://github.com/Magickbase/neuron-public-issues/issues/454

Pasted Graphic 13

### DIFF
--- a/packages/neuron-ui/src/containers/Navbar/index.tsx
+++ b/packages/neuron-ui/src/containers/Navbar/index.tsx
@@ -116,6 +116,15 @@ const Navbar = () => {
   }, [i18n.language])
 
   useEffect(() => {
+    if (verifyCkbResult && !verifyCkbResult.withIndexer) {
+      showGlobalAlertDialog({
+        type: 'warning',
+        message: t('navbar.ckb-without-indexer'),
+        action: 'ok',
+      })(dispatch)
+      return
+    }
+
     // isUpdated is true or version is not empty means check update has return
     if (!verifyCkbResult || (isUpdated !== true && !version)) {
       return
@@ -140,12 +149,6 @@ const Navbar = () => {
             ]}
           />
         ),
-        action: 'ok',
-      })(dispatch)
-    } else if (!verifyCkbResult.withIndexer) {
-      showGlobalAlertDialog({
-        type: 'warning',
-        message: t('navbar.ckb-without-indexer'),
         action: 'ok',
       })(dispatch)
     }
--- a/packages/neuron-wallet/src/block-sync-renderer/sync/full-synchronizer.ts
+++ b/packages/neuron-wallet/src/block-sync-renderer/sync/full-synchronizer.ts
@@ -26,24 +26,23 @@ export default class FullSynchronizer extends Synchronizer {
   }
 
   private async initSync() {
-    await this.processNextBlockNumber()
+    try {
+      await this.processNextBlockNumber()
 
-    while (this.pollingIndexer) {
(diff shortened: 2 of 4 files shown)


Danie0918:
@Keith-CY @yanguoyu Please have a review.

packages/neuron-ui/src/containers/Navbar/index.tsx
```diff
@@ -116,6 +116,15 @@ const Navbar = () => {
   }, [i18n.language])
 
   useEffect(() => {
+    if (verifyCkbResult && !verifyCkbResult.withIndexer) {
```
Keith-CY, line 119: What's the difference before and after moving this code block?

packages/neuron-wallet/src/services/node.ts
```diff
@@ -280,7 +280,8 @@ class NodeService {
     const network = NetworksService.getInstance().getCurrent()
     try {
       const res = await rpcRequest<{ error?: { code: number } }>(network.remote, { method: 'get_indexer_tip' })
-      if (res.error?.code === START_WITHOUT_INDEXER) {
+
+      if (res.error?.code === START_WITHOUT_INDEXER || (Array.isArray(res) && res.length === 0)) {
```
Keith-CY, line 284: In what case will `Array.isArray(res) && res.length === 0`

devchenyan: `rpcRequest<{ error?: { code: number } }>(network.remote, { method: 'get_indexer_tip' })` , which corresponds to   `curl -s -H 'content-type: application/json' -d '{"id": 0, "jsonrpc": "2.0", "method": "get_indexer_tip"}' http://localhost:8114`
The result is `{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":0}`
According to the handling in the code below, res equals [].
https://github.com/nervosnetwork/neuron/blob/7f3bc4802284529f62d50b0ca73b6e2046264c4a/packages/neuron-wallet/src/utils/rpc-request.ts#L27

packages/neuron-ui/src/containers/Navbar/index.tsx
devchenyan: Before the adjustment, when the version was "", the message could not pop up. After the adjustment, the message can now pop up normally.

Keith-CY (approved):

packages/neuron-wallet/src/block-sync-renderer/sync/full-synchronizer.ts
```diff
@@ -26,24 +26,23 @@ export default class FullSynchronizer extends Synchronizer {
   }
 
   private async initSync() {
-    await this.processNextBlockNumber()
+    try {
```
yanguoyu, line 29: What is the purpose of moving try into this function?

devchenyan: > Then https://github.com/nervosnetwork/neuron/blob/7f3bc4802284529f62d50b0ca73b6e2046264c4a/packages/neuron-wallet/src/block-sync-renderer/sync/queue.ts#L76-L81 will not catch any error. Does it matter?

Only the `initSync` method is called. Since there is no `await`, the `connect` method cannot catch errors from `initSync`. Therefore, all errors are caught within the `initSync` method itself.

yanguoyu (approved):

Keith-CY:
Verified
image

PR #3370, opened 2025-06-17
feat: Identify and prompt for multisig transaction

devchenyan:
issue:
- https://github.com/Magickbase/neuron-public-issues/issues/464
 

image


yanguoyu (approved):

silySuper:
My wallet is right,this tip may mislead user,I think it is better to move it
截屏2025-06-18 10 33 38

silySuper (approved):

Keith-CY (approved):

PR #3490, opened 2026-06-18
ci: stabilize native dependency builds

chenyukang:
## Summary
- Pin unit test and code style jobs to Node 22 instead of floating LTS/default Node.
- Install Linux libudev/libusb headers before the code style bootstrap.
- Bust stale node_modules caches for the updated Node version.
- Pin the test packaging Windows runner to `windows-2022` to avoid the current `windows-latest` VS 18/node-gyp bootstrap failure.
- Pin the test packaging macOS runner to `macos-15-intel` to avoid `macos-latest` arm64 DMG detach failures.

## Why
Current develop and PR CI fail before the package_for_test signing changes are involved: Node LTS now resolves to Node 24, the code style job lacks libusb headers for node-hid, stale node_modules caches can keep an incorrectly installed Electron package around, `windows-latest` exposes a Visual Studio 18 layout that the current node-gyp toolchain does not recognize, and `macos-latest` currently lands on an arm64 macOS image where electron-builder DMG detach can fail with `hdiutil ... Resource busy`.

The package_for_test signing hardening remains isolated in #3489.

## Testing
- ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/unit_tests.yml .github/workflows/check-code-style.yml .github/workflows/package_for_test.yml\n- git diff --check


eval-exec (approved):

eval-exec:
## Code Review

_Posted on behalf of **Claude** (Claude Code) 🤖_

CI-only change, low risk, and it matches the stated rationale. I read the full workflow files (not just the diff) to verify surrounding steps.

### What's correct / good
- ✅ **`matrix.node` is actually consumed** — `unit_tests.yml` setup-node uses `node-version: ${{ matrix.node }}` and the cache key interpolates `node${{ matrix.node }}`. No dangling matrix axis.
- ✅ **Linux guard is right per job.** `unit_tests.yml` correctly gates the `apt-get` step with `if: runner.os == 'Linux'` (multi-OS matrix); `check-code-style.yml` omits the guard, which is correct because that job is `runs-on: ubuntu-latest` only.
- ✅ **`actions/cache@v4`** (named "Restore") still handles both restore and post-job save.
- ✅ **Dropping `lts/*` is a net risk reduction** (Node 24 → 22), not just a pin.

### Suggestions
1. **(Minor, cosmetic) Misleading Windows job name.** The matrix keeps `os: windows-latest` but runs on `runner: windows-2022`, and `name:` uses `${{ matrix.os }}` — so the job displays as **"windows-latest(Node.js 22)"** while actually executing on windows-2022. Consider `os: windows-2022` so the label reflects reality.
2. **(Minor, smell) The `os` axis is now redundant** — with `node` constant at 22 and `runner` carrying the real target, `os` only feeds the display name. Could drop it and name with `matrix.runner`.
3. **(Optional) Code-style job still does a full `yarn` install** (triggering the `node-hid` build that now needs the libusb headers) just to run Prettier. Installing the headers is the correct minimal fix; `--ignore-scripts` would be faster but riskier, so this is fine as-is.
4. **(Out of scope, FYI)** Worth confirming release/packaging workflows that also build native modules aren't still on `lts/*` or missing the same headers. (`package_for_test` signing is intentionally isolated in #3489.)

### Testing note
The PR's `YAML.load_file` + `git diff --check` validation only confirms syntax. For a CI-stabilization change, the real proof is the Actions run going green — recommend confirming all three OS unit-test jobs + code-style pass on this head before merge.

**Verdict: Approve.** Correct and minimal. The only change I'd actually make is the Windows job-name mismatch (#1); the rest is optional polish.

PR #3505, opened 2026-08-27
fix(scripts): await release asset downloads

wangjingshuiku:
The release-draft helper starts each asset pipeline inside `forEach(async ...)`, so `download()` resolves before the pipelines finish and download failures are detached from its promise.

This change awaits `Promise.all(assetList.map(...))` and handles the returned top-level promise by reporting the error and setting a non-zero exit code. The script now completes only after every asset has been written.

Validation: `node --check scripts/download-binaries-from-release-draft.js`.

### DIFF
--- a/scripts/download-binaries-from-release-draft.js
+++ b/scripts/download-binaries-from-release-draft.js
@@ -25,7 +25,7 @@ const download = async (directory) => {
     .then((res) => res.json())
     .then((res) => res.map(({ name, url }) => ({ name, url })))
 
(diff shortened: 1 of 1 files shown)


wangjingshuiku:
Hi @chenyukang, when you have a moment, could you take a look at this PR? Thanks!

It fixes a small issue where the release asset download helper could return before its async download tasks had completed, and download failures could escape the top-level promise.

chenyukang (approved):
Thanks!

